PR #1, opened 2017-05-28
fix(PropTypes): remove deprecated calls to React.PropTypes

jlengstorf:
I found a few calls to `React.PropTypes` hanging around; this fixes ’em.


hellobrian:
👍

PR #5, opened 2017-06-05
chore(build): redo the build for v3

chrisdhanaraj:
There's a whole lotta changes in here to the build, let's go over them

# Storybook
Updated this to the latest version and stripped out our custom webpack config. Instead, I'm just using the unpkg cdn to reference the latest carbon-components css sheet. This is more me just wanting to drop our dependencies as much as possible

# Build
I've removed/redone most of the build. We're now using a node script to produce an `es` and a `cjs` build for our users - `es` for webpack2, `cjs` for webpack 1. Missing from all this is the push to gh-pages for our storybook and a coverage build. For the storybook, I'll be hooking up a CF app pipeline to our repo, and I'll be redoing the coverage with an open-source version ASAP.

# Other changes
- AppContainer has been removed - this used to be the point to inject global styles, but now we're requiring people to bring styles themselves anyways
- ThemeContainer removed - the centering has moved to an inline decorator.

# Sidenotes
- I'll need to go through our devDependencies and remove some things we aren't using anymore


package.json
```diff
@@ -79,28 +76,32 @@
...
+  "dependencies": {
     "classnames": "2.2.5",
     "lodash.debounce": "^4.0.8",
-    "react": "^15.3.2",
     "react-addons-css-transition-group": "^15.4.2",
-    "react-dom": "^15.3.2",
     "warning": "3.0.0"
   },
   "devDependencies": {
     "@ibma/aat": "1.2.0",
```
hellobrian: Do we still have to keep this around? 
If so, do you think we need to mention this package in install docs? 
Otherwise, users outside of IBM will always get an error and may not know that this is a package only available via private NPM registry.

package.json
```diff
@@ -79,28 +76,32 @@
...
+    "carbon-components": "^7.0.0",
     "carbon-icons": "^5.1.2",
+    "react": "^15.3.2",
+    "react-dom": "^15.3.2"
+  },
+  "dependencies": {
     "classnames": "2.2.5",
     "lodash.debounce": "^4.0.8",
-    "react": "^15.3.2",
     "react-addons-css-transition-group": "^15.4.2",
```
hellobrian: Is it possible for us to replace this with `CSSTransitionGroup`?
```
warning react-addons-css-transition-group@15.5.2: This package is deprecated and will no longer work with React 16+. We recommend you use CSSTransitionGroup from 'react-transition-group' instead.
```

hellobrian:
Remove `AppContainer` from tests

package.json
chrisdhanaraj: @ibma/aat? I'm okay with removing it, but we need to get in another tool to scan

package.json
chrisdhanaraj: We can transition, but I'm not sure the transition details. Would be another PR I think

chrisdhanaraj:
@hellobrian removed the old test

.travis.yml
```diff
@@ -1,22 +1,15 @@
...
   directories:
   - node_modules
-before_install:
-- echo "@console:registry=https://dev-console-npm.stage1.ng.bluemix.net/" > ~/.npmrc
-- echo -e "//dev-console-npm.stage1.ng.bluemix.net/:_authToken=\${CONSOLE_NPM_TOKEN}" >> ~/.npmrc
-- echo -e "@ibma:registry=https://npm-registry.whitewater.ibm.com/" >> ~/.npmrc
-- echo -e "//npm-registry.whitewater.ibm.com/:_authToken=${WW_NPM_TOKEN}" >> ~/.npmrc
 script:
 - npm run lint
 - npm run coverage
```
hellobrian: This coverage script in travis.yml should be removed too

PR #11, opened 2017-06-08
feat(inline-select): add inline-select component

tw15egan:
### Added
- `inline-select`

### Changed
- `inline-select` has the label adjacent to the select. Normal select has it outside of the `select` container. Added a check to place the label in the correct spot
- Added tests for `inline-select`

### DIFF
--- a/.storybook/components/SelectStory.js
+++ b/.storybook/components/SelectStory.js
@@ -5,13 +5,27 @@ import SelectItem from '../../components/SelectItem';
 import SelectItemGroup from '../../components/SelectItemGroup';
 
 const selectProps = {
-  onBlur: () => { action('blur'); }, // eslint-disable-line no-console
-  onClick: () => { action('click'); }, // eslint-disable-line no-console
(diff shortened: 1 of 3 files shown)


.storybook/components/SelectStory.js
```diff
@@ -5,13 +5,27 @@ import SelectItem from '../../components/SelectItem';
...
+  }, // eslint-disable-line no-console
+  onMouseEnter: () => {
+    action('mouseEnter');
+  }, // eslint-disable-line no-console
+  onMouseLeave: () => {
+    action('mouseLeave');
+  }, // eslint-disable-line no-console
+  onMouseUp: () => {
+    action('mouseUp');
+  }, // eslint-disable-line no-console
```
hellobrian: IMO, you can remove all of these events since `onChange` is the only event that's used.

chrisdhanaraj:
Repoint this PR towards Master please!

marijohannessen:
👍 ✅

PR #15, opened 2017-06-12
fix(InteriorLeftNavItem): revert back to accepting single children

hellobrian:
Issue raised in Slack:

image
image

This PR goes back to [the previous way this component accepted single children nodes](https://github.com/carbon-design-system/carbon-components-react/blob/7e195d77fe79109a86ef2d61f1c4ab08bb9de5a6/components/InteriorLeftNavItem.js#L46).

### DIFF
--- a/components/InteriorLeftNavItem.js
+++ b/components/InteriorLeftNavItem.js
@@ -10,12 +10,22 @@ const propTypes = {
   onClick: PropTypes.func,
   blankTarget: PropTypes.bool,
   children: PropTypes.node,
+  label: PropTypes.string.isRequired,
 };
 
 const defaultProps = {
   activeHref: '#',
   tabIndex: 0,
+  label: 'InteriorLeftNavItem Label',
   onClick: /* istanbul ignore next */ () => {},
+};
+
+const newChild = (children, tabIndex) => {
+  const child = React.Children.only(children);
+  return React.cloneElement(React.Children.only(child), {
+    tabIndex,
+    className: 'left-nav-list__item-link',
+  });
 };
 
 const InteriorLeftNavItem = ({
@@ -25,16 +35,11 @@ const InteriorLeftNavItem = ({
   onClick,
   tabIndex,
   children,
+  label,
   ...other
 }) => {
   const classNames = classnames('left-nav-list__item', className, {
     'left-nav-list__item--active': activeHref === href,
-  });
-
-  const child = React.Children.only(children);
-  const newChild = React.cloneElement(child, {
-    tabIndex,
-    className: 'left-nav-list__item-link',
   });
 
   return (
@@ -46,7 +51,11 @@ const InteriorLeftNavItem = ({
       onKeyPress={evt => onClick(evt, href)}
       {...other}
     >
-      {newChild}
+      {children
+        ? newChild(children, tabIndex)
+        : <a className="left-nav-list__item-link" href={href} tabIndex={tabIndex}>
+            {label}
+          </a>}
     </li>
   );
 };
--- a/components/InteriorLeftNavItem.js
+++ b/components/InteriorLeftNavItem.js
@@ -8,8 +8,8 @@ const propTypes = {
   activeHref: PropTypes.string,
   tabIndex: PropTypes.number,
   onClick: PropTypes.func,
-  label: PropTypes.string.isRequired,
   blankTarget: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


components/InteriorLeftNavItem.js
```diff
@@ -8,8 +8,8 @@ const propTypes = {
   activeHref: PropTypes.string,
   tabIndex: PropTypes.number,
   onClick: PropTypes.func,
-  label: PropTypes.string.isRequired,
   blankTarget: PropTypes.bool,
+  children: PropTypes.node,
```
jlengstorf: You'll need to add this to `defaultProps` as well to avoid linting complaints.

components/InteriorLeftNavItem.js
```diff
@@ -40,9 +46,7 @@ const InteriorLeftNavItem = ({
       onKeyPress={evt => onClick(evt, href)}
       {...other}
     >
-      <a className="left-nav-list__item-link" href={href} tabIndex={tabIndex}>
-        {label}
-      </a>
+      {newChild}
```
jlengstorf: Isn't this a breaking change if there's no fallback to creating a link?

hellobrian: That's what I'm worried about. 
This is more of a reverting back to an [older version of this component](https://github.com/carbon-design-system/carbon-components-react/blob/7e195d77fe79109a86ef2d61f1c4ab08bb9de5a6/components/InteriorLeftNavItem.js#L46).

- It would be breaking for people who  have already adapted to using the current InteriorLeftNav (ILN).
- But right now this current version doesn't work for people who need to pass in single `children` for use with React Router's `<Link>` component.

Is there a better workaround for this?

components/InteriorLeftNavItem.js
hellobrian: Linting doesn't seem to be complaining about `children` missing from `defaultProps`. How do I even do that?

jlengstorf: I'm using ESLint with `airbnb` as the preset. The rule complains if anything in `Component.propTypes` is both A) not marked `.isRequired` and B) doesn't have a deafult set in `defaultProps`.

My `.eslintrc`, in case you want to play with this:
```json
{
  "root": true,
  "parser": "babel-eslint",
  "parserOptions" : {
    "ecmaVersion": 6,
    "sourceType": "module"
  },
  "env": {
    "es6": true,
    "browser": true,
    "node": false
  },
  "ecmaFeatures": {
    "modules": true
  },
  "extends": [
    "airbnb"
  ],
  "globals": {
  "_contextRoot": true
  },
  "rules" : {
    "max-len": [2, 130, 4],
    "no-plusplus:": 0,
    "no-underscore-dangle": 0,
    "no-param-reassign": ["error", { "props": false }],
    "react/jsx-filename-extension": [1, { "extensions": [".js"] }]
  }
}
```

components/InteriorLeftNavItem.js
jlengstorf: You could check if `children` is empty. I'd probably add a helper function:

```jsx
const getLink = (children, label) => {
  if (children) {
    // return the child
  }

  // return a generated link using the current pattern
};
```

You could deprecate the `label` approach if you wanted, maybe?

hellobrian: Wonderful, let me try this and update the PR

components/InteriorLeftNavItem.js
jlengstorf: Nevermind. I just realized that `children` is a special case and everything I've been saying here is irrelevant for it. 🤷‍♀️

components/InteriorLeftNavItem.js
```diff
@@ -46,7 +51,11 @@ const InteriorLeftNavItem = ({
       onKeyPress={evt => onClick(evt, href)}
       {...other}
     >
-      {newChild}
+      {children
+        ? newChild(children, tabIndex)
+        : <a className="left-nav-list__item-link" href={href} tabIndex={tabIndex}>
+            {label}
+          </a>}
```
jlengstorf: Looks awesome. Nice work!

components/InteriorLeftNavItem.js
chrisdhanaraj: @jlengstorf note that this library doesn't use the airbnb style guide - just eslint:recommended + prettier

PR #22, opened 2017-06-13
Pagination update

tw15egan:
### Pagination Updates

To render a pagination component when the total pages is unknown, pass down `unknownPages: true` to the `pagination` component.

```jsx
<Pagination unknownPages={true} />
```

screen shot 2017-06-13 at 2 56 41 pm

### DIFF
--- a/components/Pagination.js
+++ b/components/Pagination.js
@@ -28,6 +28,7 @@ class Pagination extends Component {
     disabled: PropTypes.bool,
     page: PropTypes.number,
     pageSize: PropTypes.number,
+    pagesUnknown: PropTypes.bool,
   }
   static defaultProps = {
     backwardText: 'Backward',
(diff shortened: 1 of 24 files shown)


components/Pagination.js
```diff
@@ -45,6 +46,7 @@ class Pagination extends Component {
     page: this.props.page,
     pageSize: ((this.props.pageSize && this.props.pageSizes.includes(this.props.pageSize)) ?
       this.props.pageSize : this.props.pageSizes[0]),
+    pagesUnknown: this.props.pagesUnknown || false
```
hellobrian: For defaultProps, you should be able to write: 
```
pagesUnknown: false
```

Then the user doesn't have to explicitly give the prop a boolean value. Instead, they can write:

```
// pagesUnknown is true
<Pagination pagesUnknown />

// pagesUnknown is false
<Pagination />
```

chrisdhanaraj: +1 to this, then good to go

PR #23, opened 2017-06-13
Inline select

tw15egan:
Reopening https://github.com/carbon-design-system/carbon-components-react/pull/11 to point at master


marijohannessen:
👍 ✅

PR #32, opened 2017-06-14
fix: use propTypes from module instead of React

eddiemonge:
Use the PropTypes modules instead of using the deprecated
React.PropTypes method


chrisdhanaraj:
👍

PR #38, opened 2017-06-15
fix(activeHref): set active href in constructor, not mount

chrisdhanaraj:
Currently InteriorLeftNav wipes out initial state on `componentDidMount` - reset to let the initial state declaration accept `window.location.pathname` if nothing else is passed in


ycao56:
:+1: looks good to me

PR #63, opened 2017-06-23
fix(overflowmenu): bring over the rest of the IE11 fix

chrisdhanaraj:
Whoops! Thought just removing the blur would fix the IE11 bug - but nope, needs the rest of the old fix as well.

### DIFF
--- a/components/OverflowMenu.js
+++ b/components/OverflowMenu.js
@@ -83,12 +83,6 @@ class OverflowMenu extends Component {
     this.setState({ open: false });
   };
 
-  handleBlur = evt => {
-    if (!this.menuEl.contains(evt.relatedTarget)) {
-      this.setState({ open: false });
-    }
-  };
-
   render() {
     const {
       id,
@@ -130,7 +124,6 @@ class OverflowMenu extends Component {
           className={overflowMenuClasses}
           onClick={this.handleClick}
           onKeyDown={this.handleKeyPress}
-          onBlur={this.handleBlur}
           aria-label={ariaLabel}
           id={id}
           tabIndex={tabIndex}
--- a/components/OverflowMenu.js
+++ b/components/OverflowMenu.js
@@ -80,6 +80,10 @@ class OverflowMenu extends Component {
   };
 
   handleClickOutside = () => {
+    this.closeMenu();
+  };
+
+  closeMenu = () => {
     this.setState({ open: false });
   };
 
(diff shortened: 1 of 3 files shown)


marijohannessen:
Do we really want to have the icon be clickable and not the wrapper div? Cause I have to hit the icon perfectly in order to ope  the menu which can be a bit annoying. I'm also having issues with storybook in IE11 so I can't test this properly - do we need a polyfill or something to make it work?

chrisdhanaraj:
Storybook doesn't work with IE11 at the moment - see: https://github.com/storybooks/storybook/issues/1352. To test, I basically `yarn link` my local carbon-components-react, and then `yarn link carbon-components-react` inside a create-react-app application. Not the prettiest work around, so I'll take a look at getting IE11 and storybook operational with some of the workarounds listed in that storybook issue

As far as the icon itself - yes, I think programmatically this is what we want, otherwise the action gets triggered on _any_ click inside of the menu (including any of the menu items, which isn't exactly what we want and is clobbering IE11 events). However, you're right - it drops the click target a little bit, we should probably move the padding from the container element to the icon itself.

tw15egan:
With the padding fix in the vanilla repo this change looks good to me 👍 ✅

iangfleming:
yeah I'm having some trouble with create-react-app so I haven't tested this properly but it *looks* good to me 🙂

hellobrian:
Hey @chrisdhanaraj, when I do `yarn link` in local carbon-components-react, and then `yarn link carbon-components-react` inside a create-react-app application, I'm getting failed to compile error. I'm doing this on my windows pc, but I'm probably just doing something wrong. 

If you've already confirmed that this works on IE11, then that's good enough for me. Code looks good 👍

chrisdhanaraj:
Bluemix [toolchain](https://console.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170713-150805 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170713-150805") to [Bluemix Design System](https://console.bluemix.net/apps/33e31ce0-a0a0-4a4f-8a79-b4799ef408d8?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit 98e44188867ef3d2c328303827e24241277a98c7

PR #65, opened 2017-06-27
feat(Slider): add slider component

iangfleming:


package.json
```diff
@@ -87,7 +87,6 @@
   },
   "dependencies": {
     "classnames": "2.2.5",
-    "flatpickr": "^3.0.5-1",
```
tw15egan: Not sure we want to remove `flatpickr` and `bowser`

iangfleming: oops

iangfleming: fixed

tw15egan: 👍

components/Slider.js
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  updatePosition = (evt) => {
+    if (evt && this.props.disabled) {
+      return;
+    }
+
+    if(evt) {
+      if(evt.dispatchConfig) { evt.persist(); } // fix
```
chrisdhanaraj: what's the fix comment here for?

components/Slider.js
```diff
@@ -0,0 +1,246 @@
...
+      } = this.calcValue(evt);
+
+      this.setState({
+        left,
+        value: newValue,
+       });
+    });
+  }
+
+  calcValue = (evt) => {
```
chrisdhanaraj: this is kind of a general comment on this function (and I know this probably should've been done on the vanilla version) - but this function is... pretty long. There are a couple sections where it might help readability if it was broken into a couple different functions. 

At the end of the day, you're calculating the `left` and `newValue` values here - you could think about breaking the `mousemove | click | touchmove` into a function, same for `keydown`

components/Slider.js
```diff
@@ -0,0 +1,246 @@
...
+
+  handleTouchEnd = () => {
+    this.element.ownerDocument.removeEventListener('touchmove', this.updatePosition);
+    this.element.ownerDocument.removeEventListener('touchup', this.handleTouchEnd);
+    this.element.ownerDocument.removeEventListener('touchend', this.handleTouchEnd);
+    this.element.ownerDocument.removeEventListener('touchcancel', this.handleTouchEnd);
+  }
+
+  renderChildren() {
+    return React.Children.map(this.props.children, (child) => {
```
chrisdhanaraj: Instead of using `React.Children.map` directly, I would actually go with `React.children.toArray` and then `map` with that. The toArray discards `null` children - sometimes people have conditional children, and `React.Children.map` errors out.

components/Slider.js
```diff
@@ -0,0 +1,246 @@
...
+      transform: `translate(0%, -50%) scaleX(${left / 100})`,
+    }
+    const thumbStyle = {
+      left: `${left}%`,
+    }
+    return (
+      <div className="bx--slider-container">
+        <span className="bx--slider__range-label">{min}{minLabel}</span>
+        <div
+          id={id}
```
hellobrian: I think the `id` needs to go on the `<input type="hidden">` in order for this to be properly associated with a `<label>`. 

For example, I=in your `SliderStory.js`:

```js
<label htmlFor="slider" className="bx--label">Slider Label</label>
<Slider id="slider" value={50} min={0} max={100} step={1} onChange={(evt) => { console.log(evt)}}>
  <TextInput id="input-for-slider" className="bx-slider-text-input"/>
</Slider>
```

We want users to be able to click on the `label` element to access the `input` and this should also make it more predictable for screen-readers to know how to navigate and input data into a form. 

Resources:
- https://stackoverflow.com/a/3051510/2058360

.storybook/components/SliderStory.js
```diff
@@ -0,0 +1,22 @@
...
+  .addWithInfo(
+    'default',
+    `
+      lorem
+    `,
+    () => (
+      <div style={{ marginTop: '2rem' }}>
+        <div className="bx--form-item">
+          <label htmlFor="slider" className="bx--label">Slider Label</label>
+          <Slider id="slider" value={50} min={0} max={100} step={1} onChange={(evt) => { console.log(evt)}}>
```
hellobrian: I think it would be good to offer 2 variants of `Slider`: one with a label, and one without.
In my opinion, it's something we should do for all of our Form elements. You're already providing a variant without the label, so making a variant with a label can help to ensure that a user can give an `id` prop value once and it'll be used for `id` and `htmlFor`.

.storybook/components/SliderStory.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Slider from '../../components/Slider';
+import TextInput from '../../components/TextInput';
+
+storiesOf('Slider', module)
+  .addWithInfo(
+    'default',
+    `
+      lorem
```
hellobrian: Add a real description here and anything else that users may need to know about this React component. [Here's our description from the website](http://carbondesignsystem.com/components/slider/code):

> Sliders provide a visual indication of adjustable content, where the user can move the handle along a horizontal track to increase or decrease the value.

.storybook/components/SliderStory.js
hellobrian: Also, can you show the disabled Slider as a Story as well?

components/Slider.js
hellobrian: Are you able to write this as:

```js
if (evt && evt.dispatchConfig) {
  evt.persist();
}
```

.storybook/components/SliderStory.js
```diff
@@ -0,0 +1,22 @@
...
+    `,
+    () => (
+      <div style={{ marginTop: '2rem' }}>
+        <div className="bx--form-item">
+          <label htmlFor="slider" className="bx--label">Slider Label</label>
+          <Slider id="slider" value={50} min={0} max={100} step={1} onChange={(evt) => { console.log(evt)}}>
+            <TextInput id="input-for-slider" className="bx-slider-text-input"/>
```
hellobrian: I'm noticing that I'm unable to type new values into the `TextInput`, is this a known issue?

components/Slider.js
iangfleming: I agree in theory but if you're okay with it I'd rather merge it in its current state and revisit it later.

asudoh: @chrisdhanaraj You seem to have been concerned with long functions recently - Do you want to introduce sometihg like this? http://eslint.org/docs/rules/complexity

chrisdhanaraj:
@iangfleming just tested this as a controlled component, still works great :+1:

just one... tiny thing. I can't seem to actually erase the value of the textInput to blank the value and then put a single digit number

chrisdhanaraj:
@iangfleming I uhhhh change my mind on the last thing, probably better to have the digit issue rather than never being able to type zero. We can fiddle with that later - revert that last change and then good to merge

hellobrian:
Hey @iangfleming I setup the Slider component in a codesandbox to test if using it in a form would properly submit data.

[Edit Form Inputs with submit](https://codesandbox.io/s/DRr8zoDK6)

☝️ Unfortunately, I'm not seeing it submit data the way I expect. 
What I expected to see was this:
- I click submit
- The URL changes to include a query string

components/Slider.js
```diff
@@ -0,0 +1,260 @@
...
+      id,
+      onChange,
+      min,
+      minLabel,
+      max,
+      maxLabel,
+      labelText, 
+      step,
+      required,
+      disabled,
```
hellobrian: TL;DR: add `...other` here as well, see [original comment for details](https://github.com/carbon-design-system/carbon-components-react/pull/65#issuecomment-317458907)

components/Slider.js
```diff
@@ -0,0 +1,262 @@
...
+        <label htmlFor={id} className="bx--label">{labelText}</label>
+        <div className="bx--slider-container">
+          <span className="bx--slider__range-label">{min}{minLabel}</span>
+          <div
+            className={sliderClasses}
+            ref={node => {
+              this.element = node;
+            }}
+            onClick={(evt) => this.updatePosition(evt)}
+            {...other}
```
hellobrian: I think the `{ ...other }` props should be on the `<input>`, at least for me, I need to add `name` attribute in order for this component to submit data. And any other handlers as well

[Edit Form Inputs with submit](https://codesandbox.io/s/DRr8zoDK6)

Maybe instead, we provide a `name` prop that the user can use and `...other` can go

chrisdhanaraj:
Bluemix [toolchain](https://console-regional.ng.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console-regional.ng.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170802-190905 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170802-190905") to [Bluemix Design System](https://console.bluemix.net/apps/0a253845-97bd-46c9-aa04-a220aeb8ed06?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit fa177b4f709a8ee1112551d748117e184da5533f

PR #72, opened 2017-06-30
fix(Tooltip): update position if changed beyond threshold

Noodles006:
Applies to any of the following positions: top/right/bottom/left/width/height

### DIFF
--- a/components/Tooltip.js
+++ b/components/Tooltip.js
@@ -14,6 +14,7 @@ class Tooltip extends Component {
     showIcon: PropTypes.bool,
     iconName: PropTypes.string,
     iconDescription: PropTypes.string,
+    positionThreshold: PropTypes.number,
   };
 
   static defaultProps = {
@@ -22,6 +23,7 @@ class Tooltip extends Component {
     showIcon: true,
     iconName: 'info--glyph',
     iconDescription: 'tooltip',
+    positionThreshold: 5,
   };
 
   state = {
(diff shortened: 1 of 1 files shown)


Noodles006:
@chrisdhanaraj Could you help review this updates for Tooltip component. Thanks.

hellobrian:
For future reference: [commit message guidelines](https://github.com/carbon-design-system/carbon-components-react/blob/master/CONTRIBUTING.md#commit-message-guidelines)

components/Tooltip.js
```diff
@@ -22,23 +23,36 @@ class Tooltip extends Component {
     showIcon: true,
     iconName: 'info--glyph',
     iconDescription: 'tooltip',
+    positionThreshold: 5,
```
chrisdhanaraj: @Noodles006 apologies was out last week.

For clarification - why 5 / why a prop at all? Seems to me the default behavior should be that if it exceeds a certain distance away to always recalculate the position

Noodles006: @chrisdhanaraj there is a property named triggerPosition in the state, which indicates the position of the tooltip content, right? but this property will only been calculate when component did mounted, so if a tooltip component in dialog and the component's position is updated, then the tooltip content will not been updated.

iangfleming:
Hey! Can you go into further detail about why the threshold prop is needed? I have a similar fix but it doesn't have a threshold: https://github.com/carbon-design-system/carbon-components-react/pull/81
I feel like I'm not quite understanding your fix and I just wanted to make sure I'm not missing something important 🙂

chrisdhanaraj:
Closing for lack of response.

PR #75, opened 2017-06-30
test: add a basic test for index exports

eddiemonge:
An incorrect path in the index.js file would cause the library to fail
to import. This does a basic sanity check for that

### DIFF
--- a/__tests__/index-test.js
+++ b/__tests__/index-test.js
@@ -1,0 +1,7 @@
+import * as Carbon from '../index';
+
+describe('Carbon Components React', () => {
+  it('can be imported using the correct path', () => {
+    expect(typeof Carbon).toBe('object')
+  });
+});
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "check": "npm run lint && npm run test",
     "lint": "eslint {components,internal}/**",
-    "test": "jest '/(components|lib|internal)/'",
-    "test-watch": "jest '/(components|lib|internal)/' --watch",
+    "test": "jest '/(__tests__|components|lib|internal)/'",
+    "test-watch": "jest '/(__tests__|components|lib|internal)/' --watch",
     "prepublish": "npm run build",
     "build": "node scripts/build.js",
     "commitmsg": "validate-commit-msg",


asudoh:
Looks that the case used to be covered by [`import/no-unresolved` rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md) that is included in [AirBnB ruleset](https://github.com/airbnb/javascript/blob/3a9e1c830f9141207322254fc83ec3efae7b790b/packages/eslint-config-airbnb-base/rules/imports.js#L36), but it seems to have been lost as part of [our switch to recommended ruleset](https://github.com/carbon-design-system/carbon-components-react/commit/1e037c4c2d31cd1874542e15e1b1e6ef21fcdc36#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R59). So I’d suggest adding `import/no-unresolved` rule to our ESLint configuration.

__tests__/index-test.js
```diff
@@ -0,0 +1,7 @@
+import * as Carbon from '../index';
+
+describe('Carbon Components React', () => {
+  it('can be imported', () => {
```
hellobrian: Since the test can only fail based on path, can you change the `it` description to something like: 
`it('can be imported using the correct path')` just so that this is more specific.

chrisdhanaraj:
Bluemix [toolchain](https://console.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170707-200832 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170707-200832") to [Bluemix Design System](https://console.bluemix.net/apps/1a30065f-cef2-4933-ba3f-3ff6669a0abd?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit 04def752c5ad6efaa33e9976f6372a46412415a7

PR #86, opened 2017-07-11
fix(search): toggle the clear icon correctly and enable use on controlled components

chrisdhanaraj:
Couple of bugs with the search

1) The clear icon was showing all the time, instead of only with content
2) The clear icon did not work when controlling the value of Search

This resolves both of those issues

### DIFF
--- a/.storybook/components/SearchStory.js
+++ b/.storybook/components/SearchStory.js
@@ -16,11 +16,11 @@ storiesOf('Search', module)
     `,
     () =>
       <Search
+        key="normal"
         {...searchProps}
         className="some-class"
         id="search-1"
         labelText="Search"
-        defaultValue={'something'}
         placeHolderText="Search Bluemix Offerings"
       />
   )
@@ -34,6 +34,7 @@ storiesOf('Search', module)
     `,
     () =>
       <Search
+        key="small"
         {...searchProps}
         className="some-class"
         small
@@ -41,4 +42,39 @@ storiesOf('Search', module)
         labelText="Search"
         placeHolderText="Search Bluemix Offerings"
       />
+  )
+  .addWithInfo(
+    'Controlled',
+    `
+      You can control the Search input like you would with a normal input as well. See the Storybook source to see the source code behind this at https://github.com/carbon-design-system/carbon-components-react/blob/master/.storybook/components/SearchStory.js
+    `,
+    () => {
+      class ControlledSearch extends React.Component {
+        state = {
+          searchValue: 'something',
+        };
+
+        handleChange = evt => {
+          this.setState({
+            searchValue: evt.target.value,
+          });
+        };
+
+        render() {
+          return (
+            <Search
+              {...searchProps}
+              className="some-class"
+              id="search-1"
+              labelText="Search"
+              value={this.state.searchValue}
+              onChange={this.handleChange}
+              placeHolderText="Search Bluemix Offerings"
+            />
+          );
+        }
+      }
+
+      return <ControlledSearch />;
+    }
   );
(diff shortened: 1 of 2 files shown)


.storybook/components/SearchStory.js
```diff
@@ -14,15 +14,16 @@ storiesOf('Search', module)
...
       or as a filter to aid the user in finding content.
     `,
-    () => (
+    () =>
       <Search
+        key="normal"
         {...searchProps}
         className="some-class"
         id="search-1"
         labelText="Search"
```
hellobrian: Sorry, unrelated to the PR but related to Search:

I think we need two variants of Search controlled by props.

As I'm using the components more, I'm finding the need to have access to either an `input` or `label + input` together. I mentioned to @iangfleming on his [Slider PR](https://github.com/carbon-design-system/carbon-components-react/pull/65#discussion_r126332560) that we should start offering 2 variants of our form elements controlled by props so that the user can choose which variant they need.
Also, It looks like we never actually show the `<label>`for it to be usable. This could be confusing for users who may try to use their own `<label>` with `Search` component. If we decide

.storybook/components/SearchStory.js
```diff
@@ -14,15 +14,16 @@ storiesOf('Search', module)
       without the use of navigation. Search can be used as the primary means of discovering content,
       or as a filter to aid the user in finding content.
     `,
-    () => (
+    () =>
       <Search
+        key="normal"
```
hellobrian: What are the `key` props for?

components/Search.js
```diff
@@ -47,6 +58,16 @@ class Search extends Component {
     }
   };
 
+  handleChange = evt => {
+    this.setState({
+      hasContent: evt.target.value !== '',
+    });
+
+    if (this.props.onChange) {
```
hellobrian: Is this the same thing as putting `onChange` in `defaultProps`?

```js
static defaultProps = {
  onChange: () => {},
}
```

chrisdhanaraj:
@hellobrian removed the keys (was a test I was doing, left it in there by accident) and added the defaultProp

.storybook/components/SearchStory.js
chrisdhanaraj: Was testing something, removed them now

components/Search.js
chrisdhanaraj: ah yeah, good catch - updated

hellobrian:
fixes https://github.com/carbon-design-system/carbon-components-react/issues/85

tw15egan:
👍 ✅

chrisdhanaraj:
Bluemix [toolchain](https://console.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170717-154621 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170717-154621") to [Bluemix Design System](https://console.bluemix.net/apps/fbee5a28-786f-4c34-b0e9-4eb019356724?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit 947d4e34aaf4b0d5e23cd96017e67d7caeb7b1ef

PR #88, opened 2017-07-13
feat(pagination): items per page when pages unknown, pageInputDisabled prop, i18n

holmansze:
- enable "items per page" when `pagesUnknown` is `true`
- add new prop `pageInputDisabled`
- externalize some i18n strings
- fix some console exceptions


holmansze:
Thanks @asudoh - right.  Reverted the version.

chrisdhanaraj:
:+1: could you port these changes to v3 as well?

PR #95, opened 2017-07-14
fix(DetailPageHeader): Correctly remove scroll event listener

### LINKED ISSUE #94: DetailPageHeader detaches the wrong event listener on componentWillUnmount
In https://github.com/carbon-design-system/carbon-components-react/blob/e8faa8c0be6da847c28e61013074bb70efe99aa2/components/DetailPageHeader.js#L32 it attaches creates a debounced wrapper around the scroll handler and attaches this to the window scroll event. 

However, in componentWillUnmount(), it attempts to remove the original handleScroll method instead of the debounced wrapper.

I'll send a PR with a fix shortly.

nfriedly:
Fixed the bug, added a test.

The test is in it's own `describe()` wrapper to localize the before/after hooks. Not absolutely required, but seemed like a good idea.

I also added [sinon](http://sinonjs.org/) as a dev dependency to make stubbing global methods a bit easier and less error-prone.

Fixes #94

### DIFF
--- a/components/DetailPageHeader.js
+++ b/components/DetailPageHeader.js
@@ -30,8 +30,8 @@ class DetailPageHeader extends Component {
   };
 
   componentDidMount() {
-    const debouncedScroll = debounce(this.handleScroll, 25);
-    window.addEventListener('scroll', debouncedScroll);
+    this._debouncedScroll = debounce(this.handleScroll, 25);
+    window.addEventListener('scroll', this._debouncedScroll);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -45,7 +45,7 @@ class DetailPageHeader extends Component {
   }
 
   componentWillUnmount() {
-    window.removeEventListener('scroll', this.handleScroll);
+    window.removeEventListener('scroll', this._debouncedScroll);
   }
 
   handleScroll = () => {
--- a/components/__tests__/DetailPageHeader-test.js
+++ b/components/__tests__/DetailPageHeader-test.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
+import window from 'window-or-global';
 import DetailPageHeader from '../DetailPageHeader';
 import Icon from '../Icon';
 import Breadcrumb from '../Breadcrumb';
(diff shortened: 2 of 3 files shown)


chrisdhanaraj:
@nfriedly - I'm fine with the code change, but why add `sinon`? Could just use Jest spies

nfriedly:
I guess Jest has a bit more mocking functionality than the last time I looked at it, although it doesn't seem to have anything like sinon's sandbox for tracking and restoring multiple mocks at once. It's only two methods, though, so I'll switch it over.

nfriedly:
Done. I remove sinon and updated jest to the current release (because some of the mocking functionality I needed wasn't available in the previous version), and updated the test.

chrisdhanaraj:
Hate to do this - but Jest also has the equivalence of an assertion library in its matchers (https://facebook.github.io/jest/docs/using-matchers.html). Apologies for the nit-picking! I'm probably over-wary of adding additional libraries

nfriedly:
Ok, that's fair.  FWIW, assert is built-in to Node.js, but no worries, I switched it to Jest's `expect`.

chrisdhanaraj:
Ah good point, but thanks!

chrisdhanaraj:
Bluemix [toolchain](https://console-regional.ng.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console-regional.ng.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170802-190905 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170802-190905") to [Bluemix Design System](https://console.bluemix.net/apps/0a253845-97bd-46c9-aa04-a220aeb8ed06?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit 59cf4d9e8ea0363b9acb1b8e285d185d6dbbad24

PR #104, opened 2017-07-18
fix(ally): add alt attribute for components using Icon

lilyclyc:


lilyclyc:
@chrisdhanaraj -- Please help to review it, thank you.

components/Checkbox.js
```diff
@@ -41,6 +41,7 @@ const Checkbox = ({ className, id, labelText, onChange, ...other }) => {
           <Icon
             className="bx--checkbox-checkmark"
             description="checkmark"
+            alt="checkmark"
```
chrisdhanaraj: for these `alt` tags and below, could you actually move them to a `defaultProp` and then reference that prop here? It'll allow for people to pass in different internationalized strings if need be

hellobrian: It was pointed out in https://github.com/carbon-design-system/carbon-components-react/issues/87#issuecomment-314853593 that we need to add `alt` attributes to all of our `Icon` components any way. 

So in `Icon`, couldn't we just add `alt` as a prop and as a defaultProp?

shyfruan: Have created a prop "iconDescription" to replace the hard coded value "checkmark"

components/Checkbox.js
```diff
@@ -40,8 +42,8 @@ const Checkbox = ({ className, id, labelText, onChange, ...other }) => {
         <span className="bx--checkbox-appearance">
           <Icon
             className="bx--checkbox-checkmark"
-            description="checkmark"
-            alt="checkmark"
+            description={iconDescription}
+            alt={iconDescription}
```
chrisdhanaraj: could you actually change this into two different props

```
descriptionText
altText
```

with both defaults being the same in this scenario?

shyfruan: Have created another prop iconAlt

shyfruan:
@chrisdhanaraj conflict has been resolved and any more comments?

chrisdhanaraj:
👍  good by me

PR #137, opened 2017-07-25
fix(DetailPageHeader): identify children when minified

nfriedly:
This sets a `displayName` property on all of the children that `DetailPageHeader` has special cases for, and then updates the DPH to check `displayName` rather than `name`. (`name` changes during minification, but `displayName` does not.)

I'm not completely happy with this, partially because I'm not sure how to test for it, but it should at least fix the current buggy behavior. I suppose we could minify the code and then test it, but that sounds like a giant pain. 

Alternatively, there could be a [lint rule that just required all components to specify a `displayName`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md), but that would be a bigger change. It wouldn't be a bad thing to enforce though, especially if we want to allow some code to depend on component names internally.

(I dropped the `instanceof` idea because it's [not as straightforward as it sounds](https://stackoverflow.com/questions/39387405/using-instanceof-to-test-for-base-class-of-a-react-component). Given that some of the components are stateless functions that you don't directly create instances of, this kind of makes sense.)

Fixes https://github.com/carbon-design-system/carbon-components-react/issues/136

### DIFF
--- a/components/Breadcrumb.js
+++ b/components/Breadcrumb.js
@@ -16,6 +16,9 @@ const Breadcrumb = ({ children, className, ...other }) => {
   );
 };
 
+// needed to allow DetailPageHeader to identify this when minified
+Breadcrumb.displayName = 'Breadcrumb';
+
 Breadcrumb.propTypes = propTypes;
 
 export default Breadcrumb;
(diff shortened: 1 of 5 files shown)


germanattanasio:
@chrisdhanaraj can you help get this merge? It's a priority for us and that's why we reported and fixed the issue.

nfriedly:
Superseded in #140

PR #150, opened 2017-07-31
fix(DatePicker): fix server-side-rendering (#76)

### LINKED ISSUE #76: Bug: Server side rendering is broken after adding flatpickr
Hi @marijohannessen, When you added `flatpickr` here: https://github.com/carbon-design-system/carbon-components-react/commit/6bc3176868abed0e509d30a4bf6f899aaf1a2626 you broke server-side rendering :(

Users trying to use server-side rendering should use `v3.2.0`

See https://github.com/coderhaoxin/react-flatpickr/issues/9

nfriedly:
Adds a workaround for the flatpickr SSR bug and a basic SSR smoke test to ensure the lib can be at least loaded in a server-side context.

The workaround and the test are both a little barren, but I believe they're both good enough, and a solid improvement over the previous situation. The test has to live outside of jest because jest will pollyfill `window` and make things work in jest that  throw errors in Node.js.

It wouldn't be a bad idea to add an additional test that renders everything in `.storybook/components/` in Node.js to get a little more thorough coverage.

Fixes #76

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,9 @@
   "scripts": {
     "check": "npm run lint && npm run test",
     "lint": "eslint {components,internal}/**",
-    "test": "jest '/(__tests__|components|lib|internal)/'",
+    "test": "jest '/(__tests__|components|lib|internal)/' && npm run test-ssr",
     "test-watch": "jest '/(__tests__|components|lib|internal)/' --watch",
+    "test-ssr": "npm run build && node server-side-rendering-tests/*.js",
     "prepublish": "npm run build",
     "build": "node scripts/build.js",
     "commitmsg": "validate-commit-msg",
@@ -87,7 +88,7 @@
   },
   "dependencies": {
     "classnames": "2.2.5",
-    "flatpickr": "^3.0.6",
+    "flatpickr": "^3.0.7",
     "lodash.debounce": "^4.0.8",
     "react-addons-css-transition-group": "^15.4.2",
     "warning": "3.0.0",
--- a/server-side-rendering-tests/loads.js
+++ b/server-side-rendering-tests/loads.js
@@ -1,0 +1,9 @@
+// Simple test to ensure the components can at least be loaded in Node.js
+// Note: this test is *not* run by Jest because Jest's pollyfills mask some errors
+
+'use strict';
+
+var assert = require('assert');
+var carbonComponentsReact = require('../cjs/index');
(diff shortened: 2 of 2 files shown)


asudoh:
My quick test suggests that Rollup does not (seem to) understand a file with ES2015 `import` and CJS `require()` mixed (though the change works well with WebPack). Two possible ways I can think of to solve the problem are:

* Just let Carbon React users to import Flatpickr by themselves, instead of our DatePicker component import it
* Provide a Babel plugin to codemod Flatpickr to wrap it in IIFE in run it only when `window` object is there

nfriedly:
Good point. I'm going to switch tactics and try to get the root cause fixed: https://github.com/chmln/flatpickr/pull/969

hellobrian:
@nfriedly Any updates on this?

nfriedly:
The root cause is fixed in flatpickr, but an updated version hasn't been published to npm yet. Feel free to leave a comment on https://github.com/chmln/flatpickr/pull/969 asking for a new release.

tw15egan:
@nfriedly @hellobrian Flatpickr has been updated to `3.0.7`

nfriedly:
There we go, now that the root cause is fixed, this lib gets a much simpler change: Just bump the dependency version and add a test.

chrisdhanaraj:
Bluemix [toolchain](https://console-regional.ng.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console-regional.ng.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170810-155531 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170810-155531") to [Bluemix Design System](https://console.bluemix.net/apps/597e3f6e-c208-4e71-97d0-0e4982545e05?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit b9e69db90d1c11cd0582b72bc8869eb373ca50f7

PR #160, opened 2017-08-02
fix(overflow): only use menuEl if it exists

chrisdhanaraj:
Wrapping around to make sure this.menuEl isn't used before this.menuEl is actually defined

### DIFF
--- a/components/OverflowMenu.js
+++ b/components/OverflowMenu.js
@@ -64,8 +64,10 @@ class OverflowMenu extends Component {
   }
 
   getMenuPosition = () => {
-    const menuPosition = this.menuEl.getBoundingClientRect();
-    this.setState({ menuPosition });
+    if (this.menuEl) {
+      const menuPosition = this.menuEl.getBoundingClientRect();
+      this.setState({ menuPosition });
+    }
   };
 
   handleClick = evt => {
--- a/components/OverflowMenu.js
+++ b/components/OverflowMenu.js
@@ -64,10 +64,8 @@ class OverflowMenu extends Component {
   }
 
   getMenuPosition = () => {
-    if (this.menuEl) {
-      const menuPosition = this.menuEl.getBoundingClientRect();
-      this.setState({ menuPosition });
-    }
+    const menuPosition = this.menuEl.getBoundingClientRect();
+    this.setState({ menuPosition });
   };
 
   handleClick = evt => {
@@ -89,6 +87,10 @@ class OverflowMenu extends Component {
 
   closeMenu = () => {
     this.setState({ open: false });
+  };
+
+  bindMenuEl = menuEl => {
+    this.menuEl = menuEl;
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


components/OverflowMenu.js
```diff
@@ -64,8 +64,10 @@ class OverflowMenu extends Component {
   }
 
   getMenuPosition = () => {
-    const menuPosition = this.menuEl.getBoundingClientRect();
-    this.setState({ menuPosition });
+    if (this.menuEl) {
```
eddiemonge: what happens if it doesn't exist? Does it get retried at any point?

superkhau:
@chrisdhanaraj Can we land this? We need this landed to unblock https://github.ibm.com/velox/ui/pull/605

components/OverflowMenu.js
hellobrian: @chrisdhanaraj ? 

There's probably a good reason for this but I'm not sure exactly why this is always needed. 
My understanding is: since `getMenuPosition()` is used in `componentDidMount()`, then it will have access to `refs` ...but according to this PR, it's being implied that the component attempts to use `refs` before they're defined.

hellobrian: @chrisdhanaraj  ?

I didn't realize it was possible for `refs` to get used before they're defined since this `getMenuPosition()` method is invoked inside `componentDidMount()` ... I don't see a difference when I remove the if statement.

asudoh: Interesting in understanding what condition `this.menuEl` gets missing, too - One guess I can make is some kind of race condition, but what @chrisdhanaraj found may be something else.

asudoh: CC @jlpadilla

chrisdhanaraj:
@asudoh / @hellobrian / @eddiemonge - swapped to a different syntax, but same outcome. We were running into one of the caveat scenarios of ref callbacks

https://facebook.github.io/react/docs/refs-and-the-dom.html#caveats

Notably, that the ref actually runs twice - first with null, second with the actual DOM element. The if statement escapes it out of the first `null` value, but allows it with the DOM element

asudoh:
Got it, thanks @chrisdhanaraj for elaborating!! From what we experienced, I’d suggest we introduce [`jsx-no-bind` eslint-plugin-react rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md).

hellobrian:
✅👍

chrisdhanaraj:
Bluemix [toolchain](https://console-regional.ng.bluemix.net/devops/toolchains/6b8b7163-1ec8-4616-82fa-aa467a17af03 "storybook-toolchain-v2"): [Delivery Pipeline](https://console-regional.ng.bluemix.net/devops/pipelines/b84aaac8-8484-4a1e-902a-ebb526eb7998 "storybook-toolchain-v2") deployed [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react/releases/tag/deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170808-205936 "deploy-ibm-yp-us-south-chrisdhanaraj_org-Bluemix-Design-System-20170808-205936") to [Bluemix Design System](https://console.bluemix.net/apps/eadc6714-41e7-421f-8313-02618b55185f?env_id=ibm:yp:us-south "ibm:yp:us-south:chrisdhanaraj_org:Bluemix Design System"), including commit 7af1a0fa1c8199d86652d4ed05372e8b303a50da

Issue #186, opened 2017-08-11
Make the DetailPageHeader more composable


AnthonyOliveri:
Possibly related to issue #147.

Currently, the DetailPageHeader includes a title, and accepts up to 4 additional child components: `Breadcrumb`, `Tabs`, `OverflowMenu`, and `Icon`. There are a few more components that we want to add to it on the developer console. I would open a pull request, but I am not sure how you guys would want this to be implemented.

For context, here is the prototype of what we are trying to create. Note the `Node` language tag and `View demo example` link, as well as the description below the title.

<img width="1220" alt="screen shot 2017-08-11 at 6 18 02 pm" src="https://user-images.githubusercontent.com/5459406/29235157-1eca536c-7ec2-11e7-9dad-8de8c352edf7.png">

hellobrian:
Hey @AnthonyOliveri, in general, the most important criteria we look for in approving pull requests is making sure that the code doesn't have any breaking changes. 

You can see our full [contribution guidelines here](https://github.com/carbon-design-system/carbon-components-react/blob/master/CONTRIBUTING.md).

Specifically, the `DetailPageHeader` component is a bit tricky since there isn't one composition that works for everyone. 

Here's what you could do for now:

If you think you have a way to refactor `DetailPageHeader` so that it resolves your problem while not breaking any other scenarios we currently support, then make a pull request to `carbon-components-react` repo.

Otherwise, make a pull request to [carbon-addons-bluemix-react](https://github.com/carbon-design-system/carbon-addons-bluemix-react).

One thing I can do on my end:

- Ask my designers if they have any updated ideas about DetailPageHeader options. If there are multiple common patterns that we're seeing across IBM Cloud, then we can capture those compositions into React components.

P.S. To better help you out in the long run, can you tell me what team you work with?

AnthonyOliveri:
Thanks for the feedback @hellobrian. I will try a pull request to modify `DetailPageHeader` first. Is there any documentation on how `carbon-addons-bluemix-react` works (how to contribute, how to add it to a project, etc)? I don't see it published on either npmjs.com or npm whitewater.

I work on the Developer Experience team in the Application Runtimes and Frameworks org. For more context on what we're working on, here is a screenshot of what the header currently looks like:

![screen shot 2017-08-14 at 10 40 55 am](https://user-images.githubusercontent.com/5459406/29279354-2359adb6-80dd-11e7-8339-8de4a31e6a0f.png)

[...]

PR #194, opened 2017-08-16
feat(DetailPageHeader): Add Tag and Link children

AnthonyOliveri:
More context is given in issue #186.
 
This gives the option to provide 2 additional child nodes to the `DetailPageHeader`: a `Tag` and a `Link`. Below is a screenshot from the Storybook.

screen shot 2017-08-16 at 12 02 37 pm

### DIFF
--- a/.storybook/components/DetailPageHeaderStory.js
+++ b/.storybook/components/DetailPageHeaderStory.js
@@ -8,6 +8,8 @@ import DetailPageHeader from '../../components/DetailPageHeader';
 import Tabs from '../../components/Tabs';
 import Tab from '../../components/Tab';
 import Icon from '../../components/Icon';
+import Tag from '../../components/Tag';
+import Link from '../../components/Link';
 
 const detailPageHeaderProps = {
   title: 'Detail Page Header',
@@ -51,9 +53,14 @@ storiesOf('DetailPageHeader', module)
           {...overflowMenuItemProps}
           itemText="Delete App"
           isDelete
-          isLastItem
         />
       </OverflowMenu>
+      <Tag type="ibm">
+        Language
+      </Tag>
+      <Link href="#">
+        View demo
+      </Link>
     </DetailPageHeader>
   ))
   .addWithInfo('with tabs', () => (
@@ -85,5 +92,11 @@ storiesOf('DetailPageHeader', module)
         <Tab label="Banana" />
         <Tab label="Orange" />
       </Tabs>
+      <Tag type="ibm">
+        Language
+      </Tag>
+      <Link href="#">
+        View demo
+      </Link>
     </DetailPageHeader>
   ));
--- a/components/DetailPageHeader.js
+++ b/components/DetailPageHeader.js
@@ -7,6 +7,8 @@ import Breadcrumb from './Breadcrumb';
 import Tabs from './Tabs';
 import OverflowMenu from './OverflowMenu';
 import Icon from './Icon';
+import Tag from './Tag';
+import Link from './Link';
 
 class DetailPageHeader extends Component {
   static propTypes = {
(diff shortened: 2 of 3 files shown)


.storybook/components/DetailPageHeaderStory.js
```diff
@@ -51,9 +53,14 @@ storiesOf('DetailPageHeader', module)
           {...overflowMenuItemProps}
           itemText="Delete App"
           isDelete
-          isLastItem
```
AnthonyOliveri: This isn't related to this pull request, but I noticed that this prop no longer exists in the `OverflowMenuItem`, so I removed it.

components/DetailPageHeader.js
```diff
@@ -111,24 +113,32 @@ class DetailPageHeader extends Component {
     let tabs;
     let overflow;
     let icon;
+    let tag;
+    let link;
 
     Children.map(children, child => {
-      if (child.type === Breadcrumb) {
-        breadcrumb = child;
+      if (child != null && child.type != null) {
```
AnthonyOliveri: Preventing null reference errors.

components/DetailPageHeader.js
```diff
@@ -111,24 +113,32 @@ class DetailPageHeader extends Component {
...
+        switch (child.type) {
```
AnthonyOliveri: Changed the `if` statements to a `switch` since I thought that would be more appropriate for the logic.

components/DetailPageHeader.js
```diff
@@ -152,6 +162,12 @@ class DetailPageHeader extends Component {
               {' '}
               <span className="bx--detail-page-header-status-text">
                 {statusText}{statusContent}
+                <span className="bx--detail-page-header-tag">
+                  {' '}{tag}
```
AnthonyOliveri: Not sure if this is the preferred way to do this, but I was unsure of how else to add a bit of spacing without CSS.

components/DetailPageHeader.js
```diff
@@ -152,6 +162,12 @@ class DetailPageHeader extends Component {
               {' '}
               <span className="bx--detail-page-header-status-text">
                 {statusText}{statusContent}
+                <span className="bx--detail-page-header-tag">
```
AnthonyOliveri: I know it is not ideal to nest the `Tag` and `Link` children inside the `bx--detail-page-header-status-text`, but I did so to keep the elements inline without breaking backwards compatibility by renaming the span className.

AnthonyOliveri:
I added some code review comments to explain my reasoning.

chrisdhanaraj:
@AnthonyOliveri Hey Anthony! Sorry about the huge delay it getting this reviewed

I like what you're doing here, but I think we can further abstract it out a little bit. Odds are eventually somebody will want to add some other type of status in there, so I think it'd be best to abstract it out one step further. Instead of adding `Link` and `Tag` in there (which are basically your apps specific needs), let's just add any child that's _not_ one of the current ones in the design to where you've stuck `Link` and `Tag`.

So you could turn your switch statement into something like...

```js
const filteredChildren = Children.toArray(children).filter(child => {
      switch (child.type) {
        case Breadcrumb:
          breadcrumb = child;
          return false;
        case Tabs:
          tabs = child;
          return false;
        case OverflowMenu:
          overflow = child;
          return false;
        case Icon:
          icon = child;
          return false;
        default:
          return true;
      }
    });
```

And then use filtered children where you have `Link/Tag`

```js
<span className="bx--detail-page-header-status-text">
                {statusText}
                {statusContent}
                {filteredChildren}
              </span>
```

AnthonyOliveri:
That's an awesome idea @chrisdhanaraj! I'll try that out.

How should we handle spacing between the elements in `filteredChildren`? Should we leave them right next to each other and let the consumer modify the CSS in their project? Or add the `{' '}` elements from my 4th code review comment?

hellobrian:
@chrisdhanaraj ☝️ can you advise?

jzhang300:
@chrisdhanaraj I just recently worked on a new generic DetailPageHeader component that will supersede this.  Can we close this PR?

components/DetailPageHeader.js
```diff
@@ -112,24 +112,23 @@ class DetailPageHeader extends Component {
...
+          tabs = child;
+          return false;
+        case OverflowMenu:
+          overflow = child;
+          return false;
+        case Icon:
+          icon = child;
+          return false;
+        default:
+          return true;
```
eddiemonge: this should be `return null`

joshblack:
Hi there! 👋  

Thanks to everyone in this PR for their interest and time invested in helping out with `carbon-components-react`. I'm going to go and close this PR just because of in-activity, but please feel free to re-open if anyone believes this to be a valuable addition to the repo.

All I'm trying to do is go through older PRs and issues and clean-up ones that haven't seen as much activity recently, so am 100% willing to re-visit this and anything else that may be closed if those involved still believe them to be worthwhile 😄

Issue #218, opened 2017-09-05
ModalWrapper does not have the option to close after submission


Flyingcirc-Danger:
The `handleSubmit` property of the `ModalWrapper` will fire a function, but won't close the modal after success. While I don't thing this behavior should be default, it would be nice to have an option to toggle this, perhaps as a top level prop.

marijohannessen:
Fixed in #412

PR #219, opened 2017-09-05
Feature - Close after submit ModalWrapper 

Flyingcirc-Danger:
#218

Added an extra prop to the ModalWrapper that gives the choice of closing after submitting. This should ideally be reimplemented with async functions in the future so that the `submit` function would only close after fully resolving`.

To close after submit, the `handleSubmit` function should return the boolean  `true` and the `closeAfterSubmit` prop should be `true`

### DIFF
--- a/components/ModalWrapper.js
+++ b/components/ModalWrapper.js
@@ -51,6 +51,7 @@ class ModalWrapper extends React.Component {
       passiveModal,
       primaryButtonText,
       secondaryButtonText,
+      handleSubmit,
       ...other
     } = this.props;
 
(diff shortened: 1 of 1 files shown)


hellobrian:
Just merged your other PR, but looks like it may have caused a merge conflict. We can start reviewing this PR as soon as the conflict is resolved

Flyingcirc-Danger:
My mistake, I didn't rebase. Conflicts resolved.

joshblack:
Hi there! 👋 

If you're still interested in getting this merged into the project, could you add tests for the added behavior for `ModalWrapper`?

marijohannessen:
@Flyingcirc-Danger

PR #231, opened 2017-09-11
feat(DetailPageHeaderCustom): Make DetailPageHeader more generic

jzhang300:
This is a separate enhancement to make DetailPageHeader more generic for https://github.com/carbon-design-system/carbon-components-react/issues/186.

Here are a list of changes I made from the original DetailPageHeader component:
* Deconstructed the component into 5 generic sections: `Top`, `Middle`, `Bottom`, `Side`, `Extra`.
* Restructured the HTML and CSS in a way so that each section can show or hide independently without breaking UI of other sections,
  * Made sure that the `Side` Section maintains vertical alignment
  * `Top` and `Bottom` Sections uses `visibility: hidden; height: 0` instead of `transform` so that the hidden state is easier to maintain
  * `Side` hugs to the left while its sibling `.bx--detail-page-header-custom__content` takes up the remaining space in the row
* `Icon` uses the `-9999px` hack ensure more robust icon centering
* Allow for manual control of `Expanded` or `Collapsed` state of component
* Clean up SCSS

screen shot 2017-09-11 at 6 10 32 pm

This PR supersedes https://github.com/carbon-design-system/carbon-components-react/pull/194


jzhang300:
I wasn't sure where to put the SCSS or whether just creating a totally new component for the original DetailPageHeader was the right thing to do.

joshblack:
Hi @jzhang300! Should we talk about how this could be incorporated into the addons repo?

marijohannessen:
@jzhang300

jzhang300:
@joshblack @marijohannessen I'm going to close this pr because this should really be submitted to Carbon Add-ons

PR #264, opened 2017-10-09
feat(checkbox): allow more props to be rendered when not modified

jschaftenaar:
e.g. the checkbox labelText is limited to be used with the string proptypes, so there will be warnings when passing in a number, or a valid component. 

As no modifications happen on the props and this is only something that gets rendered down the line, why limit its use? Use the proptype.node instead of proptype.string, proptype.number to allow anything that can be rendered to be passed in.

Everywhere in the code where props are simply being rendered without being modified, change the proptype from e.g. proptype.string, proptype.number to proptype.node

### DIFF
--- a/components/Checkbox.js
+++ b/components/Checkbox.js
@@ -9,7 +9,7 @@ const propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
-  labelText: PropTypes.string,
+  labelText: PropTypes.node,
   onChange: PropTypes.func,
   iconDescription: PropTypes.string,
 };


chrisdhanaraj:
Heyo could you update your branch with the latest from master? Good by me otherwise though

joshblack:
Went ahead and kicked off the CI build with the new branch 👍

marijohannessen:
👍 ✅   Good to merge once the branch is updated

joshblack:
@marijohannessen should the commit follow our commit conventions? Not sure what needs to be used to trigger the publish step that @chrisdhanaraj has setup

chrisdhanaraj:
Yeah the commit does, but on merge you can modify what it looks like. Unfortunately husky only kicks out commits done via command line

Issue #29, opened 2017-06-14
DropdownItem exposes onClick but is useless with Dropdown


eddiemonge:
DropdownItem's docs and code has onClick as a prop but it isn't really. Dropdown's onChange is used instead.

From @chrisdhanaraj:
> Still onChange - it's designed to mimic select/textinput/other inputs

DropdownItem does actually have onClick and uses it. However, Dropdown overwrites it with its own event handler rendering anything the user puts pointless

joshblack:
Hi there! 👋 

Thanks so much for reporting this issue. I went ahead and added a simple PR for this in #272, let me know if it doesn't address everything that you've mentioned, or if the behavior doesn't match what you're expectation is for this component!

PR #272, opened 2017-10-11
fix(Dropdown): call onClick for items if one exists

joshblack:
Closes carbon-design-system/carbon-components-react#29

Bugfix PR for #29, didn't know if we wanted to support this officially but figured I'd PR it to start the discussion 😄 

#### Changelog

**New**

**Changed**

- `Dropdown`
  - `children` in Dropdown are now cloned with an `onClick` handler that invokes the child's `onClick` prop (if it exists) and then calls `handleItemClick`.
- `Dropdown-test`
  - Added a simple test to verify the above behavior

**Removed**

### DIFF
--- a/components/Dropdown.js
+++ b/components/Dropdown.js
@@ -91,7 +91,10 @@ class Dropdown extends PureComponent {
 
     const children = React.Children.map(this.props.children, child =>
       React.cloneElement(child, {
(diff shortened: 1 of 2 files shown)


marijohannessen:
👍 ✅

PR #273, opened 2017-10-11
feat(CodeSnippet): add wrappedContentRef to access <pre> contents

joshblack:
Closes carbon-design-system/carbon-components-react#55

Updates `CodeSnippet` to include a new prop, `wrappedContentRef`, that allows access to the `<pre>` element in the component (making it easier to copy!).

#### Changelog

**New**

**Changed**

- `CodeSnippet`
  - Added `wrappedContentRef` as a new prop
- `CodeSnippet-test`
  - Added test to assert ref is called and passes the correct node

**Removed**

### DIFF
--- a/components/CodeSnippet.js
+++ b/components/CodeSnippet.js
@@ -8,13 +8,21 @@ const propTypes = {
   className: PropTypes.string,
   children: PropTypes.string,
   onClick: PropTypes.func,
+  wrappedContentRef: PropTypes.func,
 };
 
 const defaultProps = {
   type: 'terminal',
 };
 
-const CodeSnippet = ({ className, type, children, onClick, ...other }) => {
+const CodeSnippet = ({
+  className,
+  type,
+  children,
+  onClick,
+  wrappedContentRef,
+  ...other,
+}) => {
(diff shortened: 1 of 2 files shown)


jzhang300:
I don't know how I feel about using `ref`, that's going away isn't it?

joshblack:
They shouldn't be, unless I missed something substantial 😱 

Do you have a reference that says that @jzhang300 ?

joshblack:
Just as a side-note for reviewers, this PR definitely doesn't have to be merged in. The requesting issue could 100% be solved a different way (I believe), especially since they have access to the content itself so setting up a click handler to pass that in should be trivial.

tw15egan:
The old format of using ref as a string has been deprecated, but it looks like this supports the format of passing in a function e.g `ref={node => this.node = node}`

https://reactjs.org/docs/refs-and-the-dom.html

tw15egan:
I'm good with this 👍 ✅

marijohannessen:
Looks good to me too! 👍 ✅  Can you update your branch so it's up to date before merging?

PR #277, opened 2017-10-11
Value and Label are put in different span elements to allow for more …

miramurali23:
In the Slider component, value and label were in the same span element..due to this it was difficult to style them based on my needs such as spacing, and having both the value and label on the same line. With them being separated on two span elements...it is easier to style in css.


tw15egan:
👍 ✅

marijohannessen:
👍 ✅

PR #282, opened 2017-10-13
Update enzyme to 3.x

joshblack:
Updates our enzyme dependency to `3.x`, including an adapter for React 15 for now but should be updated with React 16 when we update the peer dependencies for Carbon.

Most of this PR has to deal with changes to how enzyme works with the new update and fixing the tests that were broken as a result.

#### Changelog

**New**

**Changed**

- `package.json`
  - `yarn test` now refers to the `jest` script, so instead of `yarn test-watch` one would do `yarn test --watch` and yarn will pass through the args to `jest`.
  - Added/Updated the following devDeps:
    - `enzyme`
    - `enzyme-to-json`
    - `enzyme-adapter-react-15`
  - `yarn check` now is used for CI and runs linting, testing, and SSR tests
- `.travis.yml`
  - Now uses `yarn` instead of `npm`
  - Now runs `yarn check`

**Removed**


components/__tests__/AccordionItem-test.js
```diff
@@ -51,7 +51,7 @@ describe('AccordionItem', () => {
       const item = toggler.find('li');
       expect(item.hasClass('bx--accordion__item--active')).toEqual(false);
       toggler.setState({ open: true });
-      expect(item.hasClass('bx--accordion__item--active')).toEqual(true);
+      expect(toggler.find('li').hasClass('bx--accordion__item--active')).toEqual(true);
```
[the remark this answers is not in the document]
joshblack: @eddiemonge yeah, enzyme@2 and before would operate on the data structures in memory, as a result the references were always fresh (afaik).

With enzyme@3, it seems like there is a new intermediate representation of the tree that requires `find` references to be updated.

components/__tests__/CardContent-test.js
```diff
@@ -64,7 +64,7 @@ describe('CardContent', () => {
         it('has expected links', () => {
           const links = title.childAt(1);
           expect(links.length).toBe(1);
-          expect(links.node.type).toBe('a');
+          expect(links.getElement().type).toBe('a');
```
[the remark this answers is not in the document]
joshblack: Honestly was just a codemod I ran on the codebase, if that's an alias I can definitely convert to it 😄

components/__tests__/Checkbox-test.js
```diff
@@ -50,16 +50,16 @@ describe('Checkbox', () => {
       });
 
       describe('input', () => {
-        const input = wrapper.find('input');
+        const input = () => wrapper.find('input');
```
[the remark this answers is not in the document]
joshblack: Might have been more appropriate to change to `getInput` as the method, but basically helps you find the latest references for the `.find` method after the intermediate representation is updated.

You can see the errors that doing this solves when you revert back these changes and run the tests.

components/__tests__/Checkbox-test.js
```diff
@@ -70,23 +70,23 @@ describe('Checkbox', () => {
       <Checkbox id="test" disabled />
     );
 
-    const input = wrapper.find('input');
-    expect(input.props().disabled).toEqual(true);
+    const input = () => wrapper.find('input');
```
jlengstorf: It does seem like DRYing this out with a helper function might be worth it, given that it's being used in multiple places.

Something like:

```js
const getInput = wrapper => wrapper.find('input');
```

And then:
```js
expect(getInput(wrapper).props().defaultChecked).toBeUndefined();
```

components/__tests__/CopyButton-test.js
```diff
@@ -60,12 +60,12 @@ describe('CopyButton', () => {
   describe('Renders feedback as expected', () => {
     it('Should make the feedback visible', () => {
       const feedbackWrapper = mount(<CopyButton feedback="Copied!" />);
-      const feedback = feedbackWrapper.find('.bx--btn--copy__feedback');
-      expect(feedback.hasClass('bx--btn--copy__feedback--displayed')).toBe(
+      const feedback = () => feedbackWrapper.find('.bx--btn--copy__feedback');
```
jlengstorf: Given that this pattern is coming up across files, it could potentially be even more DRY:

```js
const getBySelector = (sel, wrapper) => wrapper.find(sel);

// Or, curried:
const getBySelector = sel => wrapper => wrapper.find(sel);
```

The curried version would let you do:
```js
const getInput = getBySelector('input');
expect(getInput(wrapper))...
```

Might be overkill, but could also be a handy way to consolidate this kind of lookup logic in case the API changes again in enzyme@4.

components/__tests__/Dropdown-test.js
```diff
@@ -103,32 +103,32 @@ describe('Dropdown', () => {
 
     it('should add the open dropdown class on click', () => {
       dropdown.simulate('click');
-      expect(dropdown.hasClass('bx--dropdown--open')).toEqual(true);
+      expect(wrapper.find('.bx--dropdown').hasClass('bx--dropdown--open')).toEqual(true);
```
jlengstorf: `getBySelector` could help here, too.

components/__tests__/RadioButtonGroup-test.js
```diff
@@ -26,30 +26,30 @@ describe('RadioButtonGroup', () => {
 
       it('sets disabled attribute if disabled prop is set', () => {
         wrapper.setProps({ disabled: true });
-        expect(div.props().disabled).toEqual(true);
+        expect(wrapper.first('dev').props().disabled).toEqual(true);
```
jlengstorf: Is this a typo of `div`?

components/__tests__/RadioButtonGroup-test.js
```diff
@@ -26,30 +26,30 @@ describe('RadioButtonGroup', () => {
...
+        expect(radioButton().length).toEqual(2);
       });
 
       it('should set checked property based on defaultSelected prop', () => {
-        expect(radioButton.last().props().checked).toEqual(true);
+        expect(radioButton().last().props().checked).toEqual(true);
       });
 
       it('should set checked property based on valueSelected prop', () => {
         wrapper.setProps({ valueSelected: 'male' });
```
jlengstorf: Not really related to this PR, but could we use a different example? Enforcing binary gender selection is problematic, and while I realize it's just for testing, it's a pattern we should get out of the habit of using.

config/jest/setup.js
```diff
@@ -6,6 +6,12 @@ jest.unmock('object-assign');
 
 require('../polyfills');
 
+const enzyme = require.requireActual('enzyme');
+const Adapter = require.requireActual('enzyme-adapter-react-15');
+
+// Configure enzyme to work with React 15 for now
+enzyme.configure({ adapter: new Adapter() });
```
jlengstorf: Is there a way to get a list of compatibility problems from Enzyme so we can make an issue to track remaining fixes?

components/__tests__/RadioButtonGroup-test.js
joshblack: I don't even think `first` tags in args so definitely my bad 😅

config/jest/setup.js
joshblack: This is just for this PR, when we swap to React 16 it'll just use `enzyme-adapter-react-16` 😄

components/__tests__/RadioButtonGroup-test.js
joshblack: Agreed! Will take a todo to update tests like this.

marijohannessen:
👍 ✅

PR #284, opened 2017-10-16
feat(notifications): support any node for subtitle & caption

jamesvclements:
Closes carbon-design-system/carbon-components-react#256

#### Changelog

**New**

- Allows an element as the subtitle and or caption for notifications

### DIFF
--- a/components/Notification.js
+++ b/components/Notification.js
@@ -59,8 +59,8 @@ class NotificationButton extends Component {
 class NotificationTextDetails extends Component {
   static propTypes = {
     title: PropTypes.string,
-    subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
-    caption: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+    subtitle: PropTypes.node,
+    caption: PropTypes.node,
     notificationType: PropTypes.oneOf(['toast', 'inline']),
   };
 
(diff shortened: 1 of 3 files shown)


components/Notification.js
```diff
@@ -101,9 +101,9 @@ class ToastNotification extends Component {
     className: PropTypes.string,
     kind: PropTypes.oneOf(['error', 'info', 'success', 'warning']).isRequired,
     title: PropTypes.string.isRequired,
-    subtitle: PropTypes.string.isRequired,
+    subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
joshblack: Would it be helpful instead for these props to just be `PropTypes.node`?

joshblack:
Hi @jamesvclements! 👋 

Just wanted to say thanks so much for taking the time to contribute to `carbon-components-react` 🙏  we really appreciate any and all contributions.

Since it's past our typical working hours right now, will try and get this in front of some eyes tomorrow for reviews from some of the core members, in addition to anyone else who has the time to take a look 😄

components/Notification.js
jamesvclements: @joshblack Yep that makes more sense. I'll swap them

tw15egan:
Thanks @jamesvclements for tackling this! 👍 ✅

PR #297, opened 2017-10-27
fix(pagination): make sure if total pages = 0, the page total cannot …

tw15egan:
Makes sure if `totalPages` = 0, the right arrow will not increment the page

Closes https://github.com/carbon-design-system/carbon-components-react/issues/295


components/Pagination.js
```diff
@@ -168,7 +169,8 @@ class Pagination extends Component {
           <button
             className="bx--pagination__button bx--pagination__button--backward"
             onClick={this.decrementPage}
-            disabled={this.props.disabled || statePage === 1}>
+            disabled={this.props.disabled || statePage === 1}
```
joshblack: Want to add a quick test for this case and the case below? Just to help protect against any regressions.

tw15egan: 👍

PR #301, opened 2017-10-31
fix: Modify default props values

codymwalker:
Closes carbon-design-system/carbon-components-react#299
Closes https://github.com/carbon-design-system/carbon-components-react/issues/168

Modifies default values for placeholder and label text. This allows product teams to decide if they want to use certain props, without forcing a default message.

### Changelog

**Changed**

- Changed default label/placeholder text in a number of models.
- Added tests to account for defaults.

### DIFF
--- a/src/components/Toolbar/Toolbar-test.js
+++ b/src/components/Toolbar/Toolbar-test.js
@@ -61,6 +61,20 @@ describe('Toolbar', () => {
       rootWrapper.props().onClickOutside();
       expect(rootWrapper.state().expanded).toEqual(false);
     });
+
+    describe('should use default props', () => {
+      const defaultToolbarSearch = shallow(
+        <ToolbarSearch />
+      );
+
+      it('should use default blank placeholder text', () => {
+        expect(defaultToolbarSearch.find('.bx--search-input').props().placeholder).toBe('');
+      });
+
+      it('should not attach a label', () => {
+        expect(defaultToolbarSearch.find('label').length).toBe(0);
+      });
+    });
   });
 
   describe('ToolbarItem with an overflow menu', () => {
--- a/src/components/ToolbarSearch/ToolbarSearch.js
+++ b/src/components/ToolbarSearch/ToolbarSearch.js
@@ -18,8 +18,8 @@ export default class ToolbarSearch extends Component {
   static defaultProps = {
     type: 'search',
     id: 'search__input',
-    labelText: 'Provide labelText',
-    placeHolderText: 'Provide placeHolderText',
+    labelText: '',
+    placeHolderText: '',
     role: 'search',
   };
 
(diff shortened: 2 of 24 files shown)


joshblack:
If these props are optional, should we also handle toggling the associating markup when the prop isn't being passed in?

tw15egan:
@joshblack which components are you referring to?

joshblack:
@tw15egan essentially any component that has associated markup with the optional prop.

For example, the `title` prop for `AccordionItem` has the corresponding markup that shouldn't be generated if the prop is not included:

```jsx
<p className="bx--accordion__title">{title}</p>
```

codymwalker:
Yep. I agree with that @joshblack. It actually looks like a couple of other components handle it like that already now that I look at it.

marijohannessen:
Hey @codymwalker , just checking in here. Are you in the process of updating your PR with the discussed changes?

codymwalker:
@marijohannessen Updated!

src/components/ToolbarSearch/ToolbarSearch.js
```diff
@@ -60,9 +60,11 @@ export default class ToolbarSearch extends Component {
     return (
       <ClickListener onClickOutside={this.handleClickOutside}>
         <div className={searchClasses} role={role}>
-          <label htmlFor={id} className="bx--label">
-            {labelText}
-          </label>
+          {labelText && (
```
joshblack: Is this a potential required prop since it provides the `htmlFor` attribute on the label for the `input`? Might fail a11y checks otherwise.

src/components/Toolbar/Toolbar-test.js
```diff
@@ -61,6 +61,20 @@ describe('Toolbar', () => {
...
       expect(rootWrapper.state().expanded).toEqual(false);
     });
+
+    describe('should use default props', () => {
+      const defaultToolbarSearch = shallow(
+        <ToolbarSearch />
+      );
+
+      it('should use default blank placeholder text', () => {
+        expect(defaultToolbarSearch.find('.bx--search-input').props().placeholder).toBe('');
```
joshblack:
```diff
-expect(defaultToolbarSearch.find('.bx--search-input').props().placeholder).toBe('');
+expect(defaultToolbarSearch.find('.bx--search-input').prop('placeholder')).toBe('');
```

Works too! 😄

src/components/ToolbarSearch/ToolbarSearch.js
codymwalker: We could go either way. I think it really depends on whether you want to force service teams to implement the labels or leave it up to them to handle whether or not to implement proper accessibility standards.

marijohannessen:
Closing this as it's been taken over by #382

PR #302, opened 2017-11-01
fix(FloatingMenu): support React 16 for floating menu

### LINKED ISSUE #300: Tooltip not working in React 16
TypeError
Cannot read property 'getBoundingClientRect' of null

How to reproduce:  import the Tooltip and copy the code from Storybook and try to render.
https://codesandbox.io/s/xv4w4yx0oz

asudoh:
Closes carbon-design-system/carbon-components-react#300.

Floating menu has circumvented React's rendering system as the place in DOM goes out of React render tree by nature. React 16 broke such approach by its "circumventing" nature, but introduced an API to officially introduced an API to render to alternate tree - `ReactDOM.createPortal()`. This PR makes use of such new API to fix #300.

#### Changelog

**Added**

- Code to clean up `this.menu`, etc. once it's no longer valid

**Changed**

- Moved the main rendering logic from `.renderLayer()` method to regular `.render()` method, in React 16
- Introduced states used in React 16, to trigger `.render()` as positioning info gets available
- Moved `.componentDidMount()` code to `ref` callback to more directly work with the DOM ref

### DIFF
--- a/internal/FloatingMenu.js
+++ b/internal/FloatingMenu.js
@@ -18,10 +18,25 @@ class FloatingMenu extends React.Component {
     menuDirection: 'bottom',
   };
 
-  componentDidMount() {
-    if (this.doc) {
-      this.menu = this.doc.createElement('div');
-      this.menu.ownerDocument.body.appendChild(this.menu);
+  constructor() {
+    super();
+    this.state = {};
+  }
+
+  onNewMenuRef = menu => {
+    if (!menu) {
+      return;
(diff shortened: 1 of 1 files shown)


shjunmiao:
Thanks for fixing it so quickly. Which version of carbon react would this change be included?

tw15egan:
@shjunmiao as soon as we get one more review, we'll publish a release `v4.0.3`

PR #303, opened 2017-11-02
[RFC] Update folder structure

joshblack:
Updates the project folder setup to match some updated conventions from more recent work on the addons repos. This re-organization namely revolves around moving tests, stories, and components into dedicated folders.

In addition, this updates some conventions for built assets, namely a change from `cjs` -> `lib` for the default export. The rationale is that the default export, i.e. the `main` field for a package shouldn't be tied to a specific module type, in this case `commonjs`, as that could change as the ecosystem changes.

#### Changelog

**New**

**Changed**

- Moved `components`, `internal` and `lib` under `src`
- Renamed `lib` to `tools`
- Add `engines` field to `package.json` with support `node` and `yarn` ranges
- Update storybook to load stories from `src/components`
- Move components from `components` folder to their own unique `ComponentName` folder
  - Each file is then located at `src/components/ComponentName/ComponentName.js` with a helper `index.js` to make require paths less redundant
- Move tests from `__tests__` to `<ComponentName>/<ComponentName>-test.js`
- Move stories from `.storybook/components` to `<ComponentName>/<ComponentName>-story.js`
- Update `scripts/build.js` with the new changes, default to quiet mode instead of logging out each file transformation

**Removed**


joshblack:
Would love any comments from both the team and external contribs looking at this 😄 

This is totally just an RFC, wanted to do a quick PR to show a potential change to how we organize files across our React projects in order to help out with consistency.

tw15egan:
👍 ✅

Issue #37, opened 2017-06-15
InteriorLeftNavList is not open when one of the items in active and when open={true}


ashpc:
InteriorLeftNavList is not open when one of the items is active and also when open is set to true.

When we click the back button in the browser, the  **open** props is set to true but the InteriorLeftNavList is still closed.

dcostantino:
Console is showing log when trying to open:
 Uncaught TypeError: _this.props.children.forEach is not a function
    at Object.InteriorLeftNav._this.handleListClick [as onListClick] (InteriorLeftNav.js:53)
    at InteriorLeftNavList._this.toggle (InteriorLeftNavList.js:38)
    at Object.ReactErrorUtils.invokeGuardedCallback (ReactErrorUtils.js:69)
    at executeDispatch (EventPluginUtils.js:85)
    at Object.executeDispatchesInOrder (EventPluginUtils.js:108)
    at executeDispatchesAndRelease (EventPluginHub.js:43)
    at executeDispatchesAndReleaseTopLevel (EventPluginHub.js:54)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (forEachAccumulated.js:24)
    at Object.processEventQueue (EventPluginHub.js:254)

dcostantino:
Error is happening when there is a single list item

marijohannessen:
@dcostantino did this issue get resolved?

dcostantino:
Just tested, seems to be resolved. Thanks!

ehdsouza:
This is still an issue:
![carbon](https://user-images.githubusercontent.com/13935994/32186514-a0739776-bd78-11e7-909e-aa907a832d38.gif)


Also, tried updating to 3.14 and 4.0 but still the same issue

joshblack:
Yikes, sorry about that @ehdsouza! To help track down/isolate the issue, can one of the people reporting the issue re-create it in https://codesandbox.io/?

ehdsouza:
Created a PR https://github.com/carbon-design-system/carbon-components-react/pull/312 that supports keeping the list open at all times. Please help review it

marijohannessen:
Closing this issue due to our inactivity policy. Please feel free to re-open if needed.

PR #312, opened 2017-11-06
Feat: Add lockOpen prop to InteriorLeftNavList

ehdsouza:
The lockOpen prop with property open helps to keep the InteriorLeftNavList expanded all the time. It
would also have the chevron-down hidden.

#### Changelog

**New**

- lockOpen prop in InteriorLeftNavList


joshblack:
Hi there! 👋 

Thanks for taking the time to make this PR 😄 

In the case of #37, do we need to support a new prop, in this case `lockOpen`, or do we need to fix the underlying problem that was noted in the issue? In other words, is the intended behavior what is currently happening and we want to support an additional prop to change it, or is this fixing the existing behavior that is seen as incorrect?

ehdsouza:
This PR is independent of issue 37.  In our team's use case we want to display all the nested menu items all the time.  So the solution was to have a `lockOpen` property to allow this functionality.

src/components/InteriorLeftNavList/InteriorLeftNavList.js
```diff
@@ -68,14 +72,15 @@ class InteriorLeftNavList extends Component {
...
+      lockOpen,
       ...other
     } = this.props;
 
     const classNames = classnames(
       'left-nav-list__item',
       'left-nav-list__item--has-children',
       {
-        'left-nav-list__item--expanded': this.state.open,
+        'left-nav-list__item--expanded': lockOpen || this.state.open,
```
asudoh: Hi @ehdsouza nice addition! - Given this PR adds logic to prevent `state` from being changed, ignoring the state here looks redundant?

ehdsouza: 👍 Good catch, Ill make the change

marijohannessen:
Hey, thanks for making this PR! It's looking great so far. I have a couple of comments.

1. I'm a bit wary about the styling of the parent link item when its list is locked open, cause it still has all the styles of a link but I can't interact with it. So it would be great if you can remove any active/hover states from the parent link item when the `lockOpen` prop is present.
2. Could you look into making this functionality its own component instead of integrating it into this component? You can take a look at the `TableRowExpanded` component to see an example of what I mean. So instead of having a prop added to the `InteriorLeftNavList`, you would create a separate component named for example `ExpandedNavList`. Just let me know if you have any questions!

ehdsouza:
Thanks for reviewing the PR. 
For comment 1:
Thanks for pointing it out. I was thinking of proposing the following solution as below. Though could you let me know how I could add the new css?
```
        {
          lockOpen ? <div style={{
            padding: '0.5625rem 1.25rem',
            fontSize: '0.875rem',
            cursor: 'default',
          }} className="left-nav-list__item-label">{title}</div> :
          <a className="left-nav-list__item-link">
            {title}
            <div className="left-nav-list__item-icon">
              <Icon
                name="chevron--down"
                description={iconDescription}
                className="left-nav-list__item-icon bx--interior-left-nav__icon"
              />
            </div>
          </a>
        }
```
For comment 2:
There is just a new prop that is needed to handle this functionality. Adding a new component would seem redundant.

Let me know your thoughts

src/components/InteriorLeftNavList/InteriorLeftNavList.js
asudoh: Great thanks @ehdsouza!

joshblack:
@ehdsouza for the style, this is something that we'd need to coordinate upstream with `carbon-components` and the core design team as I believe it's an un-intended pattern (but would need to verify).

Regardless, additional CSS would need to be merged up-stream on `carbon-components` and then consumed in this  package 😄 

For the second point, definitely agreed that there will be shared code! However, the changes in this PR suggest static behavior that could be created separately from the dynamic behavior of `InteriorLeftNavList`.

More specifically, if the list no longer needs to be expandable/collapsible, then one can remove any of the click or key handlers required to construct the list itself to handle the collapse behavior and any of the state inherent in the component. This makes it a great candidate for a Stateless Functional Component 😄

PR #314, opened 2017-11-07
FormLabel and FormItem components

marijohannessen:
This is part of resolving https://github.com/carbon-design-system/carbon-components-react/issues/115

There seems to be a need to have the label and form items more customizable so users can add for example a tooltip to a form label. This PR starts this conversation and proposes a solution by adding in `FormLabel` and `FormItem` components.

If this is something we should move forward with then this will mean a refactor of all form elements - which could potentially mean a breaking change? Or we can add it in as a separate variation as well. Curious to hear people's thoughts!

A CSS fix is also present in the vanilla repo here: https://github.com/carbon-design-system/carbon-components/pull/410


joshblack:
Seems super useful to me! 

Semi-related to this PR, one pattern that I see emerging is a lot of components that just apply a class prefix to a container node (typically a `div`, but can change) and then just renders any children. Seems like a good opportunity to generalize in the sense that we could have something like:

```
import cx from 'classnames';
import React from 'react';

const Prefix = ({ type, prefix, className, children, ...other }) => {
  return React.createElement(
    type,
    {
      ...other,
      className: cx(prefix, className),
    },
    ...children
  );
};

Prefix.propTypes = { /* ... */ };

export default Prefix;

// Usage in `Formitem`
import React from 'react';
import Prefix from '../Prefix';

const FormItem = props => (
  <Prefix type="div" prefix="bx--form-item" {...props} />
);

export default FormItem;
```

Might not be too useful, but does help consolidate/standardize things like `propTypes` on these kinds of components while also providing the flexibility to break out if needed.

What do you think @marijohannessen? cc @tw15egan

tw15egan:
Very useful 😄 Changes look good to me, can eventually start updating all components to follow this convention

👍 ✅

PR #323, opened 2017-11-10
fix(Slider): control when updates happen better

eddiemonge:
componentWillReceiveProps was doing a check on object vs object. Even if
the objects had the same props, it would always return false because
JavaScript. Doing a very simple toString-like check is a better check

updatePosition changed too much state and fired onChange too much as
well. It would fire on every mouse move, even if the value didn't
actually change


src/components/Slider/Slider.js
```diff
@@ -43,7 +43,7 @@ export default class Slider extends PureComponent {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps !== this.props) {
+    if (JSON.stringify(nextProps) !== JSON.stringify(this.props)) {
```
joshblack: Instead of doing a `JSON.stringify` call here, could we use just do direct comparison on the props?

joshblack: I just meant something similar to lodash's `isEqual` method, or fbjs's `areEqual` method. I didn't realize how large the props surface area is for this component though, so my recommendation seems completely inadequate haha.

asudoh: Good discussion here, great that we seriously consider performance! There are some other considerations with walking through objects with `JSON.stringify()` API:

* `JSON.stringify()` throws if the given object contains circular references (Step 1 in: http://www.ecma-international.org/ecma-262/6.0/#sec-serializejsonobject)
* `JSON.stringify()` ignores function refs and symbols (Step 11 and 12 in: http://www.ecma-international.org/ecma-262/6.0/#sec-serializejsonproperty)

One approach we may take is using [`replacer` callback](http://www.ecma-international.org/ecma-262/6.0/#sec-json.stringify), where we can at least relegate walking part to the native API. Walking though and comparing object properties all by JavaScript may be slow especially with non-modern browsers, while IIUC modern browsers has made basic ops (esp. if they are non-layout-affecting and non-memory-instensive) very fast.

marijohannessen:
Hey @eddiemonge , could you update this PR so we can get it reviewed and merged in? Thank you!

eddiemonge:
Updated

joshblack:
For checking prop equality, we'd want to use http://npmjs.com/package/lodash.isequal versus `JSON.stringify`, I think.

eddiemonge:
Updated to use lodash.isEqual

tw15egan:
Works great, thanks for this fix @eddiemonge! 👍 ✅

PR #332, opened 2017-11-10
Introduce generic ListBox component used in Combobox, MultiSelect, and Dropdown

joshblack:

Adds in a generic `ListBox` component that can be used for things like `MultiSelect`, `Combobox`, and more.

Preview URL: https://goo.gl/Z5VZPp

#### Changelog

**New**

- `ListBox` and `ListBox*` components
- `MultiSelect` component
- `Combobox` component
- `DropdownV2` component

**Changed**

**Removed**


asudoh:
Great to see progress here @joshblack! Is the code at the stage where I can try out the features? I’m getting the following (when I go to Combobox - Pre-selected, type something, then move focus out of input box) and wondering if I’m doing something stupid:

```
Uncaught TypeError: Cannot read property 'label' of null
    at Object.defaultItemToString [as itemToString] (preview.bundle.js:84940)
    at preview.bundle.js:86534
    at Downshift.<anonymous> (preview.bundle.js:85908)
    at ReactCompositeComponentWrapper._processPendingState (preview.bundle.js:42882)
    at ReactCompositeComponentWrapper.updateComponent (preview.bundle.js:42825)
    at ReactCompositeComponentWrapper.performUpdateIfNecessary (preview.bundle.js:42768)
    at Object.performUpdateIfNecessary (preview.bundle.js:4591)
    at runBatchedUpdates (preview.bundle.js:2286)
    at ReactReconcileTransaction.perform (preview.bundle.js:6907)
    at ReactUpdatesFlushTransaction.perform (preview.bundle.js:6907)
```

Another thing I found is that I see many changes in this PR that seem irrelevant to combobox. It’d be great if you can keep your Prettier setting in sync with the one in latest `master`, in case you haven’t done it already. Thanks!

joshblack:
Hi @asudoh! Thanks for checking it out. This was me initially bringing work over from `carbon-addons-beta-react`, I have some work locally that I'll push up after a design sync today 😄 

With respect to prettier, because the `lint-staged` hooks aren't hooked up files haven't been run against prettier. Since I ran the script manually, it triggered these files to appear in the diff.

joshblack:
Have a URL to view the behavior of combobox that we'll try to match 😄  (if you're curious): https://carbon-addons-beta-storybook-hotfix.mybluemix.net

asudoh:
Thanks @joshblack for more info! Great to see that `Typeahead` component works pretty well on `carbon-addons-beta-storybook-hotfix.mybluemix.net`. The only one thing I found there was that the dropdown does not open when the text box is focused initially, but when I type a character and then hit Backspace the dropdown opens. Anyhow looking forward to see `TypeAhead` code in this PR working!

Also great to see `MultiSelect` is working pretty well. Will “Clear Selection” button be added soon?

BTW thank you for clarifying the Prettier thing. Going forward, it’d be nice if you can put extra care on managing diffs. If I understand correctly on what you said about Prettier, you can stage only relevant files after running Prettier before checking stuffs in, and stage all other files and then create another commit. In this way, you can remove the latter commit e.g. with `git rebase -i` (and cherry-pick the latter commit to another branch) before creating a PR. Thanks!

src/tools/testing/items.js
```diff
@@ -0,0 +1,23 @@
+// `GenericItem` corresponds to an item in a collection that is passed to
```
joshblack: Maybe these should be under `internal` instead?

src/components/Combobox/Combobox.js
```diff
@@ -0,0 +1,303 @@
+// import cx from 'classnames';
```
asudoh: Should we remove this file for this PR? Same for the corresponding test and styles.

src/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,220 @@
...
+    // Generic `label` that will be used as the textual representation of what
+    // this field is for
+    label: PropTypes.node.isRequired,
+  };
+
+  static defaultProps = {
+    type: 'default',
+  };
+
+  static defaultProps = {
```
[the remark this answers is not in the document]
joshblack: hahaha oops 😄

src/components/Combobox/Combobox.js
joshblack: Depending on how much I can chip away at ComboBox before the sprint ends, I might just remove those files altogether and save them for another PR 😄

joshblack:
@asudoh 

> Keeping highlight when hitting Enter key

Agreed, this was a big complaint when I was working with it. We need to find a way to control the `highlightedIndex` value on `downshift` so that it doesn't just reset on each selection.

RE: the input notes you left, I was hoping `getItemProps` would take care of it but it doesn't seem like it. Maybe I'm using it incorrectly?

src/components/Combobox/Combobox.js
asudoh: @joshblack Yeah this is something I had in mind - Great if this PR can be splitted, one for `MultiSelect` and one for filtering select.

src/components/ListBox/ListBox.js
```diff
@@ -8,6 +8,7 @@ const ListBox = ({
   type = 'default',
   items,
   selectedItem,
+  selectItem,
```
joshblack: Need to remove this line, it's no longer used!

src/components/ListBox/ListBox.js
```diff
@@ -36,6 +37,7 @@ const ListBox = ({
       {isOpen && (
         <ListBoxMenu
           items={items}
+          selectItem={selectItem}
```
joshblack: Same as above

src/components/ListBoxMenu/ListBoxMenu.js
```diff
@@ -3,8 +3,15 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import Icon from '../Icon';
 
+const handleOnClick = event => {
+  // Prevent default interaction here so that the event bubblers up to
```
joshblack: bubblers -> bubbles

src/components/ListBoxMenu/ListBoxMenu.js
```diff
@@ -0,0 +1,66 @@
...
+          'bx--list-box__menu-item': true,
+          'bx--list-box__menu-item--highlighted': highlightedIndex === i,
+          'bx--list-box__menu-item--active': selectedItem.indexOf(item) !== -1,
+        })}
+        {...getItemProps({
+          item,
+          index: i,
+        })}>
+        <input
+          id={item.id}
```
asudoh: @joshblack We need to think about multiple instances of `<MultiSelect>` - What if two of `<MultiSelect>`s have items with the same ID? We’d need to assign unique ID per instance (e.g. https://github.com/asudoh/carbon-components-react/blob/74c19d33f76a64428e97726f514b1e78bde786e6/src/components/ListBoxMenu/ListBoxMenu.js#L7-L11) to make the ID here unique. Thanks!

joshblack: Fixed! Just am going to use the id provided from `getItemProps` 😄

asudoh:
Thanks for fixing things, @joshblack! Some issues I found while I fiddled with your latest code (though neither of them seems new) were:

* I cannot open the menu with down arrow key once it’s opened and closed (Spacebar works, though)
* The focus seems orphaned after
  * Hitting Enter on the badge (or clicking on the badge)
  * Clicking on check box

It’d be great if you have quick fixes for those - Thanks!

src/components/ListBoxField/ListBoxField.js
```diff
@@ -0,0 +1,41 @@
...
+  selectedItem,
+  clearSelection,
+  isOpen,
+  // `rest` here should correspond with `getButtonProps` from `downshift`
+  ...rest
+}) => {
+  const className = cx({
+    'bx--list-box__field': true,
+    'bx--list-box__field--inline': type === 'inline',
+    'bx--list-box__field--open': isOpen,
```
asudoh: Great to consolidate `--open` state class to the top-level list box component. In this way, we don't need to have multiple component managing `--open` state class.

joshblack:
Couple of TODOs:

- [ ] Update MultiSelect tests
- [ ] Update MultiSelect story
- [ ] Add complete stories for DropdownV2, MultiSelect, Combobox
- [ ] Surface `translateWithId` hooks through each component for i18n compat

bsonefeld:
Looking awesome! 🎉 A few UX bugs to address: 

- [ ] When I clear all from the Inline and regular dropdown, the drawer from the dropdown should not reopen (the number badge should disappear and the label should reposition) 

dropdown-bugs

- [ ] Was having trouble with keyboard tabbing through the multiselect (I couldn't get the drawer to open) . Using Chrome latest

- [ ] Can we add a transition to the hover? @tay-aitken maybe can you advise on what this should be?

hover

- [ ] User inputted text should be `semi-bold` instead of `regular`
(current shown below)
screen shot 2017-12-13 at 11 20 54 am


- [ ] Did we build out Multi-select + Filtering? 

- [ ] TBD after I chat with @tay-aitken , but noticed the hover on the drawers is different for multi and single select. Just want to make sure that was intended. 

screen shot 2017-12-13 at 11 26 31 am


## Naming
Just wanted to clarify a few naming things. I see ListBox, Combobox, and Dropdown. On the website we'll be displaying this component as the Dropdown, with multiselect and filter being variations of the dropdown. 

We want to keep that pretty in sync with how the code is structured naming wise. Does that still make sense?

joshblack:
Thanks @bsonefeld!

With respect to naming, ListBox is what powers the other components. The external API we surface to users should match the naming conventions you were talking about. The only concern here is that we overload `Dropdown` with too large of an API surface area, which would make it harder to use for first-time users. If we could find a way to name individual components that match the variations you're talking about, that would help out a ton. For example:

- Dropdown
- InlineDropdown
- MultiSelect
- InlineMultiSelect
- FilterableMultiSelect
- FilterableDropdown
- InlineFilterableDropdown

But I would totally get if these names aren't immediately intuitive.

Also, RE: multi-select + filtering it's on the way today! 😄

tay-aitken:
> Can we add a transition to the hover? @tay-aitken maybe can you advise on what this should be?

I would add the same transition as we have on buttons.

- [ ] Is appears the tag has a higher z-index than the dropdown drawer. The dropdown drawer should cover anything beneath it
screen shot 2017-12-13 at 1 22 26 pm
screen shot 2017-12-13 at 1 22 40 pm

- [ ] the check boxes don't seem centered in their respective rows, please center (only on multi-select versions)
screen shot 2017-12-13 at 1 20 32 pm
screen shot 2017-12-13 at 1 20 56 pm

- [ ]  it seems like the padding around the interior of the drawer is not equal on all sides, this may be due to the checkboxes + labels not being centered. 

screen shot 2017-12-13 at 1 36 49 pm

bsonefeld:
Sounds good. For reference, this is how we're referring to each bit of the Dropdown in the usage and style tabs. 

screen shot 2017-12-13 at 1 37 32 pm

joshblack:
@bsonefeld forgot to follow-up:

> Was having trouble with keyboard tabbing through the multiselect (I couldn't get the drawer to open) . Using Chrome latest

Just use spacebar!

tay-aitken:
## Focus states
 I noticed a few inconsistencies in focus states that need to be addressed. Currently the keyboard action between a plain component 9checkbox) and that same component, nested inside another component, are not the same. 

- [ ] for checkboxes, on their own, the keyboard action is to push `spacebar` to check and uncheck a box. We need to keep this consistent in the combobox, I believe this is also the inherited browser state. Currently it is using `enter`. 

However changing that keyboard action to be consistent with the component on its own, conflicts with the keyboard board action to open/close the dropdown(s), which is also `spacebar`. The old dropdown is using `spacebar`,`up and down arrows `, `enter` to open close the drawer. Select is uses `up and down arrows` and `spacebar` to open/close drawer. I think having three options for open/closing dropdown (four if you count `esc` for closing) is excessive.

Im thinking we use `enter` to open and close drawer so we can keep `spacebar` consistent with checkmarks. Thoughts @bsonefeld @joshblack 

- [ ] pushing `esc` open and closes the dropdown drawer, it should only close the dropdown drawer (only seems to be doing this for multi-select)


Regarding the comment above about hover states, they should all be the lighter gray color. They were inconsistent in my designs. It looks like the gray has been carried over properly here.

Also Josh how do you feel about adding motion to the drawer?
**Dropdown**
This occurs when the dropdown is clicked open/closed
- [x] opening (ease-out: 0, 0, 0.25, 1 )
- [x] closing (ease-in: 0.25, 0, 1, 1)
- [x] timing: 300ms

psachs21:
@joshblack is there any intention to support potentially async operations? As in, I have a large number of items that could appear but don't want them all rendered (for perf). Maybe it's an enhancement once this is done but I was thinking of something like, instead of provided an array of items, we just provide a getter (getItems) which is given the filter value and returns the items. We could do both really, for simple scenarios, `items`, for async, large datasets, `getItems`?

asudoh:
Refs https://github.com/carbon-design-system/carbon-components-react/issues/406

tay-aitken:
Hey Josh! Just took a look at this again and noticed two small errors. Please just check off if they have already been addressed but the link hasn't been updated. 

- [ ]  Inline multi-select - when an option is selected it is not floating to the top with the other selected items
inline multi-select

- [ ]  Multi-select (and eventually inline multi-select too once the above functionality is added) - Is there a way for the selected items to appear alphabetically? Right now they are appearing in the order the user selected them, from what I can tell. 
multi-select

joshblack:
Thanks @tay-aitken! Good catch on both, will fix them 😄

matuagarwal:
Hello everyone,

I am from IBM AppConnect Dev team (https://appconnect.ibmcloud.com/), we have recently moved to carbon and need this functionality( since a customer is urgently awaiting for this).
Any outlook by when this feature would be available ?

joshblack:
Hi @matuagarwal! Talking earlier, I think your team is the one is looking for MultiSelect, correct? We're splitting up this PR into several smaller ones. To track the work for MultiSelect, check out: https://github.com/carbon-design-system/carbon-components-react/pull/450 !

joshblack:
For those following this PR, we're going to be moving each of the big components introduced here into smaller PRs. We've already merged in a good chunk of the work from ListBox, and now we're going to be adding MultiSelect, Dropdown v2, and Combobox!

Going to close this one for now in preference of those PRs 😄

Issue #335, opened 2017-11-13
Provide a way to set time value programmatically in TimePicker component.


sandhya-r-reddy:
We want to update  time programmatically for certain use-cases. We are accessing internals of TimePicker components to achieve that. It would be great if these classes provide some mechanism to achieve that (via props or state).

PR #336, opened 2017-11-13
fix: Added a new prop 'value' in TimePicker Component .

sandhya-r-reddy:
Closes carbon-design-system/carbon-components-react #335

Added a new prop 'value' in TimePicker Component to set the value programmatically.

Change log
**Changed**
- Added new Prop value, to set the value programmatically
- Updated tests


tw15egan:
Seems like a nice, quick fix. I don't have any objections to it 👍 ✅

PR #337, opened 2017-11-13
fix(slider): set default input type to number

tw15egan:
Sets the default input type of the slider to `number`

Closes https://github.com/carbon-design-system/carbon-components-react/issues/278

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -277,7 +277,7 @@ export default class Slider extends PureComponent {
             <span>{max}</span>
             <span>{maxLabel}</span>
           </span>
-          {!hideTextInput ? (
+          {!hideTextInput && (
             <TextInput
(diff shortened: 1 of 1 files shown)


src/components/Slider/Slider.js
```diff
@@ -276,15 +277,15 @@ export default class Slider extends PureComponent {
...
-          {!hideTextInput
-            ? <TextInput
-                type={inputType}
-                id="input-for-slider"
-                className="bx-slider-text-input"
-                value={value}
-                onChange={this.handleChange}
-              />
-            : null}
+          {!hideTextInput ? (
```
joshblack: Since the cons case is just `null`, should we just:

```jsx
{!hideTextInput && (
  // ...
)}
```

tw15egan: works for me

PR #345, opened 2017-11-16
chore: update .gitignore

asudoh:


joshblack:
Hey @asudoh! We actually generate a `lib` folder instead of `cjs` now 😄

asudoh:
OK thanks you for clarifying @joshblack!

Issue #342, opened 2017-11-15
Change the kind of modal wrapper trigger button


chethankumar:
## Detailed description

Describe in detail the issue you're having. Is this a feature request (new component, new icon), a bug, or a general issue?

Its a feature request.

> Is this issue related to a specific component? 

Yes. ModalWrapper

> What did you expect to happen? What happened instead? What would you like to see changed?

I expected a prop which could change the kind of trigger button where we could specify the kind as primary/secondary/ghost/small etc

> What browser are you working in?

Chrome/firefox

> What version of the Carbon Design System are you using?

carbon-components : 8.1.8
carbon-components-react: 4.1.2

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Working on IBM App Launch service on bluemix

bjones526:
@tw15egan I've also thought this would be a handy feature. Mind if I take a look?

PR #346, opened 2017-11-16
feat(ModalWrapper): set kind for trigger button (#342)

bjones526:
Closes carbon-design-system/carbon-components-react#342

Allow product teams to pass in a prop to set kind for modalWrapper's trigger button 

#### Changelog

**New**

- pass props through modalWrapper to set kind for the trigger button
- default to primary button if no props passed


tw15egan:
Thanks for adding in this feature, @bjones526! Changes look great to me 😄 👍 ✅

PR #349, opened 2017-11-16
Added isExpanded prop in InteriorLeftNavList

ehdsouza:
For the InteriorLeftNav we need isExpanded behavior where we want the child lists to be kept opened unless closed by the user. Closing one child list shouldn't close the other opened lists.

fixes https://github.com/carbon-design-system/carbon-components-react/issues/328

#### Changelog

**New**

- Added props isExpanded in `InteriorLeftNavList`


ehdsouza:
@marijohannessen thanks for the review comments. Renamed `isAccordion` to `isExpanded`.

joshblack:
One question just to help catch me up, what's the motivation behind the `open` prop for the `NavList`?

ehdsouza:
@joshblack there is no `open` prop. Its `isExpanded` prop for the `InteriorLeftNav` for our console [watson console](https://console.bluemix.net/developer/watson/dashboard). So, when one list is opened, we dont want any other opened lists to close.

Look at the [328](https://github.com/carbon-design-system/carbon-components-react/issues/328) for a detailed explanation

joshblack:
Not sure if it was clear, but am talking about: `<InteriorLeftNavList className="second" open />`

ehdsouza:
The test case checks the behavior when a user is expanding one list, any other opened lists behavior is unaffected. So it still stays as what it was ie `open` when `isExpanded` is set to true

joshblack:
It's not really about the test case, regardless, what's the motivation behind having `isExpanded` on the container component versus on the child `NavList` components?

ehdsouza:
@joshblack I got your point. I added the isExpanded to the child component `InteriorLeftNavList`. Let me know if any more suggestions

germanattanasio:
when will this be release?

tw15egan:
@germanattanasio  Should be included with the latest release

joshblack:
@germanattanasio although it's not in the changelog,  it's included in the 4.2.2 tag: https://github.com/carbon-design-system/carbon-components-react/releases/tag/v4.2.2

PR #351, opened 2017-11-16
Added a new prop to set the date value programmatically, resolves #350

### LINKED ISSUE #350: Provide a way to set date value programmatically in DatePicker component.
We want to update the date  programmatically for certain use-cases. For example, if user selects Last 6 days from drop-down, we can change dates to match that selection. However, there is no easy way to do this right now.
We are accessing internals of DatePicker components to achieve that. It would be great if these classes provide some mechanism to achieve that (via props or state).

DatePicker

Expected-In case we need to update the date, we need to send the new value to the component and component should handle the change.
Happening- In order to update the date, we have to use the cal instance, in our code.

Chrome

Latest

sandhya-r-reddy:
Closes carbon-design-system/carbon-components-react#350

Added a new prop 'value' in DatePicker Component and the code to handle the date change to set the value programmatically.

Change log
Changed:
Added new Prop value, to set the value programmatically
Updated tests

### DIFF
--- a/src/components/DatePicker/DatePicker-test.js
+++ b/src/components/DatePicker/DatePicker-test.js
@@ -38,6 +38,12 @@ describe('DatePicker', () => {
       expect(wrapper.props().dateFormat).toEqual('d/m/Y');
     });
 
+    it('has the value as expected', () => {
+          expect(wrapper.props().value).toEqual(undefined);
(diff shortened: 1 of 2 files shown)


marijohannessen:
👍 ✅

PR #359, opened 2017-11-20
fix(Form): do not submit story for real

eddiemonge:
The current story tries to submit the form for real making the page change. This prevents this and shows an item in the action panel instead


tw15egan:
Thanks for fixing this!

PR #360, opened 2017-11-20
fix(NumberInput): add button type and disabled

### LINKED ISSUE #358: NumberInput event bubbles up from arrows
## Detailed description

Clicking on the arrows (either up or down) will emit an event that bubbles all the way up to the form wrapper and submits it. This happens even when NumberInput is disabled. Typing numbers into the form doesn't do this.

NumberInput

Chrome

7.x.x

API Connect, need to ship with carbon components instead of Bluemix components

1. Place the NumberInput component inside of a Form Component
2. Click on the up or down arrows inside of NumberInput
3. Form will submit

eddiemonge:
- Add button type to arrow button to prevent a form from being submitted
when it is clicked

- Disable the arrow buttons when the NumberInput is disabled just in
case

Fixes #358

### DIFF
--- a/src/components/NumberInput/NumberInput-test.js
+++ b/src/components/NumberInput/NumberInput-test.js
@@ -153,8 +153,13 @@ describe('NumberInput', () => {
       );
 
       const input = wrapper.find('input');
-      const upArrow = wrapper.find('.up-icon');
-      const downArrow = wrapper.find('.down-icon');
+      const upArrow = wrapper.find('.up-icon').parent();
+      const downArrow = wrapper.find('.down-icon').parent();
+
+      it('should be disabled when numberInput is disabled', () => {
+        expect(upArrow.prop('disabled')).toEqual(true);
+        expect(downArrow.prop('disabled')).toEqual(true);
+      });
 
       it('should not invoke onClick when up arrow is clicked', () => {
         upArrow.simulate('click');
@@ -187,8 +192,8 @@ describe('NumberInput', () => {
       );
 
       const input = wrapper.find('input');
-      const upArrow = wrapper.find('.up-icon').last();
-      const downArrow = wrapper.find('.down-icon').last();
+      const upArrow = wrapper.find('Icon.up-icon').closest('button');
+      const downArrow = wrapper.find('Icon.down-icon').closest('button');
 
       it('should invoke onClick when numberInput is clicked', () => {
         input.simulate('click');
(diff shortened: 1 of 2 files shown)


src/components/NumberInput/NumberInput-test.js
```diff
@@ -153,8 +153,13 @@ describe('NumberInput', () => {
       );
 
       const input = wrapper.find('input');
-      const upArrow = wrapper.find('.up-icon');
-      const downArrow = wrapper.find('.down-icon');
+      const upArrow = wrapper.find('.up-icon').parent();
```
eddiemonge: target the element that has the actual event handler

src/components/NumberInput/NumberInput-test.js
```diff
@@ -187,8 +192,8 @@ describe('NumberInput', () => {
       );
 
       const input = wrapper.find('input');
-      const upArrow = wrapper.find('.up-icon').last();
-      const downArrow = wrapper.find('.down-icon').last();
+      const upArrow = wrapper.find('Icon.up-icon').closest('button');
```
eddiemonge: this would use parent as well but there is a bug in enzyme with mount that I opened

src/components/NumberInput/NumberInput.js
```diff
@@ -105,6 +105,24 @@ export default class NumberInput extends Component {
       value: this.state.value,
     };
 
+    const buttonProps = {
+      type: 'button',
+      className: 'bx--number__control-btn',
+      disabled
+    };
+
+    const createButton = dir => (
```
eddiemonge: refactored this so its dry'er

joshblack: Do we need to do this if it's only at two call sites?

eddiemonge: 1 < 2 so id say yes

joshblack: I'm less concerned about file size and more concerned about closure allocs. I'm definitely down if the change made things much easier, but since it's so few call sites the change seems unnecessary.

tw15egan:
Can we also bump the version of `carbon-components` to `8.1.11`? That way the `disabled` styles are bundled in. Otherwise, looks good 👍 ✅

eddiemonge:
the peer is set to `"carbon-components": "^8.0.0",` which would update to that version. updated

PR #362, opened 2017-11-21
fix(NumberInput): call onClick, onChange after setState

bjones526:
Closes carbon-design-system/carbon-components-react#361

pass onClick and OnChange as a callback to setState to ensure they are not called before state.value is updated. This will keep any parent component that accesses this.state.value via refs during onChange synchronized with the value displayed in the input field.

**Changed**

- don't call onClick or onChange until this.state.value is changed.


tw15egan:
Looks good! 👍 ✅

PR #368, opened 2017-11-27
fix(a11y): fix a11y warnings in various components

tw15egan:
Fixes the `a11y` warnings being thrown during the linting step. To avoid breaking changes, I've removed `jsx-a11y/no-static-element-interactions

#### Changelog

**New**

- Added `tabIndex`, `role` to select components

**Changed**

- `AccordionItem`
- `ComposedModal`
- `Dropdown`
- `DropdownItem`
- `InteriorLeftNav`
- `InteriorLeftNavItem`
- `Modal`
- `Slider`
- `Switch`
- `Tab`
- `Tabs`

**Removed**

- A bit of extra things from `Dropdown` story

#### Testing

Make sure none of the above components are broken

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,8 @@
       "jsx-a11y/onclick-has-role": 1,
       "jsx-a11y/click-events-have-key-events": 1,
       "jsx-a11y/href-no-hash": 1,
-      "jsx-a11y/onclick-has-focus": 1
+      "jsx-a11y/onclick-has-focus": 1,
+      "jsx-a11y/no-static-element-interactions": 0
     },
     "env": {
       "node": true,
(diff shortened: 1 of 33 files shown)


joshblack:
@tw15egan are we safe removing that rule? Is this something that is coming up from the component stories? I know you mentioned that when we talked about this previously, was just curious. The rule itself seems valuable so was wondering why we wouldn't want it.

tw15egan:
Just turned it off while I was working on it, but to get rid of a lot of those warning's we'd have to move a bunch of the onClick handlers to different elements. Not sure how this would affect end users. I can add the warning back in for now

tw15egan:
Updated to `eslint-plugin-jsx-a11y` and fixed the new errors 😄

marijohannessen:
👍 ✅

PR #376, opened 2017-11-28
feat(NumberInput): add invalidText and invalid

### LINKED ISSUE #373: NumberInput is missing invalid and invalidText props
NumberInput is missing invalid and invalidText props. TextInput has them but NumberInput doesn't expose them

eddiemonge:
Add support for the invalid flag and for invalidText

Fixes #373 

Blocked by Carbon core not properly supporting invalidText with NumberInput (styles are wrong)

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -16,6 +16,8 @@ export default class NumberInput extends Component {
     onClick: PropTypes.func,
     step: PropTypes.number,
     value: PropTypes.number,
+    invalid: PropTypes.bool,
+    invalidText: PropTypes.string,
   };
 
   static defaultProps = {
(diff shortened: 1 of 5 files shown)


src/components/NumberInput/NumberInput.js
```diff
@@ -106,13 +112,19 @@ export default class NumberInput extends Component {
...
+      inputProps['data-invalid'] = true;
+      error = <div className="bx--form-requirement">{invalidText}</div>
+    }
+
     return (
       <div className="bx--form-item">
         <label htmlFor={id} className="bx--label">
           {label}
         </label>
         <div className={numberInputClasses}>
```
tw15egan: Based on the way we're using `data-invalid` and a general sibling selector, I think we'll have to put the `data-invalid` on the wrapping `bx--number` `div`. Putting `bx--form-requirement` adjacent to `bx--number` (as opposed to most other inputs that have it adjacent to the actual input) seems to avoid most styling issues. I've got a PR up in the base repo that addresses this.

eddiemonge:
Updated. Requires carbon>8.1.11 so might be minor/major semver change

PR #379, opened 2017-11-29
fix: remove unused dependency react-addons-css-transition-group, resolves #378

### LINKED ISSUE #378: Dependency warning for react-addons-css-transition-group with React 16
When installing the latest carbon-react-components, I see the following error:
```
npm WARN react-addons-css-transition-group@15.6.2 requires a peer of react@^15.4.2 but none was installed.
```

I believe `react-addons-css-transition-group` has been deprecated as of React 16, and `react-transition-group` should be used instead. That being said, I didn't actually see anywhere in the current code that was using `react-addons-css-transition-group`(but perhaps I missed something).

https://github.com/carbon-design-system/carbon-components-react/blob/master/package.json#L94
Version: 5.0.0
Related issue: https://github.com/carbon-design-system/carbon-components-react/issues/280

alisonemiller:
Closes carbon-design-system/carbon-components-react#378

Removed `react-addons-css-transition-group` from dependencies as it was no longer being used and causing warnings with React 16.

#### Changelog

- Removed `react-addons-css-transition-group` from `package.json` and updated the `yarn.lock` file. I doesn't look like the `yarn.lock` file was updated via https://github.com/carbon-design-system/carbon-components-react/pull/366.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "classnames": "2.2.5",
     "flatpickr": "4.0.7",
     "lodash.debounce": "^4.0.8",
-    "react-addons-css-transition-group": "^15.4.2",
     "warning": "3.0.0",
     "window-or-global": "^1.0.1"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,42 +2,106 @@
 # yarn lockfile v1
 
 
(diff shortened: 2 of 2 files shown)


joshblack:
Thanks so much @alisonemiller!

PR #382, opened 2017-11-30
299 remove default prop values

alisonjoseph:
Sets the default props to empty strings for the following components
-   Modal
-   TimePickerSelect
-   TextArea
-   DatePickerInput
-   Checkbox
-   ToolbarSearch
-   RadioButton
-   Search
-   FormGroup
-   ProgressIndicator

Closes carbon-design-system/carbon-components-react#299


joshblack:
Thanks so much @alisonjoseph for taking the time to contribute this! It's incredibly helpful 😄

For certain props that are required for a11y, we'd probably just want to make them required instead of defaulting to an empty string. How do you feel about that?

alisonjoseph:
Yes, that makes more sense, I was a little unsure when removing the defaults for some of these. 

I'm not entirely sure how to set them as required though 😄

joshblack:
I think simplest thing is just updating the `propTypes` definitions to be something like:

```js
Foo.propTypes = {
  // ...other props
  labelText: PropTypes.string.isRequired,
};
```

Would that work? And then just remove the field from `defaultProps`, if it exists.

alisonjoseph:
I updated these to be required where it made sense, but did you mean remove them entirely? Or just keep the empty string in the `defaultProps`?

joshblack:
@alisonjoseph right, remove them completely. If we leave the empty string on the default props, then the `isRequired` check will always pass. It was my assumption that we want to make people pass in the label text as a non-null, non-empty string for a11y purposes. Let me know if that's not the case!

alisonjoseph:
@joshblack that makes sense, however when I removed those then all the tests are failing for those components. 

```
Warning: Failed prop type: The prop `labelText` is marked as required in `TextInput`, but its value is `undefined`.
          in TextInput
```

joshblack:
Just means we have to update the components calling it so that it doesn't use the default prop value (which was a generic message anyway).

src/components/Checkbox/Checkbox.js
```diff
@@ -49,14 +49,13 @@ Checkbox.propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
-  labelText: PropTypes.node,
+  labelText: PropTypes.node.isRequired,
```
chrisdhanaraj: Jumping in - the reason why this was a problem was because people were making composed UI out of the elements. Requiring the label hurt the 'pro' user who was handling the a11y differently

joshblack: @chrisdhanaraj hi chris! 👋 

Do you have a code snippet in mind? Am down to implement a custom `prop` checker logic to accommodate what you are describing 😄

marijohannessen: @chrisdhanaraj any input here? 😄

chrisdhanaraj: Oh sorry, ignore me! For some reason I was mixing up the Checkbox heading and the checkbox label :) my total bad

marijohannessen: Lol @chrisdhanaraj way to keep us in limbo here!

chrisdhanaraj: https://media.giphy.com/media/d2YZg3m63zNgWa6k/giphy.gif

chrisdhanaraj: Wow that's a high resolution gif right there

alisonjoseph:
I think it would potentially be a breaking change?

marijohannessen:
Yeah, it would require people to have to update their components if they have not been using `labelText`, so I see it as a breaking change. We could avoid that by adding the defaultProps back in but that kind of goes against the purpose of this PR. I'm just a bit wary against releasing a breaking change as we've promised people that we'd only do that twice a year :-/

alisonjoseph:
We could add this back in so the default is blank, however that would allow users to add components without label fields which may be needed for A11y purposes for some components.  

labelText: '',

marijohannessen:
I think making them required and adding back in the blank ones can be a good solution for now - and then once we release the next major version we'll remove the defaultProps. That'll give people some time to adjust to the change. What do you think @joshblack ?

joshblack:
@marijohannessen for people currently using these components, what issues do you seem them having with the updates?

My hunch is that we'll see people who left the default props start getting prop type validation warnings, which thankfully won't stop them from building or anything and should nudge them to actual write out a11y text.

alisonjoseph:
If a prop is marked as `isRequired` and a user leaves the value blank, will it just throw a warning, or will it actually break their application? If it just throws a warning, then I would say that's not a breaking change and we would be fine to add it.

joshblack:
@alisonjoseph yupp! It should just use `console.error` to warn them and shouldn't break any existing components.

Should definitely try it with some team to make sure though 😅

marijohannessen:
Okay, if that's all that will happen, then I'm good with merging it once tests have been fixed! 👍

PR #385, opened 2017-12-01
fix(a11y): update jsx-a11y to v6

tw15egan:
Updates `eslint-plugin-jsx-a11y` to `v6`

- Changed empty `href` to actual links
- Changed `left-nav-list__item-link` from `a` to `div` since it was not being used to navigate

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,21 +14,15 @@
     "prepublish": "yarn build",
     "prettier": "prettier --write \"**/*.{scss,css,js}\"",
     "prettier:diff": "prettier --list-different \"**/*.{scss,css,js}\"",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -14,21 +14,15 @@
     "prepublish": "yarn build",
     "prettier": "prettier --write \"**/*.{scss,css,js}\"",
     "prettier:diff": "prettier --list-different \"**/*.{scss,css,js}\"",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
+    "semantic-release":
```
joshblack: Is this from prettier?

tw15egan: Yup, any formatting issues like this were from prettier

PR #392, opened 2017-12-05
feat(tooltip): add menuoffset prop

tw15egan:
### Small changes to Tooltip

- Changes `direction` proptypes to `PropTypes.oneOf(['bottom', 'top', 'left', 'right'])`
- Adds in `menuOffset` prop to ensure Tooltip is always positioned correctly. Usage is as follows:

```jsx
<Tooltip triggerText="Tooltip - no icon" menuOffset={{left: 10, top: 30}}>
```

Related to https://github.com/carbon-design-system/carbon-components-react/issues/390#issuecomment-349335529

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -9,7 +9,8 @@ export default class Tooltip extends Component {
     open: PropTypes.bool,
     children: PropTypes.node,
     className: PropTypes.string,
-    direction: PropTypes.oneOf(['bottom', 'top']),
+    direction: PropTypes.oneOf(['bottom', 'top', 'left', 'right']),
+    menuOffset: PropTypes.obj,
     triggerText: PropTypes.string,
     showIcon: PropTypes.bool,
     iconName: PropTypes.string,
@@ -23,6 +24,10 @@ export default class Tooltip extends Component {
     iconName: 'info--glyph',
     iconDescription: 'tooltip',
     triggerText: 'Provide triggerText',
+    menuOffset: {
+      left: 5,
+      top: 10,
+    },
   };
 
   state = {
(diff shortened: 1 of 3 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -58,6 +63,7 @@ export default class Tooltip extends Component {
       showIcon,
       iconName,
       iconDescription,
+      menuOffset,
```
DmitriWolf: When a user enters a value of `menuOffset = {left: 10}` that overwrites the object entirely, giving `menuOffset.top` a value of `undefined`. This breaks the view.

DmitriWolf: `direction="right"` default positioning is off. I think the default value of `menuOffset.left` may be putting the floatingMenu too far over.

DmitriWolf:
`direction="right"` default positioning is off. I think the default value of `menuOffset.left` may be putting the floatingMenu too far over.

src/components/Tooltip/Tooltip.js
```diff
@@ -9,7 +9,8 @@ export default class Tooltip extends Component {
     open: PropTypes.bool,
     children: PropTypes.node,
     className: PropTypes.string,
-    direction: PropTypes.oneOf(['bottom', 'top']),
+    direction: PropTypes.oneOf(['bottom', 'top', 'left', 'right']),
+    menuOffset: PropTypes.obj,
```
DmitriWolf: This should be `menuOffset: PropTypes.object,` otherwise I get a warning in my console:
```
Warning: Failed prop type: Tooltip: prop type `menuOffset` is invalid; it must be a function, usually from React.PropTypes.
```

DmitriWolf:
@tw15egan If you don't have time to work out the bugs, let me know and I can help.

tw15egan:
@DmitriWolf if you have time to improve this PR that would be great

DmitriWolf:
@tw15egan Do you have a way of developing locally with `npm link` or do you only develop with Storybook?

Issue #307, opened 2017-11-06
[RFC] Tables v2 API Design


joshblack:
This issue will be used for the API Design of the [Tables v2](http://carbondesignsystem.com/components/data-table/code) work introduced in Carbon v8.

Any and all feedback is welcome! 😄

jzhang300:
```

joshblack:
A stab at trying to see what a component might "look" like in JSX:

![screen shot 2017-11-06 at 12 23 14 pm](https://user-images.githubusercontent.com/3901764/32457054-4729e29a-c2ed-11e7-979d-98333929eb31.png)

Given the following tabular data:

```js
const data = [
  {
    name: 'Load Balancer 1',
    protocol: 'HTTP',
    something: 80,
    rule: 'Round Robin',
    attached_groups: 'Maureen\'s VM Groups',
    status: 'active',
  },
  {
    name: 'Load Balancer 2',
    protocol: 'HTTP',
    something: 80,
    rule: 'Round Robin',
    attached_groups: 'Maureen\'s VM Groups',
    status: 'active',
  },
  {
    name: 'Load Balancer 3',
    protocol: 'HTTP',
    something: 80,
    rule: 'Round Robin',
    attached_groups: 'Maureen\'s VM Groups',
    status: 'active',
  },
  {
    name: 'Load Balancer 4',
    protocol: 'HTTP',
    something: 80,
    rule: 'Round Robin',
    attached_groups: 'Maureen\'s VM Groups',
    status: 'active',
  },
  {
    name: 'Load Balancer 5',
    protocol: 'HTTP',
    something: 80,
    rule: 'Round Robin',
    attached_groups: 'Maureen\'s VM Groups',
    status: 'active',
  },
];

<DataTable>
  {({
    // Can be set in `DataTable` as props, also might contain state info
    getToolbarProps,
    // Mirrors `Pagination` component props, also might contain state info
    getPaginationProps,

    // Method that is given a table header
    sortBy,

    rows,
  }) => (
    <Toolbar {...getToolbarProps()} />
    <Table>
      <TableHead>
        {Object.keys(rows[0]).map(header => (
          <TableHeader key={header} sortable={true} onClick={sortBy(header)}>
            {header}
          </TableHeader>
        ))}
      </TableHead>
      <TableBody>
        {rows.map((row, i) => (
          <TableRow isSelected={row.isSelected} isExpanded={row.isExpanded}>
            {Object.values(row).map(cell => (
[...]

PR #393, opened 2017-12-05
feat(data-table): added data table v2 component

marijohannessen:
Ref #307 

#### Added
- [x] Batch actions
- [x] Toolbar
- [x] Sorting
- [x] Expanded rows
- [x] Searching

#### To-do still in this release:
- [ ] Finish tests

### Coming later
- Pagination hooked up with the tables (they are separate right now)


src/components/DataTable/DataTable-story.js
```diff
@@ -0,0 +1,397 @@
+import React, { Component } from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import DataTableRow from '../DataTableRow';
+import DataTableData from '../DataTableData';
+import DataTableRowExpand from '../DataTableRowExpand'; // eslint-disable-line
```
joshblack: Are the `// eslint-disable-line` lines here going to be added in in the future?

marijohannessen: Nope, those are temporary :-)

marijohannessen: But wasn't able to push up this PR without fixing lint errors

joshblack: Can always add `--no-verify` if you just want to check stuff in 😄

src/components/DataTable/DataTable.js
```diff
@@ -0,0 +1,364 @@
...
+import classNames from 'classnames';
+import Button from '../Button';
+import Icon from '../Icon';
+import Search from '../Search';
+
+export class DataTable extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      rows: props.initialRows,
```
psachs21: Do we need `initialRows` at all? It seems that the render function handles rendering entirely. This component doesn't even seem necessary.

src/components/DataTable/DataTable.js
```diff
@@ -0,0 +1,364 @@
...
+  };
+
+  removeActive = () => {
+    this.setState({
+      active: false,
+    });
+  };
+
+  handleClick = () => {
+    this.setState({
```
psachs21: Should be using callback method to avoid potential asynchronous issues.

```js
this.setState(({ sorted }) => {
  const nextSorted = !sorted;
  const direction = sorted ? 'desc' : 'asc'; // Maybe different order? Not sure what default sort is.
  this.props.onClick(this.props.sortBy, direction);
  return { sorted: !nextSorted };
});
```

src/components/DataTable/DataTable.js
joshblack: `<DataTable>` will eventually need to support a pagination use-case, and typically we start off with an initial page, or two, of data and then load additional pages when the user requests them. I think this work is just in preparation for handling that state logic 😄

src/components/DataTable/DataTable.js
joshblack: Most likely it'd be this variant, I think:

```js
this.setState(state => ({ sorted: !state.sorted }), () => {
   const direction = this.state.sorted ? 'desc' : 'asc';
   this.props.onClick(this.props.sortBy, direction);
});
```

psachs21: oooo, didn't even know setState could take more than one callback. Yes, that's better.

src/components/DataTable/DataTable.js
psachs21: Hmm, thought the pagination use case was not being handled by this component at all, since it offloaded that work to the provided render method. I like that the datatable doesn't worry about rendering the rows, because than this component could handle infinite scrolling (using react-virtualized, or something) or go the traditional pagination route, without having to touch the implementation of the datatable.

src/components/DataTable/DataTable-story.js
```diff
@@ -0,0 +1,580 @@
...
+                iconName="settings"
+                iconDescription="Settings"
+                onClick={action('Toolbar Action 1')}
+              />
+              <Button onClick={action('Add new row')} kind="primary">
+                Add new
+              </Button>
+            </DataTableToolbarContent>
+          </DataTableToolbar>
+          <DataTable
```
joshblack: What if `<DataTable>` looked something like:

```js
<DataTable
  initialRows={data.rows}
  headers={['Foo', 'Bar', 'Baz']}
  {/* enable checkboxes, name can definitely change! */}
  selectable={true}
  render={({
    // Current page to render
    page,

    // Headers passed through from above
    headers,

    // Prop getters for children components to keep in sync with internal
    // state in DataTable
    getPaginationProps,
    // Returns things like `onClick` handlers if the item is select-able
    getSelectionProps,
    // Potentially useful for inline editing
    getColumnProps,
    getHeaderProps,

    // Sorters
    sortBy,
    sortByFoo,
    sortByBar,
    sortByBaz,

    // Selectors
    selectRow,
  }) => (
    <table>
      <TableHead>
        {headers.map(header => (
          <TableHeader {...getHeaderProps({ header })}>
            {header}
          </TableHeader>
        ))}
      </TableHead>
      <TableBody>
        {page.rows.map(row => (
          <TableRow key={`row-select-${row.id}`}>
            {row.isSelectable && (
              <TableData {...getSelectionProps()} />
            )}
          </TableRow>
          {row.columns.map(column => (
            <TableData {...getColumnProps({ column })} />
          ))}
        ))}
      </TableBody>
    </table>
  )}
/>
```

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+    } = this.props;
+
+    const statePage = this.state.page;
+    const statePageSize = this.state.pageSize;
+    const classNames = classnames('bx--pagination', className);
+    const inputId = id || this.uniqueId;
+    const totalPages = Math.ceil(totalItems / statePageSize);
+    const selectItems = this.renderSelectItems(totalPages);
+
+    return (
```
joshblack: Could we break up `Pagination` into smaller components, or does it make more sense for it to be in one file?

src/components/DataTable/DataTable-story.js
```diff
@@ -0,0 +1,580 @@
...
+import {
+  DataTableExpandableRow, // eslint-disable-line
+  DataTableExpandableRowContent, // eslint-disable-line
+} from '../DataTableExpandableRow';
+import {
+  DataTable,
+  DataTableToolbar,
+  DataTableToolbarContent,
+  DataTableToolbarAction,
+  DataTableHead,
```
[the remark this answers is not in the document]
marijohannessen: Yeah, that makes sense! I named it `DataTableHead` because of the HTML element `thead`. So I might keep `DataTableHead` but then change `DataTableHeader` to `DataTableColumnHeader`.

src/components/DataTable/DataTable.js
marijohannessen: Ah gotcha, will change it to that!

src/components/PaginationV2/PaginationV2.js
marijohannessen: That could probably be a good refactor project for v9, or an enhancement for when the tables are ready.

src/components/DataTable/DataTable-story.js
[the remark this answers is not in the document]
psachs21: Makes sense.

src/components/DataTable/DataTable-story.js
psachs21: Hmm, alright. Though I think in this case, we shouldn't call it `TableData` but `TableCell` to be consistent with the html table concept.

As long as initialRows is optional and we can potentially bypass the paging entirely (for the sake of infinite scroll). 

The sorters are interesting though, are they triggering sort or are they performing the sort? I thought sort itself might sit outside this component, since it might be triggered from outside the component (say, from a url or redux action). Same concern for selectRow. I'm not sure how much the the table itself should track those properties, since they can be triggered by many external sources.

marijohannessen: I agree about the sorting part, I think having that be controlled outside the component is probably best. Having `sortBy`, `sortByFoo`, ... , seems a bit redundant as there might be a large number of queries users would want to sort by.

marijohannessen: I'm a bit confused about this suggested setup. I think the header part is not customizable enough this way, as I think the user should be able to specify what they want to sort each column by - I might be missing something though. I also feel like I should wait with the pagination parts until after the v1 release - since pagination is not part of the actual component right now? I might just need some help visualizing what the `DataTable` component would look like, it's not apparently clear to me with this pattern.

marijohannessen: Also I think `DataTableData` is fine as w3 explains the `td` element as such:
`Use the HTML <td> element to define a table data`.

psachs21: Huh , look at that. Fair enough. I always think of them as cells. It doesn't help that "DataTableData" doesn't exactly roll off the tongue I guess. Data always feels like such an overloaded term.

joshblack: Whoops, duh, the `sortBy` stuff was in an earlier draft I had as the comment which I edited/replaced. `getHeaderProps` would include the sort functionality, potentially, or they could manually put the `sortBy` field on it if they wanted and did `sortBy(header)` and it would set it up for them dynamically.

joshblack: Link to an example of what I was thinking: https://codesandbox.io/s/pnzk98jq

src/components/DataTableData/DataTableData.js
```diff
@@ -0,0 +1,63 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const DataTableData = props => {
+  const { children, className, checked, value, id, expanded, ...other } = props;
```
joshblack: For prop names, should they be `isSelectable` and `isExpandable`? The rationale being that having those props adds in the behavior to be selected or expanded, but doesn't indicate any kind of state in this component (e.g. if it's selected or checked).

For selectable vs checked, the rationale being that there is a potential that selecting a role might not be tied to a checkbox UI control so would be an abstraction on top of that.

src/components/DataTableRow/DataTableRow.js
```diff
@@ -0,0 +1,20 @@
...
+    <tr {...other} className={className}>
+      {children}
+    </tr>
+  );
+};
+
+DataTableRow.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  containerClassName: PropTypes.string,
```
joshblack: Seems like this is an extra prop, potentially, when `classnames` was being used?

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+  state = {
+    page: this.props.page,
+    pageSize:
+      this.props.pageSize && this.props.pageSizes.includes(this.props.pageSize)
+        ? this.props.pageSize
+        : this.props.pageSizes[0],
+  };
+
+  componentWillMount() {
+    this.uniqueId = `${Math.floor(Math.random() * 0xffff)}`;
```
joshblack: Do we need to have a uniquely generated id like this, or could we just have a counter in the main scope that we increment? e.g.

```js
let instanceId = 0;

...

componentWillMount() {
  this.uniqueId = ++instanceId;
}
```

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  componentWillReceiveProps({ pageSizes, page, pageSize }) {
+    if (!equals(pageSizes, this.props.pageSizes)) {
+      this.setState({ pageSize: pageSizes[0], page: 1 });
+    }
+    if (page !== this.props.page) {
```
joshblack: Are these if statements mutally exclusive? If so, should we make them `else if` just to guarantee multiple branches don't execute?

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+    isLastPage: PropTypes.bool,
+    pageInputDisabled: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    backwardText: 'Backward',
+    itemRangeText: (min, max, total) => `${min}-${max} of ${total} items`,
+    forwardText: 'Forward',
+    itemsPerPageText: 'Items per page',
+    onChange: () => {},
```
joshblack: Instead of alloc'ing an empty function, could we just check for the `onChange` prop before calling it?

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+    }
+    if (pageSize !== this.props.pageSize) {
+      this.setState({ pageSize });
+    }
+  }
+
+  handleSizeChange = evt => {
+    const pageSize = Number(evt.target.value);
+    this.setState({ pageSize, page: 1 });
+    this.props.onChange({ page: 1, pageSize });
```
joshblack: Should we add this call to the callback of `setState`? So the consumer is guaranteed that `onChange` is called while the component is in the given state.

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+  };
+
+  handlePageChange = evt => {
+    this.setState({ page: evt.target.value });
```
joshblack: Do we need to call `Number` on this like in some of the other cases?

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+  };
+
+  handlePageInputChange = evt => {
+    const page = Number(evt.target.value);
+    if (
+      page > 0 &&
+      page <= Math.ceil(this.props.totalItems / this.state.pageSize)
+    ) {
+      this.setState({ page });
+      this.props.onChange({ page, pageSize: this.state.pageSize });
```
joshblack: Same note as above, should this be added to the callback of `setState`?

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+    }
+  };
+
+  incrementPage = () => {
+    const page = this.state.page + 1;
+    this.setState({ page });
+    this.props.onChange({ page, pageSize: this.state.pageSize });
```
joshblack: Same note as above

src/components/PaginationV2/PaginationV2.js
```diff
@@ -0,0 +1,228 @@
...
+  };
+
+  decrementPage = () => {
+    const page = this.state.page - 1;
+    this.setState({ page });
+    this.props.onChange({ page, pageSize: this.state.pageSize });
```
joshblack: Same note as above

package.json
```diff
@@ -109,6 +109,7 @@
     "flatpickr": "4.0.7",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
+    "sentence-case": "^2.1.1",
```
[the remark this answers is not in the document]
marijohannessen: I'm okay with leaving it out - that's probably best for internationalization purposes as well

src/components/DataTableData/DataTableData.js
marijohannessen: I can switch it to `isSelectable` and `isExpandable`, that makes sense.

src/components/DataTableData/DataTableData.js
```diff
@@ -0,0 +1,63 @@
...
+
+  let content;
+  if (!(checked === undefined) && !(expanded === undefined)) {
+    content = '';
+  } else if (checked !== undefined) {
```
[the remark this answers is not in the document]
marijohannessen: That looks so much better, will switch!

src/components/DataTableData/DataTableData.js
```diff
@@ -0,0 +1,63 @@
...
+          checked={checked}
+          id={id}
+          className="bx--checkbox"
+          type="checkbox"
+          value={value}
+          name={id}
+        />
+        <label htmlFor={id} className="bx--checkbox-label">
+          <span className="bx--checkbox-appearance">
+            <svg
```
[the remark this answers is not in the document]
marijohannessen: I can switch the expandable icon, but for the checkmark one I needed to add the full HTML because our `Checkbox` component gets wrapped in a `bx--form-item` element.

src/components/DataTableRow/DataTableRow.js
marijohannessen: Yup, removing it! 😄

PR #401, opened 2017-12-11
chore(Pkg): add npm autoformatting

eddiemonge:
npm (or yarn, not sure which), auto-formats the package.json file. These are the changes it wants.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
+package.json
 node_modules
 lib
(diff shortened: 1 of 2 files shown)


joshblack:
Thanks @eddiemonge! The changes are coming from prettier being run on `package.json`, would you want to add that to the `.prettierignore` file just to make sure this doesn't happen again?

PR #405, opened 2017-12-12
fix(NumberInput): pass the evt and dir to events

### LINKED ISSUE #369: NumerInput no longer passes the event param
I've been using `NumberInput` in various ways (shown below) and I've had access to the event object passed by the `onChange` prop. However with the newest version (4.2.2), the param is now `undefined`. v4.2.2 includes the following PR: https://github.com/carbon-design-system/carbon-components-react/pull/362 which looks like it directly affects `NumberInput` code. 
Just for testing, I rolled back to 4.2.1 and the event param is properly being assigned and I'm able to use it as such.

Example code of how I've been using `NumberInput`

```javascript
<NumberInput
  ...
  onChange={this.updateQuantity}
  ...
/>

// or

<NumberInput
  ...
  onChange={(e) => this.updateQuantity(e)}
  ...
/>
```

eddiemonge:
onChange and onClick will now receive the events used to trigger them.
This is a regression fix.

They will also receive the direction as a second argument if the arrows
are used.

Fixes the tests as they were returning false positives when checking on
spies. The spies were not reset between test runs so passing it in one
test means it passed in the rest.

Fixes #369

### DIFF
--- a/src/components/NumberInput/NumberInput-test.js
+++ b/src/components/NumberInput/NumberInput-test.js
@@ -178,22 +178,31 @@ describe('NumberInput', () => {
     });
 
     describe('enabled numberInput', () => {
-      const onClick = jest.fn();
-      const onChange = jest.fn();
+      let onClick;
+      let onChange;
+      let input;
+      let upArrow;
+      let downArrow;
+      let wrapper;
 
-      const wrapper = mount(
-        <NumberInput
-          id="test"
-          onClick={onClick}
-          onChange={onChange}
-          min={0}
-          max={100}
-        />
-      );
+      beforeEach(() => {
+        onClick = jest.fn();
+        onChange = jest.fn();
 
-      const input = wrapper.find('input');
-      const upArrow = wrapper.find('Icon.up-icon').closest('button');
-      const downArrow = wrapper.find('Icon.down-icon').closest('button');
+        wrapper = mount(
+          <NumberInput
+            id="test"
+            onClick={onClick}
+            onChange={onChange}
+            min={0}
+            max={100}
+          />
+        );
+
+        input = wrapper.find('input');
+        upArrow = wrapper.find('Icon.up-icon').closest('button');
+        downArrow = wrapper.find('Icon.down-icon').closest('button');
+      });
 
       it('should invoke onClick when numberInput is clicked', () => {
         input.simulate('click');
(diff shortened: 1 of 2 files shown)


src/components/NumberInput/NumberInput-test.js
```diff
@@ -178,22 +178,31 @@ describe('NumberInput', () => {
...
-      const upArrow = wrapper.find('Icon.up-icon').closest('button');
-      const downArrow = wrapper.find('Icon.down-icon').closest('button');
+      let onClick;
+      let onChange;
+      let input;
+      let upArrow;
+      let downArrow;
+      let wrapper;
+
+      beforeEach(() => {
```
eddiemonge: reset the state between tests instead of using a shared state

src/components/NumberInput/NumberInput-test.js
```diff
@@ -203,18 +212,21 @@ describe('NumberInput', () => {
       it('should invoke onClick when up arrow is clicked', () => {
         upArrow.simulate('click');
         expect(onClick).toBeCalled();
+        expect(onClick).toHaveBeenCalledWith(expect.anything(), 'up');
```
eddiemonge: wasn't sure how to create a synthetic event to pass to expect so `.anything()` matches anything that is not undefined or null. It was called with undefined before

src/components/NumberInput/NumberInput-test.js
```diff
@@ -237,14 +249,22 @@ describe('NumberInput', () => {
         expect(wrapper.state().value).toEqual(90);
       });
 
-      it('should invoke onClick when down arrow is clicked', () => {
+      it('should not invoke onClick when down arrow is clicked and value is 0', () => {
```
eddiemonge: this was a false positive in the tests before because value wasn't 0. Resetting in between tests exposed this test's weakness

src/components/NumberInput/NumberInput.js
```diff
@@ -53,7 +53,7 @@ export default class NumberInput extends Component {
         {
           value: evt.target.value,
         },
-        evt => {
+        () => {
```
eddiemonge: having `evt` here was overriding the `evt` passed to the function. The callback to setState doesn't pass anything so evt became undefined

PR #408, opened 2017-12-13
fix(FormItem): use correct export

eddiemonge:
The export for formItem was not set correctly.

Updated the tests to use the exported module instead of the file directly


ajdaniel:
👍

PR #415, opened 2017-12-18
fix(slider): allow slider to use text input when min is something oth…

tw15egan:
This fixes https://github.com/carbon-design-system/carbon-components-react/issues/287

**To test**: Change the `min` to something other than `0` and confirm you can change the slider value by changing the text input

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -66,7 +66,11 @@ export default class Slider extends PureComponent {
 
     requestAnimationFrame(() => {
       this.setState((prevState, props) => {
-        if (evt && evt.target.classList.contains('bx-slider-text-input')) {
+        if (
+          evt &&
+          evt.target &&
+          evt.target.classList.contains('bx-slider-text-input')
+        ) {
           const { left } = this.calcValue(evt, prevState, props);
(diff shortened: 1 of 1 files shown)


src/components/Slider/Slider.js
```diff
@@ -66,6 +66,11 @@ export default class Slider extends PureComponent {
 
     requestAnimationFrame(() => {
       this.setState((prevState, props) => {
+        if (evt && evt.target.classList.contains('bx-slider-text-input')) {
```
joshblack: Is it possible that `evt.target` could be undefined here?

asudoh: Another thing I'm wondering is if `evt.target` could be a SVG: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/1173756/

tw15egan: Would it be good to just change this to :

```js
if (evt && evt.target && evt.target.classList.contains('bx-slider-text-input'))
```

joshblack: Sounds good to me 😄

tw15egan: 👍

PR #419, opened 2017-12-19
fix(datePicker): Send arguments to datepicker on change event

psachs21:
Closes carbon-design-system/carbon-components-react#286

Fixing up date picker.

  #

### DIFF
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -23,11 +23,13 @@ export default class DatePicker extends Component {
     datePickerType: PropTypes.string,
     dateFormat: PropTypes.string,
     value: PropTypes.string,
+    onChange: PropTypes.func,
   };
 
   static defaultProps = {
     short: false,
     dateFormat: 'm/d/Y',
+    onChange: () => {},
   };
 
   componentWillUpdate(nextProps) {
@@ -60,9 +62,7 @@ export default class DatePicker extends Component {
             ? [new rangePlugin({ input: this.toInputField })]
             : '',
         clickOpens: true,
-        onChange: () => {
-          this.props.onChange();
-        },
+        onChange: this.props.onChange,
         onReady: (selectedDates, dateStr, instance) => {
           this.updateClassNames(instance);
         },
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -68,6 +68,16 @@ export default class DatePicker extends Component {
         },
         nextArrow: this.rightArrowHTML(),
         leftArrow: this.leftArrowHTML(),
+        onMonthChange: (electedDates, dateStr, instance) => {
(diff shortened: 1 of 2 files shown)


tw15egan:
@psachs21 could you update the tests for this so it passes? Thanks in advance!

psachs21:
Woops, my bad. Fixed tests.

src/components/DatePicker/DatePicker.js
```diff
@@ -23,11 +23,13 @@ export default class DatePicker extends Component {
     datePickerType: PropTypes.string,
     dateFormat: PropTypes.string,
     value: PropTypes.string,
+    onChange: PropTypes.func,
   };
 
   static defaultProps = {
     short: false,
     dateFormat: 'm/d/Y',
+    onChange: () => {},
```
joshblack: Instead of allocating the empty function here, could we just make assertions on if this `prop` is available and then invoke it?

src/components/DatePicker/DatePicker.js
```diff
@@ -60,14 +62,21 @@ export default class DatePicker extends Component {
...
-        onChange: () => {
-          this.props.onChange();
-        },
+        onChange: this.props.onChange,
         onReady: (selectedDates, dateStr, instance) => {
           this.updateClassNames(instance);
         },
         nextArrow: this.rightArrowHTML(),
         leftArrow: this.leftArrowHTML(),
+        onMonthChange: (electedDates, dateStr, instance) => {
```
joshblack: Are these all the events that we should be exposing to `updateClassNames`? Or are these the only ones developers would care about?

psachs21: Good catch. There was another case that I didn't catch: `onValueChanged`. Updated to reflect catching that event.

src/components/DatePicker/DatePicker.js
psachs21: I am personally more a fan of a default rather than the cognitive load of having to put a value check on every use case (given the sacrifice is just a single function allocation per instance). But I will follow the repo standard if there's a good case for it.

joshblack: The rationale is that we don't want to have to initialize empty closures when we don't have to 😄 

Since we are a library, performance is one of our guiding principles that we try to follow. While this is a very small thing that ultimately wouldn't be the bottleneck in an application, the hope is that we can try to stay as lean as possible on behalf of consumers.

If the empty function is indeed a pattern that we'd want to follow, then we would want to share the same empty function reference between props, similar to Facebook's [`emptyFunction`](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/emptyFunction.js).

psachs21:
I understand now. That makes sense. I'll make the requisite changes.

psachs21:
@joshblack updated according to review comments.

tw15egan:
👍 ✅

Issue #425, opened 2017-12-22
Dropdown/overflow menu/toolbar search: Extra uncustomizable div in dom tree


psachs21:
## Detailed description

I'm trying to use flexbox to better align the styles of carbon-components with our existing projects but there are a few components that have divs that are none-customizable but also redundant. For example, dropdown:

```jsx
<div>
  <ul className="bx--dropdown"...>
  </ul>
</div>
```

> Is this issue related to a specific component?

Dropdown/overflow menu/toolbar search

> What did you expect to happen? What happened instead? What would you like to see changed?

These extra divs should not be there. They prevent us from styling the dropdown without use of excessive wrapping elements. They also make it difficult to match carbon-components html structure because they introduce functional only html elements.

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

^8.0.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

RCA (Watson Finance)

PR #426, opened 2017-12-22
fix(dropdown,overflow-menu,toolbar-search): Addresses #425

psachs21:
Extends ClickListener to allow more flexible rendering options. You can now specify what kind of element is rendered and add additional properties. Note that this makes using the ref a little harder (you need to access element within the instance).

Closes carbon-design-system/carbon-components-react#425

#### Changelog

**Removed**

- Extra divs wrapping dropdown/overflow-menu/toolbar-search

### DIFF
--- a/src/components/Dropdown/Dropdown-story.js
+++ b/src/components/Dropdown/Dropdown-story.js
@@ -22,8 +22,7 @@ storiesOf('Dropdown', module)
       <Dropdown
         {...dropdownEvents}
         onChange={action('onChange')}
-        defaultText="Dropdown label"
-        className="xxxStuff">
+        defaultText="Dropdown label">
         <DropdownItem itemText="Option 1" value="option1" />
         <DropdownItem itemText="Option 2" value="option2" />
         <DropdownItem itemText="Option 3" value="option3" />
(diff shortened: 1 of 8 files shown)


joshblack:
Hi @psachs21! Thanks so much for contributing this PR!

I think for `ClickListener`, the hope was to have it:

- Require only a single child node
- Use `cloneElement` on that that child node
- Safely copy over any props passed in so we don't overwrite consumer props with internal props

Would that also help your use-case that you identified in #425?

psachs21:
I see. That would indeed be sufficient. I'll adjust my PR accordingly.

psachs21:
I guess the only problem I see is if the user wants a ref on that child
node. I am not sure how to extract that ref for us to use properly. I
haven't tried though so I'll give it a shot.

psachs21:
@joshblack I've updated the PR according to your suggestions and I'm much happier with the results. Not sure how the error I throw if more than one child is provided so if you have suggestion for that, it's much appreciated.

joshblack:
@psachs21 thanks so much for taking the time to refactor!

With respect to your question, do you think https://reactjs.org/docs/typechecking-with-proptypes.html#requiring-single-child would be sufficient?

psachs21:
That will do just fine. I'll the adjustment when I get time.

joshblack:
@psachs21 ended up getting some time today so took some of the ideas here and put them into #430!

joshblack:
Going to close this as #430 ended up getting merged in. Feel free to re-open if that PR doesn't cover everything that you wanted covered!

PR #430, opened 2018-01-01
Update `ClickListener` and introduce `InnerClickListener`

### LINKED ISSUE #344: [ClickListener] IE11: Unable to get property 'contains' of undefined or null reference
We(IAM) are seeing this error message from ClickListener.js on IE.  Probably needs to do a check before calling .contains() 

See how React is doing it: https://zhenyong.github.io/react/docs/more-about-refs.html

screen shot 2017-11-15 at 11 29 09 am

Unable to get property 'contains' of undefined or null reference
main-bundle-6369fb459875a1f35b05.js, Line: 25, Column: 18507

### LINKED ISSUE #425: Dropdown/overflow menu/toolbar search: Extra uncustomizable div in dom tree

joshblack:
Closes #425, Closes #344

Update `ClickListener` and introduce `InnerClickListener`.

#### Changelog

**New**

- `InnerClickListener`
  - Adds ability to listen to nodes inside of a component rather than the top-level node

**Changed**

- `ClickListener`
  - removes wrapping `div` node and adds ref directly to child component

**Removed**

### DIFF
--- a/src/internal/ClickListener.js
+++ b/src/internal/ClickListener.js
@@ -1,13 +1,24 @@
 import PropTypes from 'prop-types';
-/* global document */
-
 import React from 'react';
 
-class ClickListener extends React.Component {
+/**
+ * Generic component used for reacting to a click event happening outside of a
+ * given `children` element.
+ */
+export default class ClickListener extends React.Component {
   static propTypes = {
-    children: PropTypes.node.isRequired,
+    children: PropTypes.element.isRequired,
     onClickOutside: PropTypes.func.isRequired,
   };
+
+  constructor(props) {
+    super(props);
+    // We manually bind handlers in this Component, versus using class
+    // properties, so that we can properly test the `handleRef` handler with
+    // enzyme.
+    this.handleRef = this.handleRef.bind(this);
+    this.handleDocumentClick = this.handleDocumentClick.bind(this);
+  }
 
   componentDidMount() {
     document.addEventListener('click', this.handleDocumentClick);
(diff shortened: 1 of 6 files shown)


src/internal/ClickListener.js
```diff
@@ -1,14 +1,25 @@
...
-    children: PropTypes.node.isRequired,
+    children: PropTypes.element.isRequired,
     onClickOutside: PropTypes.func.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    // We manually bind handlers in this Component, versus using class
+    // properties, so that we can properly test the `handleRef` handler with
+    // enzyme.
```
asudoh: Curious what the issue was with Enzyme?

joshblack: With class properties, the methods are only available on the instance versus the prototype, so we can't spy on it before creation time 😞

asudoh: Got it @joshblack - Thank you for explaining. Would this work? https://stackoverflow.com/a/39819539/1842550

joshblack: We totally could, would prefer to not extend React classes though.

asudoh: @joshblack You are right that extending React component class is not a good practice. The link suggests only extending the class in tests 😉

joshblack: Haha, totally understand 😄  If the `bind` in constructor is too much, can definitely add that in!

asudoh: Cool @joshblack thank you for understanding - Would be great to avoid having to change our code for the sake of tests, especially if we have a way to have tools/tactics in test solve the issue. (Note: It doesn’t mean we should introduce inherently hard-to-test code like globals, too much React states, etc.) Thanks! 😊

joshblack: The only worry that I have is that the bind pattern is pretty common, whereas the test pattern is obscure. The cost of moving to bind in a constructor seems minor, especially since it's semantic given that we want to inspect a method on a prototype versus on a class instance.

asudoh: @joshblack Fair enough - Though my link was a top hit of https://www.google.co.jp/search?q=class+field+arrow+function+spy+stub+test+mount+react, different people seems to have different opinions: https://github.com/facebook/react/issues/9851#issuecomment-306221157 https://github.com/airbnb/enzyme/issues/365#issuecomment-217368222

PR #435, opened 2018-01-03
Add `ListBox` component

joshblack:
Extracted `ListBox` and other relevant bits from #332 for a simpler diff.

#### Changelog

**New**

- `ListBox`: see [`README.md`](https://github.com/carbon-design-system/carbon-components-react/pull/435/files#diff-8d1d10a4d2221059c29d4000987f7dfb)
- `prop-types`
  - `childrenOf` `prop` validator for children of a given set of types
  - `childrenOfType` `prop` validator for children of a singletype
  - `tools`
    - `createChainableTypeChecker`: chaining required for supporting `isRequired`
    - `getDisplayName`: get name for a given React element or component

**Changed**

**Removed**

### DIFF
--- a/src/components/ListBox/ListBoxField.js
+++ b/src/components/ListBox/ListBoxField.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import ListBoxMenuIcon from './ListBoxMenuIcon';
+import ListBoxSelection from './ListBoxSelection';
+import childrenOf from '../../prop-types/childrenOf';
+
+/**
+ * `ListBoxField` is responsible for creating the containing node for valid
+ * elements inside of a field. It also provides a11y-related attributes like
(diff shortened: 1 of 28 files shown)


src/components/ListBox/ListBoxField.js
```diff
@@ -0,0 +1,21 @@
...
+ * `role` to make sure a user can focus the given field.
+ */
+const ListBoxField = ({ children, ...rest }) => (
+  <div role="button" className="bx--list-box__field" tabIndex="0" {...rest}>
+    {children}
+  </div>
+);
+
+ListBoxField.propTypes = {
+  children: childrenOf([ListBoxMenuIcon, ListBoxSelection]),
```
joshblack: Note: this might be too restrictive, I think an `input` node might also be valid for things like Combobox.

PR #439, opened 2018-01-03
fix(textInput): Add aria-invalid and aria-describedBy to invalid textInput

alisonjoseph:
Closes carbon-design-system/carbon-components-react#424

Adds `aria-invalid` to invalid textInput and `aria-describedBy` to the error message associated with the text field.


alisonjoseph:
Good idea, I'll update it 😄

PR #442, opened 2018-01-04
chore(package.json): explicitly declare support for React 15

asudoh:
To avoid the confusion on what React version carbon-components-react 5.x supports.

**Changed**

`peerDependencies` in package.json adding React 15 back.


joshblack:
Would be worried about merging this in now, given that we have the Babel v7 PR that would let us use fragments, and certain components might have us use the new portal behavior. Seems like it could change pretty quickly in the future, potentially 😞 

What is the big blocker for teams when moving from v15 to v16? Had assumed there were no changes teams would have to make.

asudoh:
@joshblack Though I didn’t look at it myself, one report I got from Catalog team was that they had to rewrite tests, which took a dev’s whole day. I guess it has something to do with Enzyme, but not fully sure.

joshblack:
@asudoh totally. Since it's going on 5 months since the release of React 16 as well. Is there an anticipated date when they would switch over?

asudoh:
@joshblack I guess we need to tap the teams to see when they can migrate. Another approach we can take is maintaining our `4.x` - The dev in Catalog had simply requested for https://github.com/carbon-design-system/carbon-components-react/pull/427 and got an impression that he had to upgrade to React 16 to get the fix.

PR #447, opened 2018-01-08
fix(interior-left-nav): refactored so it accepts react router links

marijohannessen:
Closes https://github.com/carbon-design-system/carbon-components-react/issues/169

Refactored `InteriorLeftNavItem` to have more customizable children and to accept both normal links and react-router links.

### Changelog

**Changed**

- The parent does not handle the `href` anymore
- The child can be either an anchor link or a react-router Link

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,8 +112,6 @@
     "flatpickr": "4.2.2",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
-    "react-router": "^4.2.0",
-    "react-router-dom": "^4.2.2",
     "warning": "3.0.0",
     "window-or-global": "^1.0.1"
   },
--- a/package.json
+++ b/package.json
@@ -112,6 +112,8 @@
     "flatpickr": "4.2.2",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
+    "react-router": "^4.2.0",
+    "react-router-dom": "^4.2.2",
     "warning": "3.0.0",
     "window-or-global": "^1.0.1"
   },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -111,6 +111,8 @@
     "flatpickr": "4.2.2",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
+    "react-router": "^4.2.0",
```
joshblack: Are these dependencies or do we just need them in `devDependencies`?

src/components/InteriorLeftNav/InteriorLeftNav-story.js
```diff
@@ -4,6 +4,9 @@ import InteriorLeftNav from '../InteriorLeftNav';
 import InteriorLeftNavItem from '../InteriorLeftNavItem';
 import InteriorLeftNavList from '../InteriorLeftNavList';
 
+// For demo purposes
+import { BrowserRouter, Link } from 'react-router-dom';
```
joshblack: Should we add these into a test too where we verify `Link` works inside of ILNI?

joshblack:
We should make an issue for next sprint to update ILN in `carbon-addons-cloud` with these changes, I think, just so it doesn't get lost 😄

package.json
marijohannessen: So I basically just added this to test that it would work with react-router, but what we could do is to remove it and then add a Code Sandbox example instead showing people how to use it with `react-router`

src/components/InteriorLeftNav/InteriorLeftNav-story.js
marijohannessen: If we go with the solution proposed above, I'll remove this :)

marijohannessen:
Yeah, I agree! That shouldn't take too long 😄

package.json
joshblack: If we move it to `devDependencies` seems like it'd work with what you have already, right? And then it's not included in the prod dependencies for the package.

marijohannessen: I'd prefer to just not have it added though - if anything we can add it to the add-on repo instead 🙂

tw15egan:
Looks good! 👍 ✅

PR #449, opened 2018-01-09
fix(OverflowMenu): reposition menu upon change in open state

### LINKED ISSUE #433: Overflow menu position misaligned when scrolling down on page
When scrolled partway down on the page, the overflow menu doesn't open in the expected position. Alignment is fine if the page has not been scrolled down, see attachment. This appears on Firefox and Chrome, using carbon-components-react v5.1.5. Looks like issue #190 proposes a fix for this.

[movingdropdownlist (2).zip](https://github.com/carbon-design-system/carbon-components-react/files/1598162/movingdropdownlist.2.zip)

asudoh:
Fixes #433.

#### Changelog

**New**

Code to recalculate the position of the trigger button, upon opening overflow menu.

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu-test.js
+++ b/src/components/OverflowMenu/OverflowMenu-test.js
@@ -85,6 +85,7 @@ describe('OverflowMenu', () => {
 
       expect(menuOptions.hasClass(openClass)).not.toEqual(true);
       rootWrapper.setState({ open: true });
+      rootWrapper.update();
       expect(rootWrapper.find('ul').hasClass(openClass)).not.toEqual(false);
     });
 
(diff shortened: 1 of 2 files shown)


src/components/OverflowMenu/OverflowMenu-test.js
```diff
@@ -3,6 +3,10 @@ import OverflowMenu from '../OverflowMenu';
 import Icon from '../Icon';
 import { shallow, mount } from 'enzyme';
 
+window.requestAnimationFrame = cb => {
```
joshblack: Does this line not work for rAF? https://github.com/carbon-design-system/carbon-components-react/blob/master/config/polyfills.js#L20

asudoh: Oh didn’t notice we have had the same thing already! Thanks @joshblack for pointing it out!

PR #454, opened 2018-01-10
feat: add ComboBox component

joshblack:
Adds in the Combobox component!

#### Changelog

**New**

- `src/components/Combobox`

**Changed**

- `<ListBox.Field>` now accepts `input` as a valid child

**Removed**


joshblack:
For those reviewing, the styles need to be updated in `carbon-components` first for this PR to look 👌  visually 😄

joshblack:
PR added: https://github.com/carbon-design-system/carbon-components/pull/496

joshblack:
TODO:

- [ ] Rename `Combobox` -> `ComboBox` to follow conventions
- [ ] Add `.bx--form-item` to class names, I think

psachs21:
@joshblack Apologies for sounding like a broken record, but are there plans to support a non-static set of items? So, say you want to make a API call based on the filter value. To me, this seems like an obvious, if not default, use case, given the idea of a combobox is to filter a large set of data/

joshblack:
Hi @psachs21! There are definitely plans to support this, for sure! This is just the first wave of work coming in, since the original requests coming for the ComboBox didn't require a data-fetching pattern 😄

psachs21:
@joshblack excellent. Once this is merged, I can submit a PR with an initial API and we can discuss the API details there. Thanks!

joshblack:
For those reviewing, should be good to go with `carbon-components@8.4.0` included 😄

PR #456, opened 2018-01-10
feat: Indeterminate checkbox

### LINKED ISSUE #448: Indeterminate Checkbox
Add Indeterminate Checkbox 

https://github.com/carbon-design-system/carbon-components/issues/358

alisonjoseph:
Adds indeterminate checkbox 

resolves #448

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -15,23 +15,22 @@ const Checkbox = ({
 
   return (
     <div className="bx--form-item bx--checkbox-wrapper">
+      <input
+        {...other}
+        type="checkbox"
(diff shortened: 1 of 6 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -1,43 +1,37 @@
...
 import classNames from 'classnames';
-import Icon from '../Icon';
 
 const Checkbox = ({
   className,
   id,
   labelText,
   onChange,
-  iconDescription,
+  indeterminate,
```
joshblack: Should we default this prop to `false`? Or should we just add a check to the ref method that `indeterminate` is not `undefined` before applying it?

PR #458, opened 2018-01-11
fix(OverflowMenu): clobber DOM for closed overflow menu

### LINKED ISSUE #420: [OverflowMenu]: Menu always in DOM even when not open
## Detailed description

The OverflowMenu always puts the menu in the DOM even when the menu is closed.  This is true for both normal and floating menu.  This causes performance issues when OverflowMenu is used multiple times.

OverflowMenu

The menu should only be put in the DOM when the open.

Chrome

carbon-components 8.1.8
carbon-components-react 4.2.0

IBM Watson Conversation

1. Use an OverflowMenu
2. Use the browser debugger to inspect the DOM, and notice that the menu is always there

asudoh:
Fixes #420. Wanted to see if this kind of change makes sense.

#### Changelog

**Changed**

With this change, the DOM for overflow menu is not rendered until it's open.

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -122,17 +122,19 @@ export default class OverflowMenu extends Component {
       ...other
     } = this.props;
 
+    const { open } = this.state;
+
     const overflowMenuClasses = classNames(
       this.props.className,
       'bx--overflow-menu',
       {
-        'bx--overflow-menu--open': this.state.open,
+        'bx--overflow-menu--open': open,
       }
     );
 
     const overflowMenuOptionsClasses = classNames('bx--overflow-menu-options', {
       'bx--overflow-menu--flip': this.props.flipped,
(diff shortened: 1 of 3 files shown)


src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -164,7 +166,7 @@ export default class OverflowMenu extends Component {
             description={iconDescription}
             style={{ width: '100%' }}
           />
-          {floatingMenu ? (
+          {!open ? null : floatingMenu ? (
```
joshblack: Could we extract the ternary logic into a function for readability?

marijohannessen: Ooops just realized I merged this before this comment was addressed 😢 I can revert it if you want?!

asudoh: No worries @marijohannessen! 🙂

@joshblack Thank you for giving it a thought, did some clean-up there: https://github.com/carbon-design-system/carbon-components-react/pull/465

PR #460, opened 2018-01-11
chore(Prettier): Upgrade Prettier

asudoh:
Prettier changes DatePicker.js upon pre-commit hook. This change is for prevent that from happening.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
     "node-sass": "4.6.0",
     "object-assign": "^4.1.1",
     "postcss-loader": "^2.0.5",
-    "prettier": "^1.7.4",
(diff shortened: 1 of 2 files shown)


joshblack:
@asudoh since we use `prettier` to check for differences, it looks like the CI is failing. Do you want to just add a `// prettier-ignore` comment (or whatever the syntax is) so that this doesn't happen?

asudoh:
@joshblack Sorry for the CI fail, it was because of stale Prettier version - Fixed it.

joshblack:
Hi @asudoh! Do you want to update this so that we can merge it in?

asudoh:
@joshblack Did it - Thank you for pinging!

PR #465, opened 2018-01-11
chore(OverflowMenu): hoist menu options nodes

asudoh:
Refs #420.

#### Changelog

**Changed**

- Hoisted the code to generate menu options nodes.

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -147,8 +147,19 @@ export default class OverflowMenu extends Component {
         closeMenu: this.closeMenu,
       })
     );
-    const options = !open ? null : (
+
+    const menuBody = (
       <ul className={overflowMenuOptionsClasses}>{childrenWithProps}</ul>
+    );
+    const wrappedMenuBody = !floatingMenu ? (
+      menuBody
+    ) : (
+      <FloatingMenu
+        menuPosition={this.state.menuPosition}
+        menuDirection="bottom"
+        menuOffset={flipped ? menuOffsetFlip : menuOffset}>
+        {menuBody}
+      </FloatingMenu>
     );
 
     return (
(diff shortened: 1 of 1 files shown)


src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -166,17 +169,15 @@ export default class OverflowMenu extends Component {
             description={iconDescription}
             style={{ width: '100%' }}
           />
-          {!open ? null : floatingMenu ? (
+          {!options || !floatingMenu ? (
```
joshblack: This is really challenging to parse, what is the rendering logic that we need to express here?

asudoh: @joshblack Made a bit of change to make it more idiomatic - Hope it makes sense: https://github.com/carbon-design-system/carbon-components-react/pull/465/commits/c7296d0249be6d7d6df8a8458d7e27d87f75c74c

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -166,17 +169,15 @@ export default class OverflowMenu extends Component {
             description={iconDescription}
             style={{ width: '100%' }}
           />
-          {!open ? null : floatingMenu ? (
+          {!containerWithChildrenWithProps || !floatingMenu ? (
```
joshblack: Is this block basically saying the following:

```
{!open &&
  !floatingMenu
    ? containerWithChildrenWithProps
    : <FloatMenu />
}
```

asudoh: @joshblack Oops I had a mistake translating the old logic to the new one. Fix pushed: https://github.com/carbon-design-system/carbon-components-react/pull/465/commits/9d2120deba9baf6c6423adbf2ce6360bd026a7ea

To address your original question - It’s rendering the menu body when it’s open, `null` otherwise. The menu body is wrapped as floating menu as necessary.

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -166,18 +180,7 @@ export default class OverflowMenu extends Component {
...
-              menuDirection="bottom"
-              menuOffset={flipped ? menuOffsetFlip : menuOffset}>
-              <ul className={overflowMenuOptionsClasses}>
-                {childrenWithProps}
-              </ul>
-            </FloatingMenu>
-          ) : (
-            <ul className={overflowMenuOptionsClasses}>{childrenWithProps}</ul>
-          )}
+          {!open ? null : wrappedMenuBody}
```
joshblack: Is the following equivalent:

```
{open && wrapperMenuBody}
```

asudoh: @joshblack If React stop rendering for any falsy values, yes. (Was unsure about that)

joshblack:
Hi @asudoh! Do you want to update this so that we can merge it in?

asudoh:
Did it - Thank you for pinging @joshblack!

PR #467, opened 2018-01-11
fix(breadcrumb): refactored so it accepts react router links

### LINKED ISSUE #279: BreadcrumbItem component should accept react-router Links
We recently added the `<Breadcrumb>` component to our app. It initially worked fine, until we implemented `react-router`.

We would like the `<BreadcrumbItem>` component to work with `react-router` links, and not strictly `<a>` tags.

-------

We have code something like this:
```
import { Link } from 'react-router-dom';
import { Breadcrumb, BreadcrumbItem } from 'carbon-components-react';

...

<Breadcrumb>
  <BreadcrumbItem>
    <Link to="/documents">All documents</Link>
  </BreadcrumbItem>
</Breadcrumb>
```

Which causes the following warning:

```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
    in a
    in Link
    in a
    in Link
    in div
    in BreadcrumbItem
    in div
    in Breadcrumb
```

marijohannessen:
Closes #279 

Refactored BreadcrumbItem to accept `react-router` links without showing a warning. 

#### Changelog

**Changed**

- `BreadcrumbItem`

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb-story.js
+++ b/src/components/Breadcrumb/Breadcrumb-story.js
@@ -19,8 +19,10 @@ storiesOf('Breadcrumb', module).addWithInfo(
     `,
   () => (
     <Breadcrumb {...additionalProps}>
-      <BreadcrumbItem href="#">Breadcrumb 1</BreadcrumbItem>
-      <BreadcrumbItem href="w#">Breadcrumb 2</BreadcrumbItem>
+      <BreadcrumbItem>
+        <a href="/#">Breadcrumb 1</a>
+      </BreadcrumbItem>
+      <BreadcrumbItem href="#">Breadcrumb 2</BreadcrumbItem>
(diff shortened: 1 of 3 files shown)


src/components/Breadcrumb/Breadcrumb-story.js
```diff
@@ -19,8 +19,10 @@ storiesOf('Breadcrumb', module).addWithInfo(
     `,
   () => (
     <Breadcrumb {...additionalProps}>
-      <BreadcrumbItem href="#">Breadcrumb 1</BreadcrumbItem>
-      <BreadcrumbItem href="w#">Breadcrumb 2</BreadcrumbItem>
+      <BreadcrumbItem>
+        <a href="/#">Breadcrumb 1</a>
```
marijohannessen: Added this to show that you can now add a child to the BreadcrumbItem, and if you do then it won't generate an extra `a` tag. Tested it with `react-router` Link as well, and it worked without any warnings.

PR #471, opened 2018-01-11
Add onSelectionChange prop to Tabs component

michead:
This  pull request introduces `onSelectionChange` prop in `Tabs` component in order to let consumers execute custom logic whenever a different tab is selected.
This prop would turn useful whenever showing the tab content is not the only action to trigger on tab selection.

#### Changelog

**New**

- Add `onSelectionChange` prop to `Tabs` component

**Removed**

- Remove `selectedLabel` from `Tabs`' state as it's now been made redundant by `selected`
- Remove `firstSelectedLabel` from `Tabs`' `propTypes` as it was never used

### DIFF
--- a/package.json
+++ b/package.json
@@ -242,6 +242,7 @@
       "^(?!.*\\.(js|jsx|css|json)$)": "<rootDir>/config/jest/fileTransform.js"
     },
     "testPathIgnorePatterns": [
+      "/node_modules/",
       "/examples/",
       "/config/",
       "/lib/",
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     }
   },
   "peerDependencies": {
-    "carbon-components": "^8.2.0",
+    "carbon-components": "^8.4.0",
     "carbon-icons": "^5.1.2 || ^6.0.0",
     "react": "^15.3.2 || ^16.1.0",
     "react-dom": "^15.3.2 || ^16.1.0"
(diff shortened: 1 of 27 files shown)


michead:
@asudoh thanks for your review! The latest commit should address your point 👍

joshblack:
Thanks so much @michead for taking the time to contribute this! It's incredibly helpful 😄

Just taking a look at this now, is there a specific issue that helps us understand the motivations behind these changes?

michead:
Hi @joshblack, my pleasure -- this is a such great project!
I did not create an issue which explains the motivations behind these changes -- these are explained in short in the PR description.
In particular, currently the consumers of the `Tabs` component have no way to be notified whenever a different tab is selected.
So, I think adding the `onSelectionChange` prop would make the component way more flexible by letting consumers execute custom logic on tab selection.
If you prefer I can create an issue for it and then reference it here.

asudoh:
@michead Thank you for making the `ref` change! With this change, I think we need to:

1. Clean-up the refs upon unmounting
2. Existence checks of what `getTabAt()` returns as well as `tabAnchor` (Either making the referring code no-op or throwing)

Thanks!

BTW agreed that the consumer of tab, menu, etc. components need to have selection change event (callback in React case) unless they relegate selection entirely to consumers (via props). I see current `Tab` manages selection in its state.

michead:
Hi @asudoh, thanks again for the review!
I added the checks for the return value of `getTabAt()` in the latest commit.
As regards cleaning up the `refs` upon unmounting -- isn't that taken care of already?

From the [docs](https://reactjs.org/docs/refs-and-the-dom.html):
> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts.

Am I missing something? Thanks! :)

PR #501, opened 2018-01-17
fix(components): updated carbon components to fix checkboxes

marijohannessen:
Updated the version of `carbon-components` to fix the alignment of the checkboxes in the tables.


marijohannessen:
Sure thing! 👍

PR #504, opened 2018-01-17
fix(date-picker): reset calendar when inputs are cleared

### LINKED ISSUE #402: Is there any way to reset range DatePicker ?
## Detailed description

Now I need to reset DataPicker (range type) - it means after I select a date range, I need to reset it to the initial state without re-rendering the component. 

DatePicker/DatePickerInput

`<DatePickerInput
    labelText="Start date"
    onChange={this.handleDateChange}
    onClick={this.handleFromToClick}
    placeholder="mm/dd/yyyy"
    id="fromDate"
    value={this.state.startDate}
    />`
Now I bind a property value with a property "startDate" in local state, by reset startDate value in local state, the input text is gone, but in the popup, the date range is still there. The expected behavior is there should be a way to "completely" reset the component - the date range in popup should be reset as well. 

image

Chrome

7.26.10

Yes. It is better to be fixed ASAP.

## Additional information

* Screenshots or code
* Notes

## Add labels
Please choose the appropriate label(s) from our existing label list to ensure that your issue is properly categorized. This will help us to better understand and address your issue.

marijohannessen:
Closes #402 

Fixed DatePicker so when the inputs are cleared, the calendar gets reset.

#### Changelog

**Changed**

- `DatePicker.js`

### DIFF
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -106,7 +106,7 @@ export default class DatePicker extends Component {
       this.toInputField.addEventListener('blur', () => {
         this.cal.close();
       });
-      this.toInputField.addEventListener('input', () => {
+      this.toInputField.addEventListener('change', () => {
         if (
           this.toInputField.value === '' &&
           cal.selectedDates.length > 0 &&
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -88,7 +88,17 @@ export default class DatePicker extends Component {
     ) {
       this.cal.destroy();
     }
+    this.inputField.removeEventListener(this.onChange);
+    if (this.toInputField) {
+      this.toInputField.removeEventListener(this.onChange);
+    }
   }
+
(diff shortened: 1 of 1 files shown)


src/components/DatePicker/DatePicker.js
```diff
@@ -97,10 +97,24 @@ export default class DatePicker extends Component {
         cal.calendarContainer.focus();
       }
     });
+    input.addEventListener('change', () => {
```
asudoh: Given this line deals with DOM directly, the event listeners need to be be removed when the React component is unmounted. [`carbon-components/es/globals/js/misc/on`](https://github.com/carbon-design-system/carbon-components/blob/a4b23e4/src/globals/js/misc/on.js) may help - e.g. `this.hChange = on(input, 'change', () => { … })` when mounting and `if (this.hChange) { this.hChange = this.hChange.release(); }` when unmounting.

marijohannessen: Ahh right, I'll make the change, thanks!

PR #509, opened 2018-01-18
fix(DataTable): handle sorting for different primitive types

ryanomackey:
Hi guys! I've been working with the new DataTable and wanted to contribute this upstream. Currently, the sorting only works if the table data is a string. This just checks what the primitive type is and sorts it appropriately.

I didn't think it made sense to sort boolean values, but if you disagree, let me know.

#### Changelog

**New**

- N/A

**Changed**

- Fixed a bug where the v2 table couldn't sort non-string values.

**Removed**

- N/A

### DIFF
--- a/src/components/DataTable/DataTable-story.js
+++ b/src/components/DataTable/DataTable-story.js
@@ -34,26 +34,26 @@ const initialRows = [
   {
     name: 'Load Balancer 3',
     protocol: 'HTTP',
-    something: '80',
+    something: 80,
     rule: 'Round Robin',
     attached_groups: 'Kevins VM Groups',
-    status: 'Active',
+    status: true,
   },
   {
     name: 'Load Balancer 1',
     protocol: 'HTTP',
-    something: '80',
+    something: 79,
     rule: 'Round Robin',
     attached_groups: 'Maureens VM Groups',
-    status: 'Active',
+    status: false,
   },
   {
     name: 'Load Balancer 2',
     protocol: 'HTTP',
-    something: '80',
+    something: 42,
     rule: 'Round Robin',
     attached_groups: 'Andrews VM Groups',
-    status: 'Active',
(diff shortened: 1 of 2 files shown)


joshblack:
Might be helpful to reconcile the sorting stuff with how Filterable MultiSelect is doing it in its PR:

PropTypes:

https://github.com/joshblack/carbon-components-react/blob/85543e6b233f1e208bc4f9256186f0cd13d24898/src/components/MultiSelect/MultiSelectPropTypes.js

Sorting Utility:

https://github.com/joshblack/carbon-components-react/blob/85543e6b233f1e208bc4f9256186f0cd13d24898/src/components/MultiSelect/tools/sorting.js

Or, at least, instead of doing assertions on the type I believe we can just do this with `localeCompare`:

```
itemA.localeCompare(itemB, locale, {
  numeric: true,
});
```

So that it can handle number comparisons 😄

ryanomackey:
Hey @joshblack, @asudoh. I pushed up a quick refactor, but that's probably all I have time for.

Turns out the `numeric` option won't work with numbers. It just sorts numeric strings in proper order, i.e. "1" < "2" < "10".

Still probably good to enable, but unrelated to the issue.

joshblack:
Hey @ryanomackey! The work introduced with the following PR might address this, specifically: https://github.com/carbon-design-system/carbon-components-react/pull/562/files#diff-c76ca798b46b59c76e2028165a12be50R15

Let me know if this would help solve your use-case!

ryanomackey:
Absolutely. Thanks, @joshblack. Any ETA on the rollout?

joshblack:
100% will be end of this week @ryanomackey 😄  should include:

- Filter by search bar
- Sorting
- Row Expansion
- Row selection
- Batch actions called with row selections

PR #514, opened 2018-01-19
fix(OrderSummary): Add aria-labelledby, missing attributes fail A11Y tests

### LINKED ISSUE #513: A11Y: OrderSummary component missing aria-label or aria-labelledby
## Detailed description

Elements containing a 'region' role and section element must be labeled with an aria-label or aria-labelledby to facilitate page navigation by screen reader users.

OrderSummary, OrderSummaryTotal, OrderSummaryFooter

All of the components listed above product a `<section>` tag that must have an aria-label or aria-labelledby.

Chrome Version 63.0.3239.132 (Official Build) (64-bit) 
OS X (El Capitan)

v3.14.0

Instantiate any of the components, check the dom, notice that the section tags are missing the aria labels.

zeusorion:
Resolves #513.

Added aria-labelledby to OrderSummary* components, tests shown to fail in issue #513 are now passing.

### DIFF
--- a/src/components/OrderSummary/OrderSummary.js
+++ b/src/components/OrderSummary/OrderSummary.js
@@ -25,20 +25,24 @@ export class OrderSummaryHeader extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
+    id: PropTypes.string,
     title: PropTypes.string,
   };
 
   static defaultProps = {
+    id: 'order__summary__header__id',
     title: 'Order Summary',
   };
 
   render() {
-    const { children, className, title, ...other } = this.props;
+    const { children, className, id, title, ...other } = this.props;
     const classes = classNames('bx--order-header', className);
 
     return (
-      <section className={classes} {...other}>
-        <p className="bx--order-header-title">{title}</p>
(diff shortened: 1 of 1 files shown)


src/components/OrderSummary/OrderSummary.js
```diff
@@ -25,20 +25,24 @@ export class OrderSummaryHeader extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
+    id: PropTypes.string,
     title: PropTypes.string,
   };
 
   static defaultProps = {
+    id: 'order__summary__header__id',
```
asudoh: Given hard-coded ID tends to cause ID duplication (though you made it configurable), I’d suggest making it random, e.g. `order__summary__header__id_${Math.random().toString(36).substr(2)}`. Same for `order__summary__footer__id`.

zeusorion:
@asudoh made the suggested changes, made the `id` random to avoid possible duplication. thx.

asudoh:
Thanks @zeusorion! My apologies not mentioning that the default props are static, meaning reused fr every component instance. So good to have them as default values destructuring `id` from `props`. Thanks!

PR #520, opened 2018-01-22
[RFC] feat: support UMD

asudoh:
This change supports UMD in the build in our `lib` directory. This allows consumers to get quick prototype with `carbon-components-react` without scaffolding the build tool-chain.

#### Changelog

**Changed**

- CJS build now support UMD, while keeping compatible as CJS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 node_modules
 es
 lib
+umd
 storybook-static
 
 # Logs
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
   ],
   "files": [
     "lib/**/*",
-    "es/**/*"
+    "es/**/*",
+    "umd/**/*"
   ],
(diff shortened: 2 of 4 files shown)


joshblack:
Thanks for taking the time to do this! Couple questions:

- Should we have a dedicated `umd` directory for this?
- How does it impact the bundle size of people using the `lib` directory now?

asudoh:
@joshblack Interesting to see that you are more interested in bundle for UMD, which makes sense given the purpose we’d have with UMD (consumers without build toolchain). Unless anybody objects against doing so, I’ll go that route. Thank you for your feedback!

asudoh:
Made an update to make the UMD stuffs as a bundle. Also added sizes report as @joshblack suggested, though I didn't have time to old/new comparison stuff. Thanks @joshblack for your suggestions!

PR #524, opened 2018-01-22
Fix(dropdown): handle null children

jamesvclements:
If the `Dropdown` is given `null` children, it breaks on rendering. You can see this behavior by editing the `story` for `Dropdown`:

```javascript
.addWithInfo(
    'with default text',
    `
      The Dropdown component is used for navigating or filtering existing content.
      Create Dropdown Item components for each option in the dropdown menu.
    `,
    () => (
      <Dropdown
        {...dropdownEvents}
        onChange={action('onChange')}
        defaultText="Dropdown label">
        <DropdownItem itemText="Option 1" value="option1" />
        <DropdownItem itemText="Option 2" value="option2" />
        {null}
        <DropdownItem itemText="Option 3" value="option3" />
        <DropdownItem itemText="Option 4" value="option4" />
        <DropdownItem itemText="Option 5" value="option5" />
      </Dropdown>
    )
  )
```

Results in this error:
image

The `React.Children` methods don't verify whether the child actually exists, so we added this safety net to make sure first. We ran into this because of some Dropdowns we had that were being populated with both synchronous & asynchronous information, and the `map` for the async data was temporarily returning `null`.

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -42,8 +42,8 @@ export default class Dropdown extends PureComponent {
     let matchingChild;
     React.Children.forEach(children, child => {
       if (
-        child.props.itemText === selectedText ||
-        child.props.value === value
+        child &&
+        (child.props.itemText === selectedText || child.props.value === value)
       ) {
         matchingChild = child;
       }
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/Dropdown.js
```diff
@@ -96,14 +96,17 @@ export default class Dropdown extends PureComponent {
...
-      React.cloneElement(child, {
-        onClick: (...args) => {
-          child.props.onClick && child.props.onClick(...args);
-          this.handleItemClick(...args);
-        },
-      })
-    );
+    const children = React.Children
+      .toArray(this.props.children)
+      .filter(child => child)
```
asudoh: Instead of creating a new function here, we may probably use `Boolean` (as a function)?

PR #526, opened 2018-01-23
feat(Icon): Add support for Custom icons

asfordmatt:
Closes carbon-design-system/carbon-components-react#515

Add support for custom icons using an 'xlink' prefix in the icon name

#### Changelog

**New**

- Add support for custom icons using an 'xlink' prefix in the icon name

### DIFF
--- a/src/components/Icon/Icon-story.js
+++ b/src/components/Icon/Icon-story.js
@@ -11,16 +11,47 @@ const props = {
   className: 'extra-class',
 };
 
-storiesOf('Icon', module).addWithInfo(
-  'Default',
-  `
-    Icons are used in the product to present common actions and commands. Modify the fill property to change the color of the icon. The name property defines which icon to display. For accessibility, provide a context-rich description with the description prop. For a full list of icon names, see carbondesignsystem.com/style/iconography/library
-  `,
-  () => (
-    <div>
-      <Icon name="icon--add" {...props} />
-      <Icon name="add--glyph" {...props} />
-      <Icon name="add--outline" {...props} />
-    </div>
+storiesOf('Icon', module)
+  .addWithInfo(
+    'Default',
+    `
+      Icons are used in the product to present common actions and commands. Modify the fill property to change the color of the icon. The name property defines which icon to display. For accessibility, provide a context-rich description with the description prop. For a full list of icon names, see carbondesignsystem.com/style/iconography/library
+    `,
+    () => (
+      <div>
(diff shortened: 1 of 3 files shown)


asfordmatt:
@asudoh the key point of this approach is to support all the other components that take an icon name prop and pass that to an Icon component - such as Button, OverflowMenu, CardContent, Tooltip etc.  Otherwise each one of those components would need to be enhanced individually to support rendering a custom icon instead of an Icon component. This way every component that takes an icon name gains support for custom icons in one go.

tw15egan:
If we're going to expand support for the `Icon` component, I think we need to support any `svg` icon, not specifically `svg` sprites.

asudoh:
@asfordmatt OK good attempt trying to kill many birds with one stone! And great that you listed the components that need custom icon support. Also @tw15egan has a valid point - The fundamental requirement is to allow any `<svg>` icon. Given that, even though it’s more effort, we’d change the components you listed to support custom icons.

asudoh:
^ Another way I can think of is letting `<Icon>` accept something like `<Icon><title>…</title><path>…</path></Icon>`, though the caller of `<Icon>` (e.g. `<CardContent>`) seems to still need some changes.

asfordmatt:
It would save a lot of effort if we can find a way to add support via icon name, and therefore not need to update every current component that uses Icon, plus remembering to do the same for new components. I think the use of SVG sprites is pretty common, but could another option to be to also support passing SVG content directly as a string? e.g. iconName='<svg>...</svg>'

asudoh:
^ Honestly I thought about it - And the I got a second thought of how to keep the API of `<Icon>` itself intuitive, and came up with custom icon content as child nodes. Curious what other people think.

joshblack:
Just for consensus, custom icons seem to fall in the following categories:

- Someone used `raw-loader` get the SVG source string and want to use `dangerouslySetInnerHTML` to add it
- Someone used `react-svg-loader` and now just has an icon component
- Someone needs to reference an SVG sprite sheet with a symbol
- Someone wants to use a preset icon from Carbon

Does that match the expectations that people have for custom icons? Are there any others I should include? 😄

joshblack:
Hi there! 👋 since this PR is going to require some coordination to plan out what the best API could be, let's try and use #537 to continue this discussion and see what kinds of things we could do to support this PR 😄 

Going to close this for now, feel free to re-open though if you'd rather continue the discussion on this PR!

PR #528, opened 2018-01-23
feat(Card): support custom icon

### LINKED ISSUE #515: allow user to pass a custom svg icon to CardContent
CardContent component currently only supports carbon icons. We would like to pass our own svg markup to the component and have it rendered.

asudoh:
Resolves #515.

#### Changelog

**New**

- `cardIcon` prop in `CardContent` now supports `PropTypes.node` - A custom icon (instead of `carbon-icons` sprite) in such case.

### DIFF
--- a/src/components/CardContent/CardContent.js
+++ b/src/components/CardContent/CardContent.js
@@ -1,3 +1,4 @@
+import warning from 'warning';
 import PropTypes from 'prop-types';
 import React, { isValidElement } from 'react';
 import classNames from 'classnames';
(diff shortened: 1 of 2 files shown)


src/components/CardContent/CardContent.js
```diff
@@ -37,17 +37,36 @@ const CardContent = ({
   const cardLinkContentArray = Object.keys(cardLinkContent);
   const cardInfoContentArray = Object.keys(cardInfoContent);
 
+  if (
+    isValidElement(cardIcon) &&
+    iconDescription !== CardContent.defaultProps.iconDescription
+  ) {
+    // eslint-disable-next-line no-console
+    console.warn(
```
joshblack: Could use `invariant` or `warning` here, I think, just so we can get the dev-time checks and removal in prod

asudoh: Great point @joshblack! - Did it: https://github.com/carbon-design-system/carbon-components-react/pull/528/commits/fc24f4fbae8527cf32b5c76e997e8667a7017051

PR #538, opened 2018-01-26
feat(build): update to babel v7

joshblack:
Closes carbon-design-system/carbon-components-react#375

Updates build process and related tooling to use new Babel v7 packages. This has the added advantage of adding PURE notations to our components to help with tree shacking and DCE 😄 

#### Changelog

**New**

**Changed**

- Update config to reference new Babel v7 packages

**Removed**

### DIFF
--- a/.storybook/.babelrc
+++ b/.storybook/.babelrc
@@ -1,7 +1,7 @@
 {
   "presets": [
     [
-      "env",
+      "@babel/preset-env",
       {
         "modules": false,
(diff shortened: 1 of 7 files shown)


joshblack:
@asudoh I believe we do for the export stuff that we're doing, unfortunately 😞

asudoh:
I see 😢 - Thanks @joshblack for clarifying!

PR #543, opened 2018-01-26
fix(Tile): prevent event bubbling and expose event object to callbacks

jzhang300:
**Changelog**

**New**

* Callback events from `Tile` components’ props now have access to `event` object.
* `handleClick` and `handleKeyDown` in the `SelectableTile` component now prevents event bubbling so that they don’t trigger twice.


jzhang300:
@joshblack can you elaborate what you mean?  Or point me to a line number?

src/components/Tile/Tile.js
```diff
@@ -39,11 +39,11 @@ export class ClickableTile extends Component {
     handleKeyDown: () => {},
   };
 
-  handleClick = () => {
+  handleClick = evt => {
     this.setState({
       clicked: !this.state.clicked,
     });
-    this.props.handleClick();
+    this.props.handleClick(evt);
```
joshblack: Could we move this to the callback of the `setState` call? e.g.

```js
this.setState(state => ({ clicked: !state.clicked }), () => {
  this.props.handleClick(evt);
});
```

src/components/Tile/Tile.js
```diff
@@ -52,7 +52,7 @@ export class ClickableTile extends Component {
         clicked: !this.state.clicked,
       });
     }
-    this.props.handleKeyDown();
+    this.props.handleKeyDown(evt);
```
joshblack: Same note as above

joshblack:
Added the notes @jzhang300!

jzhang300:
🎉 I finally get to put my face on that README!

PR #545, opened 2018-01-26
Update DropDownItem role to option

loganmccaul:
The role menuitem attached to DropDownItem is usually used as a child under [menu or menubar](https://www.w3.org/TR/wai-aria/roles#menuitem).

With the change to the role to listbox, #544 , the children will need to have the role option.

References:
https://www.w3.org/TR/wai-aria/roles#option

### DIFF
--- a/src/components/DropdownItem/DropdownItem.js
+++ b/src/components/DropdownItem/DropdownItem.js
@@ -9,6 +9,7 @@ const DropdownItem = ({
   onClick,
   onKeyPress,
   href,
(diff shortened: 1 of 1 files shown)


joshblack:
Thanks for the change! Seems like there is a lint violation that's causing the build to fail, unfortunately.

loganmccaul:
@joshblack Oh no. This is why I don't try to edit from github, I'll pull it down and take a look 😄

Issue #533, opened 2018-01-25
Pagination is using includes feature, but includes is not supported by IE 11


bjakob01:
## Detailed description
The IE 11 cannot render a table with pagination - the IE throws an exception: "SCRIPT438: Object doesn't support property or method 'includes'" 

The reason for this problem is, that Pagination.js uses .includes
>>>>>
...
return _ret = (_temp = (_this = _possibleConstructorReturn(this, (_ref = Pagination.proto || Object.getPrototypeOf(Pagination)).call.apply(_ref, [this].concat(args))), _this), _this.state = {
page: _this.props.page,
pageSize: _this.props.pageSize && _this.props.pageSizes.**includes**(_this.props.pageSize) ? ....
...
<<<<<

For this carbon rect component, the following browser list is defined:
"browserslist": [
   "ie >= 9",
   "last 2 versions",
   "safari >= 7"
 ],

If IE is supported, then includes must not be used and must be replaced by e.g. indexOf, or the missing includes feature must be implemented by a polyfill

> Is this issue related to a specific component?
I detected this problem just in Pagination component

> What did you expect to happen? What happened instead? What would you like to see changed?
I want to have the includes replaced by e.g. indexOf or polyfill must be implemented for includes

> What browser are you working in?
IE 11

> What version of the Carbon Design System are you using?
"carbon-components-react": "3.14.0",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
continuous

## Steps to reproduce the issue

1. Step one
set pagination properties: ... pageSize: this.state.pageSize
2. Step two
render pagination with IE 11

joshblack:
Hi there! 👋  Thanks so much for taking the time to make this issue 😄

We currently only offer support for the latest version and previous versions of major browsers, in addition to Firefox ESR. However, this issue can easily be resolved by adding the appropriate polyfill for `Array#includes`. Hope that helps!

PR #548, opened 2018-01-30
chore(devenv): load polyfill into Storybook env

asudoh:
Refs #533.

#### Changelog

**New**

- `.storybook/polyfills.js`, which let you run Storybook on IE11 as well as works as the list of required polyfills.

### DIFF
--- a/.storybook/Container.js
+++ b/.storybook/Container.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import './polyfills';
 import './_container.scss';
 
 export default class Container extends Component {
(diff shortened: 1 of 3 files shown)


joshblack:
Just curious, where is the list of polyfills coming from?

asudoh:
@joshblack Good question - Was from my quick code scan - Which is, taking the list of polyfills from core-js, esp. of frequently used ones, and doing search. After that, I quickly ran Storybook on IE and saw no JS error came up.

joshblack:
Cool! Should we document these just to make sure consumers know they should have these included?

asudoh:
@joshblack Put the document to README as discussed.

PR #561, opened 2018-01-31
fix($time-picker): show invalidText when invalid and invalidText properties are passed

scottdickerson:
Closes carbon-design-system/carbon-components-react#556

#### Changelog

**Changed**

-(TimePicker) Added data-invalid attribute whenever invalid was true so that the css classes would apply as they do to TextInput


src/components/TimePicker/TimePicker.js
```diff
@@ -128,6 +128,7 @@ export default class TimePicker extends Component {
             <input
               {...other}
               {...timePickerInputProps}
+              data-invalid={invalid ? invalid : undefined}
```
joshblack: Could this just be `data-invalid={invalid}`?

scottdickerson: we could do that, but if they pass invalid={false}, it will still show the message because the css selector that shows the message is searching for just the simple existence of data-invalid attribute.  Compare this to how the TextInput does it: https://github.com/carbon-design-system/carbon-components-react/blob/f1f0f89882a753261330f87276f268e479e7d037/src/components/TextInput/TextInput.js#L52

scottdickerson: more likely they would do a comparison but false would be the default output unless they were smart enough to return undefined themselves.  invalid={detectedErrors > 0}

PR #562, opened 2018-01-31
feat: add basic table state behavior

### LINKED ISSUE #485: Data Table: Sorting
This issue is for refining the initial behavior for sorting in a Data Table. Ideally, we'd like to provide a basic set of functionality for sorting that works for most consumers, while providing custom hooks for people to override the sorting logic.

### LINKED ISSUE #486: Data Table: Selection
This issue is for refactoring/polishing the existing selection API for `DataTable` to support the following:

- [ ] Making sure the selection API design is consistent with the planned feature development
- [ ] Passing in the given collection to any Toolbar handlers

### LINKED ISSUE #487: Data Table: Expansion
This issue is for the implementation of the expansion feature for `DataTable`, including:

- [ ] Create a Storybook story for this feature
- [ ] Implement the API for this feature and integrate this into the codebase
- [ ] Add associated tests for this feature
- [ ] Conduct self audit of any a11y issues we might encounter with whatever implementation we go with

### LINKED ISSUE #503: Data Table: toggles are visible for inactive sort fields
Based on this code example (https://codesandbox.io/s/6lyvq4nozw), the toggles for inactive sort fields are visible. If I click a field in the table header to sort by, I expect to see the toggle for that field only. In the screenshotted example, I see the toggle for any field permanently once it's clicked, even the the sort field isn't active.

screenshot 55

joshblack:
Closes #485
Closes #486
Closes #487
Closes #503
Closes #562

Moves some of the existing pieces around, plus some general clean-up. Also adds in the following behaviors to DataTable:

- Sorting
- Expansion
- Filtering

#### Changelog

**New**

- `src/components/DataTable`
- `src/components/DataTable/tools`
- `src/components/DataTable/state`
- `src/tools/wrapComponent`

**Changed**

- Moved around components from existing `DataTable`

**Removed**


joshblack:
TODO:

- [ ] Finish up documentation
  - [ ] Flesh out Storybook, make sure parity exists with what is already shown
  - [ ] Add documentation per included behavior
- [ ] Tests
  - [ ] Add missing tests
  - [ ] Update DataTable tests

joshblack:
Should subsume #509

src/components/DataTable/TableBatchActions.js
```diff
@@ -0,0 +1,45 @@
...
+    className
+  );
+
+  return (
+    <div className={batchActionsClasses} {...other}>
+      {children}
+      <div className="bx--batch-summary">
+        <p className="bx--batch-summary__para">
+          <span>{totalSelected}</span>{' '}
+          {totalSelected > 1 ? 'items selected' : 'item selected'}
```
ryanomackey: Could this be made into a prop? It'd be nice to have it for internationalization.

joshblack: Good call!

src/components/DataTable/tools/instanceId.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Generic utility to initialize a method that will return a unique instance id
+ * for a component.
+ */
+const setupGetInstanceId = () => {
```
asudoh: Great to have this code used library-wide, not just for table!

src/components/DataTable/tools/filter.js
```diff
@@ -1,5 +1,17 @@
 import { getCellId } from './cells';
 
+/**
+ * Default implemention of how we filter rows internally. The idea behind this
+ * implementation is to use the given list of row ids and headers to get the
+ * individual cell values for a row. Then, we go through each cell value and see
+ * if any of them includes the given inputValue.
+ *
+ * @param {Array[string]} rowIds array of all the row ids in the table
```
asudoh: Great to have this as something like `Array<string>` or `string[]` so that more JSDoc-based tools would understand it.

src/components/DataTable/DataTable.js
```diff
@@ -104,55 +115,104 @@ export default class DataTable extends React.Component {
...
       isExpanded: row.isExpanded,
     };
   };
 
+  /**
+   * Gets the props associated with selection for a header or a row, where
+   * applicable. Most often used to indicate selection status of the table or
+   * for a specific row.
+   *
+   * @param {Object?} row an optional row that we want to access the props for
```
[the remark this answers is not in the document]
joshblack: Thanks @asudoh!

src/components/DataTable/tools/filter.js
joshblack: Makes sense. Thanks @asudoh!

src/components/DataTable/tools/instanceId.js
joshblack: Good call, I think there is `src/tools/uniqueId` but it initializes the id variable in the module scope. We should definitely update that one to follow this pattern so that the value isn't shared across modules 😄

src/components/DataTable/README.md
```diff
@@ -0,0 +1,573 @@
...
+            </TableHead>
+            <TableBody>
+              {rows.map(row => (
+                <TableRow key={row.id}>
+                  {row.cells.map(cell => (
+                    <TableCell key={cell.id}>{cell.value}</TableCell>
+                  ))}
+                </TableRow>
+              ))}
+            </TableBody>
```
jlengstorf: It seems to me like this is a lot of structural code that could be abstracted away for an easier API.

This is great for an advanced use case, but if I just need to list some sortable data, I'd much prefer to be able to do something like this:

```
import { DataTable } from 'carbon-components-react';

const headers = [
  { key: 'name', header: 'Name' },
  { key: 'date', header: 'Date' },
];

const rows = [
  { id: 1, name: 'Foo', date: '2018-01-01' },
  { id: 2, name: 'Bar', date: '2018-02-01' },
  { id: 3, name: 'Baz', date: '2018-02-06' },
];

export default () => (
  <DataTable rows={rows} headers={headers} title="DataTable />
);
```

Any modifications would require the full markup, but I think having a "composite component" to simplify the API for default use cases would be really well received and probably cut down on your issues significantly. :)

joshblack: I think having a table that just works OOTB would be amazing 🎉 

The reason it’s being built with this approach is to allow flexibility so that it can be used in any use-case, and then we can start seeing what patterns are being used and can create dedicated components for those.

It’s also nice since the way it’s structured right now, the API you’re after can be implemented in user land 😄 

Could kind of see this methodology in how it was released:

- v1 release is just presentation, notice teams implementing ad-hoc sort/expansion/etc implementations
- v2 release consolidates basic ad-hoc implementations of state/behavior
- v3 release adds advanced behavior like inline edits
- v4 release makes it easier to use common patterns/behaviors found in the platform

We didn’t start the other way because often times teams needed direct access to inner markup in order to add handlers for the behavior we didn’t implement yet.

src/tools/wrapComponent.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+const wrapComponent = ({ name, className, type }) => {
+  const Component = props => {
+    const componentClass = cx(className, props.className);
+    return React.createElement(type, {
```
asudoh: Curious why `React.createElement()` is used instead of `<type {…props} className={!!componentClass ? componentClass : undefined} />` - Is it for the comment of `className` or something else? Also good to have more explanation of `wrapComponent()`. Thanks!

src/components/InlineCheckbox/InlineCheckbox.js
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const InlineCheckbox = ({ id, checked, name, onClick, onKeyDown }) => (
+  <React.Fragment>
```
asudoh: Are we officially dropping React 15 support?

src/tools/wrapComponent.js
joshblack: `<type>` is invalid in React as it's lowercase 😞

src/components/InlineCheckbox/InlineCheckbox.js
joshblack: I think when using DataTable, yeah. There is another Fragment in the DataTable story as well.

joshblack:
@asudoh for comments on propTypes, are these not sufficient: https://github.com/carbon-design-system/carbon-components-react/pull/562/files#diff-16b180768413647ce20626ec847d891aR25 ?

src/tools/wrapComponent.js
asudoh: Yeah given `type` is a React component/node wondered if it should be uppercased, though I may be missing something.

src/components/InlineCheckbox/InlineCheckbox.js
asudoh: OK let’s make sure we document that!

asudoh:
@joshblack Great that you added the `propTypes` comments to several components - Great if you can add ones to e.g. https://github.com/joshblack/carbon-components-react/blob/08f3e7faa63e61353543df7b363c77732c8dda1f/src/components/DataTable/TableBatchAction.js#L11, too. Thanks!

src/tools/wrapComponent.js
joshblack: Can definitely rebind it to an uppercase variable in the scope, might help with the class name check too.

src/components/InlineCheckbox/InlineCheckbox.js
joshblack: Good call 👍

src/components/DataTable/DataTable.js
```diff
@@ -1,403 +1,374 @@
...
   constructor(props) {
     super(props);
-    this.state = {
-      rows: props.initialRows,
-      sortKey: null,
-      sortDirection: null,
-    };
-    this.sortBy = this.sortBy.bind(this);
-    this.getHeaderProps = this.getHeaderProps.bind(this);
+    this.state = DataTable.getDerivedStateFromProps(props);
```
asudoh: While `Class.method()` works, `this.constructor.method()` is the way of calling static method - More robust way against change in class name, etc.

src/components/DataTable/tools/events.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param {Array<Function>} fns array of functions to apply to the event
+ * @returns {Function}
+ */
+export const composeEventHandlers = fns => (event, ...args) => {
+  for (let i = 0; i < fns.length; i++) {
+    if (event.defaultPrevented) {
+      break;
+    }
+    if (typeof fns[i] === 'function') {
```
asudoh: Just a thought - Should we emit `invariant()` warning if `fns[i]` is not `undefined` but not a function?

src/components/DataTable/index.js
```diff
@@ -1 +1,64 @@
...
+import TableHeader from './TableHeader';
+import TableRow from './TableRow';
+import TableSelectAll from './TableSelectAll';
+import TableSelectRow from './TableSelectRow';
+import TableToolbar from './TableToolbar';
+import TableToolbarAction from './TableToolbarAction';
+import TableToolbarContent from './TableToolbarContent';
+import TableToolbarSearch from './TableToolbarSearch';
+
+DataTable.Table = Table;
```
asudoh: Do we really need to make all sub-components hard-dependencies of `<DataTable>`? Woud named exports below suffice?

src/components/DataTable/TableBatchActions.js
```diff
@@ -0,0 +1,62 @@
...
+        </button>
+      </div>
+    </div>
+  );
+};
+
+TableBatchActions.translationKeys = Object.keys(translationKeys);
+
+TableBatchActions.propTypes = {
+  children: PropTypes.node,
```
asudoh: Let's document `propTypes` here! Seems that some other sub-components misses `propTypes` descriptions.

src/components/InlineCheckbox/InlineCheckbox.js
```diff
@@ -0,0 +1,27 @@
...
+      checked={checked}
+      onClick={onClick}
+      onKeyDown={onKeyDown}
+    />
+    <label htmlFor={id} className="bx--checkbox-label" />
+  </React.Fragment>
+);
+
+InlineCheckbox.propTypes = {
+  id: PropTypes.string.isRequired,
```
asudoh: Let's document `propTypes` here!

src/components/DataTable/index.js
joshblack: I consider this a temporary solution until we deprecate the old Table components from table v1. Once those are officially removed, then we expose all of these components as named exports like you said 😄

src/components/DataTable/tools/events.js
joshblack: I definitely would not mind. Love introducing `invariant` for dev-time suggestions 😄

src/components/DataTable/TableBatchActions.js
joshblack: I'll add them in, not for children and className though. Sound good?

src/components/DataTable/index.js
joshblack: Actually, thinking about this, should we just suffix them with `V2` so people can refer to the components themselves in the default exports?

src/components/DataTable/DataTable-story.js
```diff
@@ -1,503 +1,459 @@
...
 import Button from '../Button';
-import OverflowMenu from '../OverflowMenu';
-import OverflowMenuItem from '../OverflowMenuItem';
-
-const paginationProps = {
-  pageSizes: [10, 20, 30, 40, 50],
-};
 
 const initialRows = [
   {
```
marijohannessen: Small thing - but until we have a more fleshed out dummy content created, do you mind re-naming the examples to at least have different values in each row (so different port numbers, different protocols, different rules, etc)? Just so the sorting and search functionality becomes more apparent :-)

joshblack: Yupp! Makes sense to me 😄  Let me add a TODO

joshblack:
TODO:

- [ ] Update example content to provide more realistic data

src/components/DataTable/index.js
asudoh: Yes I think it makes sense :+1:

src/components/DataTable/TableBatchActions.js
asudoh: Yes it makes sense :+1:

Issue #508, opened 2018-01-18
Remove extra props being passed to components


joshblack:
In our test suite, I noticed a couple of warnings being thrown by React for us passing in extra props to certain components, namely the following:

```bash
 PASS  src/components/Tabs/Tabs-test.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Unknown event handler property `onSelectionChange`. It will be ignored.
          in nav (created by Tabs)
          in div (created by Tabs)
          in Tabs (created by WrapperComponent)
          in WrapperComponent

 PASS  src/components/Tooltip/Tooltip-test.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: React does not recognize the `clickToOpen` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `clicktoopen` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
          in div (created by Tooltip)
          in div (created by FloatingMenu)
          in FloatingMenu (created by Tooltip)
          in div (created by Tooltip)
          in Tooltip (created by WrapperComponent)
          in WrapperComponent
    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Failed prop type: Invalid prop `menuPosition` of type `string` supplied to `FloatingMenu`, expected `object`.
          in FloatingMenu (created by Tooltip)
          in Tooltip (created by WrapperComponent)
          in WrapperComponent

 PASS  src/components/OverflowMenu/OverflowMenu-test.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: React does not recognize the `closeMenu` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `closemenu` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
          in div
          in ul (created by OverflowMenu)
          in div (created by ClickListener)
          in ClickListener (created by OverflowMenu)
          in OverflowMenu (created by WrapperComponent)
          in WrapperComponent
```

This issue is to clean up our usage of these components so the tests don't warn 😄

PR #564, opened 2018-02-01
fix: bug fixes for #508

JMGordon:
Bug fixes for #508.

#### Changelog

**New**

- N/A

**Changed**

- Renamed child prop `closeMenu` to `onClick` in `OverflowMenu`.

- In `Tabs`, removed `onSelectionChange` from the props rest variable to eliminate it being passed into the `nav` element as an event handler. Reworked methods `handleTabClick`, `handleTabKeyDown`, and `handleTabAnchorFocus`, to receive `onSelectionChange` as an argument.

- Changed test `getTriggerPosition` in `Tooltip-test` so that an object is passed to `triggerPosition`, instead of a string.

**Removed**

- N/A

### DIFF
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -108,7 +108,7 @@ export default class Tabs extends React.Component {
       this.setState({
         selected: index,
       });
-      if (onSelectionChange !== undefined) {
+      if (typeof onSelectionChange === 'function') {
         onSelectionChange(index);
       }
     }
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -52,43 +52,49 @@ export default class Tabs extends React.Component {
   };
 
   // following functions (handle*) are Props on Tab.js, see Tab.js for parameters
-  handleTabClick = (index, label, evt) => {
-    evt.preventDefault();
-    this.selectTabAt(index);
-    this.setState({
-      dropdownHidden: !this.state.dropdownHidden,
-    });
-  };
-
-  handleTabKeyDown = (index, label, evt) => {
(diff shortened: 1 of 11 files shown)


src/components/Tabs/Tabs.js
```diff
@@ -97,13 +103,13 @@ export default class Tabs extends React.Component {
...
 
-  selectTabAt = index => {
+  selectTabAt = (index, onSelectionChange) => {
     if (this.state.selected !== index) {
       this.setState({
         selected: index,
       });
-      if (this.props.onSelectionChange) {
-        this.props.onSelectionChange(index);
+      if (onSelectionChange !== undefined) {
```
asudoh: Just a little thing - If we’d do more strict check than is-truthy, I’d recommend `typeof f === 'function'`.

src/components/Tooltip/Tooltip-test.js
```diff
@@ -162,19 +162,33 @@ describe('Tooltip', () => {
     });
   });
 
-  describe('getTriggerPosition', () => {
+  describe.skip('getTriggerPosition', () => {
```
asudoh: Is `.skip` here intentional?

asudoh:
Nice fix @JMGordon!

src/components/Tabs/Tabs.js
JMGordon: Good idea. I'll switch out the conditional with the recommended check.

src/components/Tooltip/Tooltip-test.js
JMGordon: Whoops... I forgot to remove the skip. Will remove on next commit.

PR #572, opened 2018-02-05
feat(SmallToggle): Add small toggle component

alisonjoseph:
https://github.ibm.com/Bluemix/carbon-issues/issues/453

#### Changelog

**New**

Small Toggle Component

### DIFF
--- a/src/components/ToggleSmall/ToggleSmall.js
+++ b/src/components/ToggleSmall/ToggleSmall.js
@@ -61,17 +61,18 @@ const ToggleSmall = ({
 
 ToggleSmall.propTypes = {
   className: PropTypes.string,
+  /**
+   * `true` to make it toggled by default.
+   */
   defaultToggled: PropTypes.bool,
   onToggle: PropTypes.func,
   id: PropTypes.string.isRequired,
   toggled: PropTypes.bool,
-  small: PropTypes.bool,
   ariaLabel: PropTypes.string.isRequired,
 };
 
 ToggleSmall.defaultProps = {
   defaultToggled: false,
-  small: false,
   onToggle: () => {},
 };
 
--- a/src/components/ToggleSmall/ToggleSmall.js
+++ b/src/components/ToggleSmall/ToggleSmall.js
@@ -60,14 +60,29 @@ const ToggleSmall = ({
 };
 
 ToggleSmall.propTypes = {
+  /**
+   * The CSS class for the toggle
+   */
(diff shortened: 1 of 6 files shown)


src/components/ToggleSmall/ToggleSmall.js
```diff
@@ -0,0 +1,78 @@
...
+            <path d="M2.2403 2.7299L4.9245 0 6 1.1117 2.2384 5 0 2.6863 1.0612 1.511z" />
+          </svg>
+        </span>
+      </label>
+    </div>
+  );
+};
+
+ToggleSmall.propTypes = {
+  className: PropTypes.string,
```
asudoh: Great to have comments explaining what those properties are for. If you do that, `Info` button at the top right will show those properties. The comment should be styled in:

```javascript
/**
 * `true` to make it toggled by default.
 */
defaultToggled: PropTypes.bool,
```

src/components/ToggleSmall/ToggleSmall.js
```diff
@@ -0,0 +1,78 @@
...
+  defaultToggled: PropTypes.bool,
+  onToggle: PropTypes.func,
+  id: PropTypes.string.isRequired,
+  toggled: PropTypes.bool,
+  small: PropTypes.bool,
```
asudoh: Did you mean to remove this prop (along with the `defaultProps`)?

alisonjoseph: Oops, I don't need the small prop. Is that what you are referring to, or something else?

src/components/ToggleSmall/ToggleSmall.js
alisonjoseph: Should we be adding a comment for every property moving forward, or just the properties we think need an explanation?

src/components/ToggleSmall/ToggleSmall.js
asudoh: Yes @alisonjoseph it’s what exactly I meant. Thank you for your quick fix!

src/components/ToggleSmall/ToggleSmall.js
asudoh: Good question - I tend to comment everything just for a consistency check, but I see comments for some self-explanatory ones may be useless. My suggestion is being careful with ones that *look* self-explanatory and see if they really are (e.g. if we have `id` where is it applied to?). Thank you for asking!

PR #575, opened 2018-02-06
fix: resolves #505 by adding aria-label to ul component

### LINKED ISSUE #505: A11Y: aira-lable could not be inserted to InteriorLeftNavList
## Detailed description
Followed with accessibility checkpoint, WAI-ARIA widgets must have an accessible name specified with aria-label, aria-labelledby or the widget's inner text. But, InteriorLeftNavList is failed in this point.

it is a general issue.

No

WAI-ARIA widgets have aria-label.

Chrome

"carbon-components-react": "5.0.6",

## Additional information

* Screenshots or code
image

## Add labels
aria-lable

JMGordon:
#### Changelog

**New**

- N/A

**Changed**

- Addressed #505 by adding aria-label attribute to ul element within the InteriorLeftNavList component.

**Removed**

- N/A

### DIFF
--- a/src/components/InteriorLeftNavList/InteriorLeftNavList.js
+++ b/src/components/InteriorLeftNavList/InteriorLeftNavList.js
@@ -105,7 +105,8 @@ export default class InteriorLeftNavList extends Component {
         <ul
           role="menu"
           className="left-nav-list left-nav-list--nested"
-          aria-label={title}>
+          aria-label={title}
+          aria-hidden>
           {newChildren}
         </ul>
       </li>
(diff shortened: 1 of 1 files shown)


src/components/InteriorLeftNavList/InteriorLeftNavList.js
```diff
@@ -105,7 +105,7 @@ export default class InteriorLeftNavList extends Component {
         <ul
           role="menu"
           className="left-nav-list left-nav-list--nested"
-          aria-hidden>
```
marijohannessen: I'm a bit concerned about removing this as it's making sure that the content is not being read by a screen reader when it's hidden (at least that's my understanding).

PR #577, opened 2018-02-06
fix: resolves #409 by honoring selected attribute when selectedIndex is not set

### LINKED ISSUE #409: [ContentSwitcher]: Doesn't respect Switch selected prop
## Detailed description

ContentSwitcher currently respects the `selectedIndex` prop. Although if this is not used, and instead the `selected` prop is used on the containing `Switch` components, the proper class is not applied.

Ideally, if ContentSwitcher is not given a selectedIndex, it's containing `Switch` components should render selected if given the prop `selected=true`.

I'm unsure how to treat edge cases of both `selectedIndex` being used as well as applying `selected` to a `Switch`.

I'm developing in Chrome, v^4.2.0 of this library.

1. Do not set `selectedIndex` on `ContentSwitcher`
2. Add `selected='true'` to one containing `Switch` component
3. Set `selected='false'` on all the other `Switch`es
4. View that the proper class of selected is not applied to `Switch`es with `selected='true'`

JMGordon:
#### Changelog

**Changed**

- Addressed #409 by honoring selected property on child if selectedIndex is not specified.

### DIFF
--- a/src/components/ContentSwitcher/ContentSwitcher.js
+++ b/src/components/ContentSwitcher/ContentSwitcher.js
@@ -10,10 +10,6 @@ export default class ContentSwitcher extends React.Component {
     selectedIndex: PropTypes.number,
   };
 
-  static defaultProps = {
-    selectedIndex: 0,
-  };
-
   state = {
     selectedIndex: this.props.selectedIndex,
   };
@@ -24,7 +20,9 @@ export default class ContentSwitcher extends React.Component {
         index,
         onClick: this.handleChildChange,
         onKeyDown: this.handleChildChange,
-        selected: index === this.state.selectedIndex,
+        ...(typeof this.state.selectedIndex === 'number'
+          ? { selected: index === this.state.selectedIndex }
+          : {}),
       })
     );
   }
--- a/src/components/ContentSwitcher/ContentSwitcher-test.js
+++ b/src/components/ContentSwitcher/ContentSwitcher-test.js
@@ -8,7 +8,7 @@ describe('ContentSwitcher', () => {
     const wrapper = shallow(
       <ContentSwitcher onChange={() => {}} className="extra-class">
         <Switch kind="anchor" text="one" />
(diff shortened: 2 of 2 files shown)


src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -10,10 +10,6 @@ export default class ContentSwitcher extends React.Component {
     selectedIndex: PropTypes.number,
   };
 
-  static defaultProps = {
-    selectedIndex: 0,
```
JMGordon: My concern over removing this is breaking someones code that relies upon this default behavior. Does anyone forsee this as an issue?

marijohannessen: Yeah, I'm a bit concerned about that as well. What's the motivation behind removing this?

JMGordon: Two reasons: 
(1) The default selection is surprising, as my expectation is that if `selectedIndex` were undefined, then no item would be selected (it could be that I don't fully understand the use case for this component).

(2) To honor the use case in #409 where child components are injected into `ContentSwitcher`, where one child component is selected, and `selectedIndex` is not set. However, after thinking about this part more, a user could work around this issue by noting which of their child components is selected, and then setting the corresponding value for `selectedIndex`.

asudoh: Sorry for being late on this, but here’s my thought - `selectedIndex` is not present should mean delegating the control of selection fully in `<ContentSwitcher>`. If it’s present that should be honored (that said, I’m a bit surprised that `<ContentSwitcher>` does not have `cWRP` to update the state upon change in props). Wrt #409 I’d assume the selection state of `<Switch>` should be updated upon parent `<ContentSwitcher>`’s selection, if I’m not missing anything.

JMGordon: Thanks @asudoh. Closing PR.

PR #583, opened 2018-02-07
chore(lint): lint JSDoc comments

asudoh:
#### Changelog

**New**

- ESLint rule for check for JSDoc comments.

### DIFF
--- a/src/tools/array.js
+++ b/src/tools/array.js
@@ -1,7 +1,7 @@
 /**
  * Shallow compare two arrays for equality
- * @param {array} arr1 the first array
- * @param {array} arr2 the second array
+ * @param {Array} arr1 the first array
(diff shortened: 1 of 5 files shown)


src/tools/array.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Shallow compare two arrays for equality
- * @param {array} arr1 the first array
- * @param {array} arr2 the second array
+ * @param {Aarray} arr1 the first array
```
joshblack: Oops, seems like a mis-type?

asudoh: Great catch - Thanks @joshblack! BTW wrt your another question, I have got one: https://github.com/carbon-design-system/carbon-components/pull/545

PR #587, opened 2018-02-08
Adding DataTable components to export list

ntsaini:
Not sure if there was a reason DataTable components were not in the export list. If there wasn't here is a pull request to add them.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,23 @@ export ComposedModal, {
 export ContentSwitcher from './components/ContentSwitcher';
 export CopyButton from './components/CopyButton';
 export DangerButton from './components/DangerButton';
+export DataTableRow from './components/DataTableRow';
+export DataTableData from './components/DataTableData';
+export {
+  DataTable,
+  DataTableToolbar,
+  DataTableToolbarContent,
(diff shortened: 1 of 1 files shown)


joshblack:
Hi @ntsaini! 👋  thanks so much for taking the time to make this PR 😄 

For DataTable, it isn't included in the default exports yet because we've only done a beta release of the component. When we do the GA release, we'll add it to the exports list 🙂

ntsaini:
@joshblack  thanks for the response, i thought that might be the case but submitted the pull request just in case it wasn't. 
Do you have a component status page like [this](http://carbondesignsystem.com/component-status) on the main carbon design system website? If not should this status information be indicated somehow in the storybook?

joshblack:
@ntsaini thankfully this specific component will be exposed in #562. In general, if a component isn't being exported then it's considered to be beta. Once we add it to the default exports, it's considered stable 😄  Will try and make sure that the storybook story descriptions reflect this accurately!

Closing this in favor of that PR, hope you don't mind!

ntsaini:
@joshblack that works, no problem at all.

PR #588, opened 2018-02-08
feat(IconV2): introduce pass-thru flavor of icon component

asudoh:
The new `<IconV2>` component allows icon data directly passed via props, and using it along with specific icon data from tree-shakable ESM module of `carbon-icons` beginning `6.3.0` release, app can optimize the bundle size.

Note: Our components using `<Icon>` (e.g. `<Accordion>`) should be updated to `<IconV2>` in future, but updating ones taking arbitrary icon name (e.g. `<Button>`) requires a breaking change.

#### Changelog

**New**

- `<IconV2>` component, as mentioned above.


joshblack:
Also, does this help with people who want to support custom icons at all?

asudoh:
@joshblack Thank you for the feedback! This PR does *not* reduce the size of `carbon-components-react.js` bundle, given it still has `<Icon>` - It rather is for reducing app’s bundle size by switching `<Icon>` to `<IconV2>`. For your latter comment, definitely yes!

joshblack:
@asudoh totally! Just meant an image similar to what is in: https://github.com/carbon-design-system/carbon-components-react/issues/537#issue-291725124 that shows how the consumer's bundle changes (i.e. does `carbon-icons` get smaller)

asudoh:
@joshblack I just took a look at what Rollup generates - I saw only a particular icon (instead of a whole) is imported. Not sure if WebPack is as good as Rollup on tree-shaking front, and thus didn’t use WebPack visualizer.

src/components/IconV2/IconV2-story.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import {
+  iconAdd,
+  iconAddGlyph,
+  iconAddOutline,
+} from 'carbon-icons/dist/carbon-icons-list';
```
joshblack: One thing I noticed, should we keep the build dir consistent between the libraries? e.g. here it's `dist`, react does `lib`, etc.

src/components/IconV2/IconV2-story.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import {
+  iconAdd,
+  iconAddGlyph,
```
joshblack: Would this be `iconAdd` and `glyphAdd` versus `iconGlyphAdd`? From a naming convention perspective, that is. Totally get that's what is exported from that package currently 😄

src/components/IconV2/IconV2-story.js
asudoh: @joshblack Interesting point - I guess the icons package chose `dist` given there is not much code in it. Also changing the distribution directory would be (minor) breaking change. But feel free to go forward with it if you feel strongly about it!

src/components/IconV2/IconV2-story.js
marijohannessen: @asudoh - going to merge this PR once this comment is addressed 👍

asudoh: Hi @joshblack trying to understand what you meant here - Are you suggesting the names of exports in `carbon-icons` (that’s auto-generated from icon IDs with `s.replace(/\-+([A-z])/g, (match, token) => token.toUpperCase())` logic)? Or something else?

joshblack: Yeah, I think this requires a change in the exports of `carbon-icons`, in this case. I'm not fully aware of all the icon permutations though so might be better to have a naming issue to discuss this kind of stuff just so everyone can weigh in.

asudoh: Created: https://github.com/carbon-design-system/carbon-icons/issues/59

joshblack:
Closing this due to inactivity. Feel free to re-open if we want to think about bringing this in soon!

PR #595, opened 2018-02-08
feat(Switch): add the icon prop to the Switch component to allow rendering images in ContentSwitcher

scottdickerson:
Closes carbon-design-system/carbon-components-react#585
Adds the icon prop to the Switch component.  Ideally styles should be added to the carbon-components css to handle positioning and coloring the svg icon based on the current selection state (https://github.com/carbon-design-system/carbon-components/issues/547), but these can be worked around by local css until this is incorportate.
#### Changelog

**New**

- Switch.icon property takes an element allowing you to send an svg in to be rendered next to the text


src/components/Switch/Switch.js
```diff
@@ -12,7 +12,9 @@ const Switch = props => {
     onKeyDown,
     selected,
     text,
+    icon,
     href,
+    icon,
```
asudoh: Looks a duplicate?

scottdickerson:
sorry @asudoh this was a merge error somewhere, refixed and resubmitted.

PR #597, opened 2018-02-09
fix(NumberInput): cope with user events being referred in async code

### LINKED ISSUE #566: NumberInput onChange produces unhelpful event
I am using NumberInput and am trying to use the onChange property to update the state when the user updates input. For most input based react components, event.target.value or event contain the value, but in this case I'm not sure how to derive the value from the event produced. 

I tried reproducing the issue in Sandbox, https://codesandbox.io/s/nwn6mozxnp. And you can see the issue when the function, handleCellCount, logs event.

In my project, we use 5.17.1, and i believe the same in the codesandbox.

The project is part of the CFaaS offering that we are developing.

the sandbox link above

asudoh:
Fixes #566.

#### Changelog

**New**

- `<NumberInput>` code to retain event after the event loop to cope with async event firing.


asudoh:
Thank you for pointing this out @tw15egan! Will address it next week.

giltohar:
Any Update, eta on integrating fix into released code?

asudoh:
Set the `<input>` as "imaginary target" to the evens being fired: https://github.com/carbon-design-system/carbon-components-react/pull/597/commits/4060efb326f9084cf52ace40eac28a705038cfbd. Hope this works for the use case of the original issue.

tw15egan:
Maybe I'm doing this wrong, but I am still unable to get the value of the `NumberInput` onchange or onclick, either returns the svg / path / button or null

screen shot 2018-02-13 at 10 13 11 am

asudoh:
@tw15egan Yes - I tried to keep it as natural as handling events on `<input>`, but please feel free to suggest for an alternative. Thanks!

tw15egan:
As long as we can ensure users know they need to access `evt.imaginaryTarget.value` to get the value of the input, I'm good with it! Thanks for tackling this , @asudoh 🙂

joshblack:
Coming in late to this, could we just update this block so that they can access the value? 

```
  handleChange = evt => {
    if (!this.props.disabled) {
      this.setState(
        {
          value: evt.target.value,
        },
        () => {
          this.props.onChange({ value: evt.target.value });
        }
      );
    }
  };
```

Could even do:

```js
this.props.onChange({ target: { value: evt.target.value }});
```

If we wanted consumers to think they were still accessing the event value.

asudoh:
@joshblack Good question - Just didn’t want to lose any other values in `event`, but please feel free to make a change if you feel that `value` is the only thing consumers would need. Thanks!

marijohannessen:
Any input here before merging @joshblack ?

joshblack:
It does feel like we should do the simpler thing here and not persist the event if we can get away with it. Are there strong use-cases for including other event fields for consumers that would require us persisting the event?

asudoh:
@joshblack `min`/`max`/[`validity`](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) are some of those properties consumers would be interested in. In a bigger thought, I attempted ensure that the fired event works like what `<input type=“number”>` fires on stepping up/down. You may know better on the best practice on this space, and please feel free to bring up a counterproposal. Thanks!

PR #607, opened 2018-02-13
fix(TextInput): aria-describedby casing

psachs21:
React warning throws if casing is off.

Pretty self explanatory. Without this fix, this warning shows:

```
Warning: Unknown ARIA attribute aria-describedBy. Did you mean aria-describedby?
```


joshblack:
Thanks so much @psachs21!

PR #627, opened 2018-02-16
feat(fileUploaded): Added accept attribute

loganmccaul:
The input type=file has an attribute, `accept`, that allows the user to set which file types the input accepts. It doesn't actually prevent upload of other files, but it does make the user's process of finding files that will be accepted easier.

Reference:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

### DIFF
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -18,6 +18,7 @@ export class FileUploaderButton extends Component {
     role: PropTypes.string,
     tabIndex: PropTypes.number,
     buttonKind: PropTypes.oneOf(['primary', 'secondary']),
+    accept: PropTypes.arrayOf(PropTypes.string),
   };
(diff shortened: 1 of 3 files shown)


src/components/FileUploader/FileUploader.js
```diff
@@ -18,6 +18,7 @@ export class FileUploaderButton extends Component {
     role: PropTypes.string,
     tabIndex: PropTypes.number,
     buttonKind: PropTypes.oneOf(['primary', 'secondary']),
+    accept: PropTypes.arrayOf(PropTypes.string),
```
joshblack: Are the potential string values bounded here? If so, might be helpful to use `PropTypes.oneOf` to make the prop type more strict, but understand if the list is wayyyy to large for that lol.

joshblack:
Thanks for doing this @loganmccaul!

Issue #632, opened 2018-02-19
[File Uploader] Need a way to clear inputs


loganmccaul:
## Detailed description

The File Uploader component currently keeps track of the file names that were uploaded in an internal state. The internal state can currently only be cleared by clicking on the remove icon tied to each uploaded file. However, we need the parent component to be able to clear the internal state.

We are using the File Uploader in a modal and we want to be able to clear the uploaded files when we hide the modal. Therefore the File Uploader component needs to expose a way to reset it's internal state.

> Is this issue related to a specific component?
File Uploader

marijohannessen:
Fixed in #633

PR #633, opened 2018-02-19
feat(FileUploader): Added method to clearFiles

loganmccaul:
Exports a method to clear the filenames state from the File Uploader component.

This method can be used by the parent using a ref.

Related to https://github.com/carbon-design-system/carbon-components-react/issues/632. Potentially a solution, but at minimum it's a workaround until a larger solution is decided on.


joshblack:
Hey @loganmccaul! How are you anticipating consumers to use the new method?

loganmccaul:
@joshblack Consumer's can use a ref to directly call the method from the parent component. It may not be the best solution to #632, but it works for our use case at the moment.

I can try to add some documentation to the storybook if that will be helpful.

I'm also open to alternative solutions.

joshblack:
Adding an example to the storybook and adding a test would be really helpful here 👍

src/components/FileUploader/FileUploader.js
```diff
@@ -125,7 +125,7 @@ export class Filename extends Component {
       return (
         <div
           className="bx--loading"
-          style={Object.assign(style, { width: '1rem', height: '1rem' })}
+          style={{ ...style, ...{ width: '1rem', height: '1rem' } }}
```
loganmccaul: Changed this to a spread operator because of some issue I was having in tests with the Object.assign running into read only properties.

Let me know if I should switch it back and work through the issues.

src/components/FileUploader/FileUploader-story.js
```diff
@@ -22,16 +23,28 @@ storiesOf('FileUploader', module)
...
     () => (
-      <FileUploader
-        labelTitle="Upload"
-        labelDescription="only .jpg and .png files at 500mb or less"
-        buttonLabel="Add files"
-        filenameStatus="edit"
-        accept={['.jpg', '.png']}
-        multiple
-      />
+      <div>
```
tw15egan: If we have to have a wrapping div, could we add a classname in case a user needs to target the container specifically? Maybe something like `bx--file__container`

loganmccaul: Absolutely! I wasn't sure if it mattered for the story.

src/components/FileUploader/FileUploader.js
joshblack: Do we need the second spread on the object literal with width/height? Seems like we'd just spread style and then add those attributes directly, unless I'm missing something.

loganmccaul: Nope, I'm just being dumb 😄

PR #639, opened 2018-02-21
feat(Tile): Add new tile group component and selectable tile (radio button)

alisonjoseph:
Closes carbon-design-system/carbon-components-react#586

Added a selectable tile that uses a radio button as the base so that you are only able to select one in a group. 

#### Changelog

**New**

`<TileGroup>` component
`<TileRadio>` component


alisonjoseph:
Does anyone have a better idea on naming for the `RadioTile` component, since `SelectableTile` is already taken by the checkbox select tile. 

Also, is it confusing to have them named the way I do in the storybook nav, where **multi-select** is `SelectableTile` and **Selectable** is `RadioTile`?

src/components/Tile/Tile-story.js
```diff
@@ -25,16 +28,52 @@ storiesOf('Tile', module)
     () => <ClickableTile>Clickable Tile</ClickableTile>
   )
   .addWithInfo(
-    'Selectable',
+    'Multi-select',
```
tw15egan: Could we have 3 tiles in this example like we have in `Selectable`?

marijohannessen:
I don't mind `RadioTile`, I can't think of anything better 😂  I noticed that in this PR it's `TileRadio` though, and I think I prefer `RadioTile`. Curious to see what the rest of the group thinks!

src/components/Tile/Tile-story.js
```diff
@@ -8,6 +9,8 @@ import {
   TileAboveTheFoldContent,
   TileBelowTheFoldContent,
 } from '../Tile';
+import TileGroup from '../TileGroup';
+import TileRadio from '../TileRadio';
```
tw15egan: To keep with the current convention, could we name this `RadioTile`? I agree, not the greatest naming convention, but will have to do. Can we keep track of this somewhere to make this change in v9 @marijohannessen ?

marijohannessen: We have a `v9` label that we can throw on issues, so once we decide on naming here we can create one

alisonjoseph:
I'm fine with `RadioTile` and agree we should make a note so we can fix the naming for v9.

marijohannessen:
Cool, let's go with `RadioTile`! I can open an issue to fix the naming 👍

PR #643, opened 2018-02-22
fix(pagination): remove | when pageInputDisabled is true

tw15egan:
Instead of placing a `|` with no padding, it will now render nothing. Added some conditional styles to prevent a double border / outline being hidden under other button.

screen shot 2018-02-22 at 10 23 06 am

screen shot 2018-02-22 at 10 23 00 am

Closes https://github.com/carbon-design-system/carbon-components-react/issues/637

### DIFF
--- a/src/components/PaginationV2/PaginationV2.js
+++ b/src/components/PaginationV2/PaginationV2.js
@@ -188,16 +188,17 @@ export default class PaginationV2 extends Component {
           <button
             className="bx--pagination__button bx--pagination__button--backward"
             onClick={this.decrementPage}
-            disabled={this.props.disabled || statePage === 1}>
+            disabled={this.props.disabled || statePage === 1}
+            style={
+              pageInputDisabled ? { borderRight: 0, marginRight: '1px' } : null
+            }>
             <Icon
               className="bx--pagination__button-icon"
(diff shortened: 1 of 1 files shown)


src/components/PaginationV2/PaginationV2.js
```diff
@@ -188,16 +188,17 @@ export default class PaginationV2 extends Component {
           <button
             className="bx--pagination__button bx--pagination__button--backward"
             onClick={this.decrementPage}
-            disabled={this.props.disabled || statePage === 1}>
+            disabled={this.props.disabled || statePage === 1}
+            style={
+              pageInputDisabled ? { borderRight: 0, marginRight: '1px' } : null
```
asudoh: Thank you for fixing this @tw15egan! Any way to add a custom style to avoid in-line?

tw15egan: Would you prefer we add a class here (`bx--pagination__button--no-index`)? And then have the style come from a change in `carbon-components`?

asudoh: @tw15egan Thank you for double-checking! Yes I think this is the right way to move forward.

PR #653, opened 2018-02-26
feat(Checkbox): add hideLabel prop to allow hiding the label but preserving accessibility

scottdickerson:
Closes carbon-design-system/carbon-components-react#649

adds hideLabel prop to Checkbox that acts similar to Select and TextInput, this allows us to hide the Label but still have it be there for screenreaders to pickup and read.  I had to implement slightly differently than TextInput (with an internal span) as the Checkbox Label element is what is used to actually display the checkbox.

#### Changelog

**Changed**

- (Checkbox) new hideLabel prop storybook and testcase
- (TextInput) added hideLabel storybook example


scottdickerson:
apologies for all the unnecessary commits listed here, will delete and recreate my fork once this is in to clean things up.

PR #660, opened 2018-02-28
feat(Dropdown): Add onOpen and onClose props

s100:
Closes carbon-design-system/carbon-components-react#655

#### Changelog

**New**

- To the `Dropdown` component, added two new properties `onOpen` which is called with no arguments when the dropdown is opened, and `onClose` which is called with no arguments when the dropdown is closed.
- Added these to `Dropdown-story.js` for demonstration purposes too.
- Tests.


s100:
@asudoh Thank you! How long is it likely to be before this can be merged and a new version of `carbon-components-react` released with these changes?

PR #661, opened 2018-02-28
docs(multiselect): add type PropType definition

tw15egan:
Adds in the `type` `PropType` for the `MultiSelect` component


joshblack:
Ooof, good catch! Should this be `PropTypes.oneOf(['default', 'inline'])`?

PR #662, opened 2018-02-28
feat(modal & button): adds the danger flavor for modals and buttons

davidmenendez:
Closes https://github.com/carbon-design-system/carbon-components-react/issues/664

Adds the danger primary button, tertiary button, and danger modal.

#### Changelog

**New**

* `Modal`
  * Add `danger` prop to display danger variant 
* `Modal-test` now includes danger variant-related tests
* `Button`
  * Extend `kind` prop to include `dangerPrimary` and `tertiary` variant
* `Button-test` now includes test for new `dangerPrimary` and `tertiary` variant


joshblack:
Hi there! 👋 Thanks for taking the time to contribute. It'd be helpful before we review to fill out the changelog to call out things that are additions to the library, things that are updates, or the removal of things in the library. This helps us look for patterns that could influence which semver range the merge commit might need to trigger, and is especially helpful when it comes to refactoring/extending existing components. Thanks!

davidmenendez:
updated

src/components/Button/Button.js
```diff
@@ -56,7 +58,14 @@ Button.propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
   small: PropTypes.bool,
-  kind: PropTypes.oneOf(['primary', 'secondary', 'danger', 'ghost']).isRequired,
+  kind: PropTypes.oneOf([
+    'primary',
+    'secondary',
+    'danger',
+    'ghost',
+    'dangerPrimary',
```
joshblack: Should we do `primary--danger`? cc @carbon-design-system/ui am unsure what naming convention we want to do here.

tw15egan: `danger--primary` (matches the class) or `primary--danger` works for me

davidmenendez:
alright

src/components/Button/Button.js
joshblack: How about `danger--primary` since it matches the class right now?

davidmenendez: updated

Issue #670, opened 2018-03-01
Support spying on the text input in a combobox


psachs21:
## [Support spying on the text input in a combobox]

I'd like to know when the user changes the filter input on a combobox so I can make an API request.

## Detailed description

> Describe in detail the issue you're having.
My combobox items are dependent on the filter input in a non-trivial way. In particular, I need to make an API request to get updated items. Current API doesn't support it.

> Is this a feature request (new component, new icon), a bug, or a general issue?
Feature request

> Is this issue related to a specific component?
Combobox

> What did you expect to happen? What happened instead? What would you like to see changed?
A handler for the input is exposed.

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
8.15

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
ICM Web (IBM Watson Finance)

PR #671, opened 2018-03-01
feat(ComboBox): Add support to spy on input filter value (#670)

psachs21:
Closes carbon-design-system/carbon-components-react#670

This allows the consumer to modify items in a more customized manner.
Like making an API call.

#### Changelog

**New**

* new prop on combobox: onFilterInputChange


psachs21:
@joshblack I went back and forth on naming that prop. I'm up for any recommendations. `onInputChange` is a little more concise and I suppose `onFilterInputChange` is unnecessary because we don't really refer to anything as filters.

PR #675, opened 2018-03-02
fix(dropdown): ensure items in dropdown cannot be tabbed to when menu…

tw15egan:
Closes carbon-design-system/carbon-components-react#617
Closes https://github.com/carbon-design-system/carbon-components-react/issues/609

#### Changelog

**New**

* Adds correct tabbing sequence when `Dropdown` is closed
* `aria-label` to appropriate regions

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -6,6 +6,7 @@ import Icon from '../Icon';
 
 export default class Dropdown extends PureComponent {
   static propTypes = {
+    ariaLabel: PropTypes.string.isRequired,
     children: PropTypes.node,
     className: PropTypes.string,
     defaultText: PropTypes.string,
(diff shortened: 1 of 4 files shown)


src/components/Dropdown/Dropdown.js
```diff
@@ -6,6 +6,7 @@ import Icon from '../Icon';
 
 export default class Dropdown extends PureComponent {
   static propTypes = {
+    ariaLabel: PropTypes.string,
```
joshblack: Should this be required?

tw15egan: I set a basic one by default, but you're right we should also make it required

s100:
@tw15egan This change has caused our application, which uses `Dropdown`, to begin outputting console errors:

image

This hasn't broken our application but it has created a substantial amount of work for us. For the future I would suggest not adding new required properties without a major version bump.

PR #680, opened 2018-03-06
[WIP] fix(multiselect): ensure there are not duplicate id's

tw15egan:
Closes carbon-design-system/carbon-components-react#615

Fixes a11y issues with `MultiSelect`

#### Changelog

**Changed**

* Added `-input` to the end of the checkbox `id` to ensure no duplicates.
* Added `evt.stopPropagation()` to inner `Checkbox` to ensure selection works properly


marijohannessen:
multiselect

The checkboxes are acting a bit weird when I select them - not sure if that's because of this PR or not?

marijohannessen:
Hey @joshblack , the weird focus issue happens when the `-input` part is applied to the checkbox id. When I move it back to what it was, it goes away but then the a11y problem still exists. Do you know what might be causing this?

joshblack:
@marijohannessen honestly not sure 😞 seems like the selection is causing the focus state to be triggered for the underlying node, despite the fact that it is moving to the top of the list. First guess would be some index-related bug with how downshift places items in its class instance.

shinytoyrobots:
Closing as unmerged.

PR #684, opened 2018-03-06
fix(overflow-menu): a11y fixes and keyboard interaction

### LINKED ISSUE #618: a11y audit: OverflowMenu a11y fixes
### Detailed description

- [x] An element with WAI-ARIA role menuitem is not contained in or owned by an element with one of the following WAI-ARIA roles: group, menu, menubar. See rule help for remediation guidance.
screen shot 2018-02-15 at 2 38 39 pm
screen shot 2018-02-15 at 2 41 03 pm

- [x] For the FloatingMenu variation: A WAI-ARIA widget must have an accessible name (see message help text for details)
screen shot 2018-02-15 at 2 43 41 pm

- [x] Cannot use keyboard interaction to navigate up and down the Overflow Menu using the up and down arrows

marijohannessen:
Closes #618 

a11y fixes for OverflowMenu. Added roles and aria-labels, and added keyboard interaction allowing the user to use arrow keys to navigate the menu.

#### Changelog

**Changed**

* `OverflowMenu.js`

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -203,15 +203,31 @@ export default class OverflowMenu extends Component {
   };
 
   handleKeyDown = evt => {
-    if (evt.which === 40) {
-      this.setState({ open: !this.state.open });
-      this.props.onClick(evt);
-    }
-  };
-
-  handleKeyPress = evt => {
     const key = evt.key || evt.which;
-
+    const isOverflowMenuItem = evt.target.classList.contains(
+      'bx--overflow-menu-options__btn'
+    );
+    const isOverflowMenuIcon = evt.target.classList.contains(
+      'bx--overflow-menu'
+    );
+    if ((key === 'ArrowDown' || key === 40) && isOverflowMenuIcon) {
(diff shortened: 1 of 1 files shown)


asudoh:
@marijohannessen Good try to add keyboard nav support to overflow menu! Seems that the keyboard nav behavior between floating vs. non-floating mode is different, and we need to put it in consideration that the menu body with floating mode has distant DOM location from the trigger button. Vanilla floating menu code attempts to [focus on the first focusable element when it gets open](https://github.com/carbon-design-system/carbon-components/blob/27765a4/src/components/floating-menu/floating-menu.js#L227), and [puts the focus back to the trigger button when it loses focus](https://github.com/carbon-design-system/carbon-components/blob/27765a4/src/components/floating-menu/floating-menu.js#L120).

Also great to have more flexibility art DOM structure, esp. parent-child relationship. Some [logic in vanilla dropdown](https://github.com/carbon-design-system/carbon-components/blob/5888289/src/components/dropdown/dropdown.js#L112-L128) may help.

marijohannessen:
Closing this as it needs to be a bigger change reflected in both vanilla and the React repo.

PR #685, opened 2018-03-06
feat: open prop will override any event handler

nelsonchen90:
To address this: https://github.com/carbon-design-system/carbon-components-react/issues/669#issuecomment-370884282 
Closes carbon-design-system/carbon-components-react#669

Tooltip will be fully controlled by `open` if it's defined.

#### Changelog

**Changed**

* if `props.open` is defined on the Tooltip, then all event handler will be skipped /overridden. Changing the `open` from prop is the only way to toggle the display of the tooltip

* Tooltip `state.open` will be initialized to `false` if `open` on props is undefined.

**Removed**

* removed the default prop value from the Tooltip component.

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -141,7 +141,7 @@ export default class Tooltip extends Component {
   };
 
   state = {
-    open: this.props.open === undefined ? false : this.props.open,
+    open: Boolean(this.props.open),
   };
 
   componentDidMount() {
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -132,7 +132,6 @@ export default class Tooltip extends Component {
   };
 
   static defaultProps = {
-    open: false,
     direction: DIRECTION_BOTTOM,
     showIcon: true,
     iconName: 'info--glyph',
@@ -142,7 +141,7 @@ export default class Tooltip extends Component {
   };
 
   state = {
-    open: this.props.open,
+    open: this.props.open === undefined ? false : this.props.open,
   };
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -142,7 +141,7 @@ export default class Tooltip extends Component {
   };
 
   state = {
-    open: this.props.open,
+    open: this.props.open === undefined ? false : this.props.open,
```
asudoh: Would we anticipate non-boolean and non-null/undefined values for `state.open`? If not, wondering if we can do `Boolean(this.props.open)`.

nelsonchen90:
@asudoh Thanks for reviewing.

I am trying to implement a 'popover' type of tooltip which user can hover on the tootlip content. In order to do that, I need a wrapper around the tooltip and that wrapper is going to control the tooltip thru the `open` prop.

if we don't ignore the other events when the `open` is defined (true/false), those events such as `mouseout` will not honor the value of `open` and close the tooltip instead of keeping it open.

Let me know if this use case make sense. Thanks again!

src/components/Tooltip/Tooltip.js
nelsonchen90: I will make the code change. thanks for pointing our

nelsonchen90:
@asudoh Yes, I think that's exactly what we wanted. Maybe we can just close this PR then. I think your implementation makes things a lot easier!

asudoh:
Thanks @nelsonchen90! Will port https://github.com/carbon-design-system/carbon-components/pull/639 to this repo.

tw15egan:
Closing PR in favor of https://github.com/carbon-design-system/carbon-components-react/pull/696

PR #694, opened 2018-03-07
fix(tabs): fix a11y issues

tw15egan:
Closes carbon-design-system/carbon-components-react#623

Adds `aria-label`

#### Changelog

**New**

* Adds `aria-label` to inner `Tab` dropdown (when viewing `Tab` on smaller viewports)


joshblack:
Is this something we'd expect the screen reader to announce since it's an aria label? If so, might help out to make it configurable and human-readable like "Select a tab".

joshblack:
Any thoughts on this PR @tw15egan ?

tw15egan:
@joshblack updated to make `aria-label` configurable

PR #696, opened 2018-03-08
feat(Tooltip): handle hover-over on tooltip

### LINKED ISSUE #669: Tooltip: control the open/closed state thru props after mounted
## Detailed description

I need to control the display of the tooltip programmatically. We have requirement that we want the mouse cursor to be able to hover on the tooltip. In order to do that, I need to control the open/close state of the tooltip thru the prop `open`

This is a feature request. (I will create PR)

Yes. Tooltip

Expectation is that the we can control the open/close state thru the prop `open`. But it turns out it's only there to set the initial state of the tooltip. I want to make it dynamic by updating the `state.open` in `componentWillReceiveProps`.

Chrome, Firefox, IE 10

5.20.0

Information server / Unified governance and integration. 

1. Set the initial value of `open` to false.
2. once it's rendered, set the `open` to true.
3. you won't see the tooltip as the state is not updated by the update of the prop

## Additional information

* Screenshots or code
* Notes

asudoh:
Closes #669.
Refs carbon-design-system/carbon-components#266.

This PR is for preventing tooltip from being closed when the mouse is on the open tooltip.

#### Changelog

**New**

* Code to detect if the mouse is is on the open tooltip
* Code to debounce event handlers that open/close tooltip (Moving mouse cursor from the trigger button to the open tooltip causes mouse pointer to go to another element between them, and thus detecting mouse position right after the mouse cursor goes out of the trigger button will detect another element)

### DIFF
--- a/src/components/Tooltip/Tooltip-test.js
+++ b/src/components/Tooltip/Tooltip-test.js
@@ -1,8 +1,13 @@
 import React from 'react';
+import debounce from 'lodash.debounce';
 import Icon from '../Icon';
 import FloatingMenu from '../../internal/FloatingMenu';
 import Tooltip from '../Tooltip';
 import { mount } from 'enzyme';
+
+jest.mock('lodash.debounce');
+
+debounce.mockImplementation(fn => fn);
 
 describe('Tooltip', () => {
   describe('Renders as expected with defaults', () => {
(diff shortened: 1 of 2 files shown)


nelsonchen90:
@asudoh 
@tw15egan 
is it possible we get this in today? I have dependency needs to finish before end of this week https://github.com/carbon-design-system/carbon-components-react/issues/669

Thanks!

asudoh:
Thanks @tw15egan for trying out the fix! Added some code to detect the context menu.

nelsonchen90:
@tw15egan is there any ETA on the merge/publish?

Thanks!

PR #701, opened 2018-03-08
fix(FileUploader): Reset the internal value so onChange works as expected

loganmccaul:
**BUG**
The FileUploader component currently doesn't allow you to delete a file and then reupload the exact same. The value of `<input type=file ...` isn't reset.

Steps to reproduce:
1. Navigate to the FileUploader: http://react.carbondesignsystem.com/?selectedKind=FileUploader&selectedStory=FileUploader&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

2. Upload a file

3. Click the X or the Clear Files button to clear the file

4. Upload the same file again

5. The file doesn't upload a second time

**FIX**
As advised by:
https://stackoverflow.com/questions/39484895/how-to-allow-input-type-file-to-select-the-same-file-in-react-component

Add an onClick listener to the input file element that clears the value of the element.

### DIFF
--- a/src/components/FileUploader/FileUploader-test.js
+++ b/src/components/FileUploader/FileUploader-test.js
@@ -55,6 +55,15 @@ describe('FileUploaderButton', () => {
     it('does not have default role', () => {
       expect(mountWrapper.props().role).not.toBeTruthy();
     });
+
+    it('resets the input value onClick', () => {
+      const input = mountWrapper.find('.bx--visually-hidden');
+      input.instance().value = '';
+      const evt = { target: { value: input.instance().value } };
+      input.simulate('click', evt);
+
+      expect(evt.target.value).toEqual(null);
+    });
   });
 
   describe('Unique id props', () => {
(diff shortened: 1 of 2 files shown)


src/components/FileUploader/FileUploader-test.js
```diff
@@ -55,6 +55,15 @@ describe('FileUploaderButton', () => {
     it('does not have default role', () => {
       expect(mountWrapper.props().role).not.toBeTruthy();
     });
+
+    it('resets the input value onClick', () => {
+      const input = mountWrapper.find('.bx--visually-hidden');
+      input.instance().value = '';
+      const evt = { target: { value: input.instance().value } };
+      input.simulate('click', evt);
+
```
loganmccaul: This feels so funky, but since the onClick function just clears the elements value I couldn't figure out a good way to pass the element's value in and get it out without wrapping it in the mock evt object.

PR #709, opened 2018-03-13
feat(overflowmenu,icon): Support custom icons in icon and overflow menu.

psachs21:
As we expand carbon-components adoption, we need to make some icons rendering more flexible because carbon-icons may not contain the icon required by design.

#### Changelog

**New**

* renderIcon prop on OverflowMenu: An optional render prop that allows overriding the icon being rendered by overflow menu.
* children prop on icon: An optional render prop on icon to override the rendered icon.


asudoh:
Thanks @psachs21 for trying to address this problem space! Trying to understand one thing; Is your specific issue similar to, or same as, one seen in https://github.com/carbon-design-system/carbon-components-react/pull/588, where icon data instead of name can be passed in?

psachs21:
@asudoh I didn't see that PR but looking over it, I'm not familiar with how svgData is being build or how it's expected. It may provide a way if i can provide an arbitrary svg file but I'm not sure what format that is. The carbon-icons seems to build a json file to represent the SVG files. It seems unnecessary when we can use something like `svg-sprite-loader` to automatically build our sprites. However, what i get out of svg-sprite-loader is just a viewBox and an href, which makes using this loader with Icon problematic, IconV2 as well.

asudoh:
@psachs21 Thank you for elaborating your need - Looking at https://www.npmjs.com/package/svg-sprite-loader#runtime-configuration - Wanted to see if the markup (with `<svg>` and `<use>`) is what you want to have your app render. If it's the case, wondering why you wanted to use `carbon-components-react` there. Thanks!

psachs21:
@asudoh This is a valid point. I could just make my own component to do the same. I suppose it doesn't serve much a point here. I can remove the change to Icon but what of the change to OverflowMenu? I'd like to use that with a different icon set.

asudoh:
Definitely a good question @psachs21! I’d do:

```javascript
icon ? React.cloneElement(icon, {
  onClick: this.handleClick,
  onKeyDown: this.handleKeyDown,
}) : (
  <Icon
    onClick={this.handleClick}
    onKeyDown={this.handleKeyDown}
    className={overflowMenuIconClasses}
    name={iconName}
    description={iconDescription}
    style={{ width: '100%' }}
  />
)
```

Where `icon` is `PropTypes.node`. Thanks!

psachs21:
@asudoh I feel the render callback is the more modern way to do this, since it allows the user to override the onclick handler (or other props) if they so desire. It's also the pattern being used in many new carbon components (DataTable).

src/components/OverflowMenu/OverflowMenu-story.js
```diff
@@ -56,4 +56,28 @@ storiesOf('OverflowMenu', module)
         />
       </OverflowMenu>
     )
+  )
+  .addWithInfo(
+    'custom icons',
```
marijohannessen: Can we change this to say `custom trigger` since it's showing an example without an icon?

src/components/OverflowMenu/OverflowMenu-story.js
```diff
@@ -56,4 +56,28 @@ storiesOf('OverflowMenu', module)
...
+    `
+      Sometimes you just want to render something other than an icon
+    `,
+    () => (
+      <OverflowMenu
+        {...overflowMenuEvents}
+        style={{ width: 'auto' }}
+        floatingMenu
+        renderIcon={iconProps => <div {...iconProps}>Custom stuff</div>}>
```
marijohannessen: Can we change this to say `Custom trigger` instead of `Custom stuff`?

psachs21:
@marijohannessen updated according to recommendation.

marijohannessen:
Thanks for fixing that, merging now!

PR #713, opened 2018-03-15
chore(release): let semantic-release release manually

asudoh:
With this change, `semantic-release` stops triggering a release unless there is a commit whose subject starts with `release:` or `release(subheader)`.

#### Changelog

**Changed**

`semantic-release` stops triggering a release unless there is a commit whose subject starts with `release:` or `release(subheader)`.


joshblack:
Hey @asudoh! Do you have a reference issue for this to help catch up on reasoning/context for this change?

asudoh:
Hi @joshblack this is for address the problem we had with https://github.com/carbon-design-system/carbon-components/issues/620 - Meaning, this is so completing code review and merging a PR does not accidentally break our services. Feel free to speak up if you need more info. Thanks!

joshblack:
If we're changing our release strategy, I think we should do a little research and also come up with a game plan for types of releases, when releases are dropped, etc. before merging this in. What do you think?

asudoh:
@joshblack This PR does not radically change our release strategy, just turning "full-auto mode” to “semi-auto” mode. e.g. Definitely not changing anything about _types of releases_ as you brought up.

When releases are dropped is a good question - The thought here is doing it whenever we’d want to do, just dropping a note to our key stakeholders and given it a day or two beforehand. If you’d want to take this as an opportunity for coming up with more sophisticated release plans it’s great and let’s discuss further. Thanks!

joshblack:
@asudoh I think I'd feel more comfortable merging this in if we had this process defined, documented, and agreed upon by the team + consumers beforehand. Hope that makes sense. If others think this approach is valuable right now without that work, definitely will understand 😄

joshblack:
Closing in favor of discussion in https://ibm.biz/BdZnzU feel free to re-open if we want to introduce this sooner!

Issue #705, opened 2018-03-12
DataTable should preserve sort when re-rendered and new data should be re-sorted


AndySeymour2904:
## Detailed description

> Describe in detail the issue you're having.
When I update data in a DataTable the sort is reset

> Is this a feature request (new component, new icon), a bug, or a general issue?
Feature Request / bug

> Is this issue related to a specific component?
DataTable

> What did you expect to happen? What happened instead? What would you like to see changed?
Expect sort to be preserved and new data to be sorted according to the current sortKey and sortDirection

Happy to provide a PR for this if agreed

joshblack:
Hi @AndySeymour2904! 👋 Thanks so much for taking the time to make this issue 😄 

Definitely agreed that `DataTable` should preserve sort order when re-rendered. Unfortunately we don't have the bandwidth to fix this right now, but will gladly accept any/all PRs to provide this behavior!

joshblack:
Closed by #728 and in subsequent bug fixes, I think 😄

PR #728, opened 2018-03-20
feat(DataTable): Preserve sorting in DataTable when it is re-rendered

AndySeymour2904:
Closes carbon-design-system/carbon-components-react #705 

Changes to make DataTable preserve the sort when it is re-rendered. This includes re-sorting the rows when new props are received

#### Changelog

**New**

* getCurrentSortState - Returns the sorted state based on the current sort options in the state

**Changed**

* DataTable.componentWillReceiveProps - Now triggers a resort of new data based on sort options in state, this is to preserve sort


joshblack:
Just seems like the snapshots need to be updated and then we're good to go! 🎉

PR #747, opened 2018-03-26
fix ComposedModal does not close when open, resolves #738

### LINKED ISSUE #738: ComposedModal does not close when open
@bellsworth13 reported an issue in CloudFoundry where the ComposedModal doesn't close when expected. This issue is to:

- Properly identify the root underlying issue causing the modal not to close
- Fix the issue

rezak-otmani:
Closes #738 

add constructor,

improve `componentWillReceiveProps`

and add the state for the render

#### Changelog

**New**

* `ComposedModal` can close when expected

**Changed**
**Removed**


src/components/PaginationV2/PaginationV2.js
```diff
@@ -186,7 +185,7 @@ export default class PaginationV2 extends Component {
                 )}
           </span>
         </div>
-        <div className="bx--pagination__right bx--pagination--inline">
+        <div className="bx--pagination__right">
```
joshblack: Is the `bx--pagination--inline` class not needed anymore in this component?

joshblack:
cc @carbon-design-system/ui for those who have context for this component. I'm unsure about background for `ComposedModal` so wanted to check before reviewing.

src/components/PaginationV2/PaginationV2.js
rezak-otmani: it's just that there is a slight shift of the number in using it, it's not centred, but it can be used

rezak-otmani: yes, I can change this

asudoh:
The history of the commits tells that @tw15egan may have more insights on this, but here's my finding:

* The open/close UI state is controlled by [`is-visible` CSS class](https://github.com/rezak-otmani/carbon-components-react/blob/3cb99f2393a8c5f7df36413a2036c73dd4e9cd27/src/components/ComposedModal/ComposedModal.js#L64)
* The are two sources for open/close state - [`prop.open`](https://github.com/rezak-otmani/carbon-components-react/blob/3cb99f2393a8c5f7df36413a2036c73dd4e9cd27/src/components/ComposedModal/ComposedModal.js#L56) in addition to [`state.open`](https://github.com/rezak-otmani/carbon-components-react/blob/3cb99f2393a8c5f7df36413a2036c73dd4e9cd27/src/components/ComposedModal/ComposedModal.js#L19). Would require `cWRP` to keep them in sync.

@rezak-otmani Do you want to work on changing this PR to put these into cosideration? Thanks!

rezak-otmani:
Hi @asudoh  thanks for your finding
yes I will work to change this PR

src/components/ComposedModal/ComposedModal.js
```diff
@@ -15,9 +15,12 @@ export default class ComposedModal extends Component {
     onKeyDown: PropTypes.func,
   };
 
-  state = {
-    open: this.props.open,
-  };
+  constructor(props) {
```
joshblack: Could we keep this block the same?

rezak-otmani: yes I will hand over the block

PR #748, opened 2018-03-26
feat(DataTable): Row cell has header key

ajdaniel:
DataTable, the `rows` argument in the renderProp has a new prop in the `cells` array. The `cell` object has a new `info` property with an object value of `{ header : string }` which can indicate the key of the header. Useful to determine the current column for side-effects.

Closes carbon-design-system/carbon-components-react#737

#### Changelog

**New**

* `row.cells[]` objects have a new `info.header` property to indicate the header key of the current cell.

Signed-off-by: Andrew J Daniel <andrewjd1@gmail.com>

### DIFF
--- a/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
+++ b/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
@@ -262,6 +262,9 @@ exports[`DataTable selection should render 1`] = `
                   Object {
                     "errors": null,
                     "id": "b:fieldA",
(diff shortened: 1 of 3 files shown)


joshblack:
One book-keeping note, could you remove the `closes` keyword? I think there are some parts of that issue that we'll also want to add in as well that are worth keeping around 😄 

One option is to update the issue to use a checklist with the different additions and include this change as one of the checks.

joshblack:
Sounds good @ajdaniel 👍

PR #749, opened 2018-03-27
fix(Tile): ensure prop-state sync

### LINKED ISSUE #677: SelectableTile not making use of "selected" prop
## Detailed description

I am passing the `selected` prop into `SelectableTile`. This only appears to set the default state for the component.

Subsequent changes in the `props.selected` don't result in a change in the tile's `state.selected`. In fact, the `selected` state appears to change independently of the `selected` prop.

Can someone verify if this is intentional? Thanks.

asudoh:
Fixes #677.

#### Changelog

**New**

* Prop-state sync code for `ClickableTile`, `SelectableTile` and `ExpandableTile`.


src/components/Tile/Tile.js
```diff
@@ -222,6 +234,18 @@ export class ExpandableTile extends Component {
     handleClick: () => {},
   };
 
+  componentWillReceiveProps({ expanded, tileMaxHeight, tilePadding }) {
```
joshblack: Should we batch these all in one `setState` call? e.g.

```js
const nextState = {};

if (expanded !== this.props.expanded) {
  nextState.expanded = expanded;
}

// ...

this.setState(nextState);
```

asudoh: OK was wondering, and hoped that it happens automagically. Definitely can do!

joshblack: I know that React batches setState calls in `cDM`, wasn't sure about `cWRP` though.

asudoh: Thought of seeing if batch happens with `cWRP`. It appears that it's true for all React lifecycle methods. Quickly created a test case and I saw render call only happens once after `cWRP` does multiple `.setState()` calls.

joshblack: Was basing off of:

> React ensures that any setState calls that happen during componentDidMount and componentDidUpdate are flushed before the user sees the updated UI.

[Reference](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)

asudoh: I see - BTW this is from my Google search - Which tells batching happens for all React lifecycle methods, more precisely, whenever the `setState()` call is in React call stack:

> It appears that it's true for all React lifecycle methods

PR #751, opened 2018-03-27
fix page size number overflow when it's 3 digits #732

### LINKED ISSUE #732: Paginator: page size number overflow when it's 3 digits
## Detailed description

screen shot 2018-03-20 at 1 51 02 pm

bug

PaginatorV2

icon should flow to right

Chrome

5.29.4

IBM Cloud

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

rezak-otmani:
Closes #732

#### Changelog

**New**

no overflow when it's 3 digits

**Changed**
**Removed**

* `inline` pros of select components

### DIFF
--- a/src/components/PaginationV2/PaginationV2.js
+++ b/src/components/PaginationV2/PaginationV2.js
@@ -165,7 +165,6 @@ export default class PaginationV2 extends Component {
             id={`bx-pagination-select-${inputId}`}
             labelText={itemsPerPageText}
             hideLabel
-            inline
             onChange={this.handleSizeChange}
             value={statePageSize}>
             {pageSizes.map(size => (
@@ -207,7 +206,6 @@ export default class PaginationV2 extends Component {
               id={`bx-pagination-select-${inputId + 2}`}
               labelText={itemsPerPageText}
               hideLabel
-              inline
               onChange={this.handlePageInputChange}
               value={statePage}>
               {selectItems}


joshblack:
Hi @rezak-otmani! 👋 Thanks so much for taking the time to make this PR 😄 

One thing that would be incredibly helpful is to fill out the PR template when creating each PR. This helps us know what to look for in the changes, including what could potentially be a major change.

In addition, would help to include a brief description of the changes right below the `Closes #` line that discusses the approach taken in the PR and how it fixes the corresponding issue.

Thanks again!

asudoh:
Need devs who have more insights in this area - @carbon-design-system/ui Can we apply non-inline `<Select>` for pagination? Do we think applying `width:auto` there will be a safer fix? Thanks!

rezak-otmani:
Hi @asudoh you're right, I think that the origin of the issue comes from this line  https://github.com/carbon-design-system/carbon-components/blob/master/src/components/pagination/_pagination.scss#L108  in  *carbon-components*
 the size must be increased to support the size of the number more than 2 digits

asudoh:
Thanks @rezak-otmani for sharing your thought! Closing this given the fix should be applied to another repo.

@carbon-design-system/ui Do we really need https://github.com/carbon-design-system/carbon-components/blob/master/src/components/pagination/_pagination.scss#L113 line? Thanks!

PR #754, opened 2018-03-27
fix(DataTable): update state properly in componentWillReceiveProps

### LINKED ISSUE #740: Data Table does not render row data entered through this.props
The datatable component will not render the data when it is passed through props, but is rendering when I pass the data directly from a json from another file. Please note the data passed directly through with a  json file is passed through at the very start upon rendering of the whole page, while the data passed through props is through user generated input, so happens during app usage. 

Also not no error pops up in console. 

After digging through source code, we think it may have something to do with this code block below:

https://github.com/carbon-design-system/carbon-components-react/blob/38b8f56daaed101a6cef69662e6d7b941220013a/src/components/DataTable/DataTable.js#L109-L126

It looks like the table will only render newly imported rows when you are sorting the header keys or sorting the direction. However in our case, we are injecting new rows into the rows prop of <DataTable>, see code below:

          <DataTable
            rows={this.props.investors} //why wont this render???
            headers={TableHeaders}
            filterRows={(rows) => {
              console.log('rows adam', rows)
              return rows.map((row) => row.id)
            }}
            render={({
              rows,
              headers,
              getHeaderProps,
              getSelectionProps,
              getBatchActionProps,
              onInputChange,
              // selectedRows,
            }) => {
              console.log('inside2', rows, headers)

              return (
                <TableContainer title='DataTable with batch actions'>
                  <TableToolbar>
                    {/* make sure to apply getBatchActionProps so that the bar renders */}
                    <TableBatchActions {...getBatchActionProps()}>
                      {/* inside of you batch actinos, you can include selectedRows */}
                      <TableBatchAction >
                        Remove Investor
                      </TableBatchAction>
                      <TableBatchAction >
                        Modify Restriction Dates
                      </TableBatchAction>
                    </TableBatchActions>
                    <TableToolbarSearch onChange={onInputChange} />
                    <TableToolbarContent>
                      <TableToolbarAction
                        iconName='download'
                        iconDescription='Download'
                      // onClick={action('TableToolbarAction - Download')}
                      />
                      <TableToolbarAction
                        iconName='edit'
                        iconDescription='Edit'
                      // onClick={action('TableToolbarAction - Edit')}
                      />
                      <TableToolbarAction
                        iconName='settings'
                        iconDescription='Settings'
                      // onClick={action('TableToolbarAction - Settings')}
                      />
                      <ModalWrapper
                        // modalProps={{ onBlur: { onBlur() }, onClick: { onClick() }, onFocus: { onFocus() }, …}}
                        id='transactional-modal'
                        buttonTriggerText='Add New'
                        // modalLabel='Add New Investor'
                        modalHeading='Add New Investor'
                        onSubmit={this.handleInvestorSubmit}
                        shouldCloseAfterSubmit
                      >
                        <p className='bx--modal-content__text'>
                          Please enter the information below to add a single investor.
                        </p>
                        <br />
                        <InvestorForm onSubmit={this.handleInvestorSubmit} />
                      </ModalWrapper>
                    </TableToolbarContent>
                  </TableToolbar>
                  <Table>
                    <TableHead>
                      <TableRow>
                        <TableSelectAll {...getSelectionProps()} />
                        {headers.map((header) => (
                          <TableHeader {...getHeaderProps({ header })}>
                            {header.header}
                          </TableHeader>
                        ))}
                      </TableRow>
                    </TableHead>
                    <TableBody>
                      {rows.map((row) => (
                        <TableRow key={row.id}>
                          <TableSelectRow {...getSelectionProps({ row })} />
                          {row.cells.map((cell) => (
                            <TableCell key={cell.id}>{cell.value}</TableCell>
                          ))}
                        </TableRow>
                      ))}
                    </TableBody>
                  </Table>
                </TableContainer>
              )
            }}
          />

So we think it may be that an injection of new items does not trigger updating of the internal state of the component.

Please also note I am using the pagination component to get the length of the this.props.investors, and it is correctly reading it and rendering it, so I am definitely not passing through an empty rows prop.
It really seems like something is happening through the internal sorting/filtering of the rows prop, because this.props.investor has real data and then when it renders inside datatable it is just picking up a blank array. 

Any help or direction is greatly appreciated! 

appears to be a bug

chrome

5.31.1

joshblack:
Closes #740 

Adds an extra check in the `cWRP` hook that will check to see if there are additional rowIds in the `nextState` and will update state accordingly, if so.

#### Changelog

**New**

**Changed**

* `DataTable`
  * Add behavior to `cWRP` to validate rowIds
  * Add corresponding tests for bugfix
  * Add `with dynamic rows` story to `DataTable-story`

**Removed**

datatable

### DIFF
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -108,16 +108,16 @@ export default class DataTable extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const rowIds = this.props.rows.map(row => row.id).sort();
-    const nextRowIds = nextProps.rows.map(row => row.id).sort();
+    const rowIds = this.props.rows.map(row => row.id);
+    const nextRowIds = nextProps.rows.map(row => row.id);
 
     if (!isEqual(rowIds, nextRowIds)) {
       this.setState(state => getDerivedStateFromProps(nextProps, state));
       return;
     }
 
-    const headers = this.props.headers.map(header => header.key).sort();
-    const nextHeaders = nextProps.headers.map(header => header.key).sort();
+    const headers = this.props.headers.map(header => header.key);
+    const nextHeaders = nextProps.headers.map(header => header.key);
 
     if (!isEqual(headers, nextHeaders)) {
       this.setState(state => getDerivedStateFromProps(nextProps, state));
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import isEqual from 'lodash.isequal';
 import getDerivedStateFromProps from './state/getDerivedStateFromProps';
 import { getNextSortState, getCurrentSortState } from './state/sorting';
 import denormalize from './tools/denormalize';
@@ -120,6 +121,22 @@ export default class DataTable extends React.Component {
           key: this.state.sortHeaderKey,
         }),
       });
+      return;
+    }
+
+    const rowIds = this.state.rowIds.sort();
+    const nextRowIds = nextState.rowIds.sort();
+
+    if (!isEqual(rowIds, nextRowIds)) {
+      this.setState(nextState);
+      return;
+    }
+
+    const headers = this.props.headers.map(header => header.key);
+    const nextHeaders = nextProps.headers.map(header => header.key);
+
+    if (!isEqual(headers, nextHeaders)) {
+      this.setState(nextState);
     }
   }
 
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import isEqual from 'lodash.isequal';
 import getDerivedStateFromProps from './state/getDerivedStateFromProps';
-import { getNextSortState, getCurrentSortState } from './state/sorting';
+import { getNextSortState } from './state/sorting';
 import denormalize from './tools/denormalize';
 import { composeEventHandlers } from './tools/events';
 import { defaultFilterRows } from './tools/filter';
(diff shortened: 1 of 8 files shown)


src/components/DataTable/DataTable.js
```diff
@@ -120,6 +121,22 @@ export default class DataTable extends React.Component {
           key: this.state.sortHeaderKey,
         }),
       });
+      return;
+    }
+
+    const rowIds = this.state.rowIds.sort();
```
ajdaniel: I don't think this does exactly what you think it does. This will sort the IDs of the rows, not the data itself. It passes our tests because in the tests, we are using IDs of 'A' 'B' and 'C' and expecting that order, it's a coincidence. I think this passes tests for now but may lead to odd behaviour

joshblack: @ajdaniel just was using `sort` as a deterministic hash in this case so the equality check would work.

ajdaniel: Ah. I understand now!

joshblack: Like you commented in the other thread though, this simple heuristic probably doesn't capture all the use-cases 😞

joshblack:
cc @ajdaniel if you want to re-review 🙂

src/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -303,4 +308,201 @@ describe('DataTable', () => {
...
+          id: 'd',
+          fieldA: 'Field 4:A',
+          fieldB: 'Field 4:B',
+        },
+      ];
+
+      wrapper.setProps({ rows: nextRows });
+
+      const nextArgs = mockProps.render.mock.calls[1][0];
+      expect(nextArgs.rows.length).toBe(nextRows.length);
```
asudoh: @joshblack Are you updating test for row ID order here...? Thanks!

joshblack: Just added!

Issue #736, opened 2018-03-21
The Pagination with known total number of items control is too strict when validating the page number input


zgrossbart:
1. Go to the pagination with known total number of items here:

http://react.carbondesignsystem.com/?selectedKind=Pagination&selectedStory=with%20known%20total%20number%20of%20items&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

2. Click on the input control which indicates the current page.  It will say 1 when you start.
3. Try to go to page four.

You can't press backspace and delete the 1 because that would make the current page be blank for a moment.  You can't type 4 and then delete the 1 because that would take you to either page 14 or 41 and there are only 11 pages.  The only thing you can do is highlight the 1 and then press 4 which will set the page number to 4 with just one change event.

It would be great if you could delete the 1 and then type 4.  It would be even better if there was a slight delay on this field.  Changing this field will likely cause a request back to the server to load the data for that page.  If you have 500 pages and are trying to go to page 349 you don't want to have to load the data for page 3, page 34, and then page 349.  If this input field had a 0.25 second delay before it took the input that would provide a much better user experience.

ryshah:
Looking in to this

PR #755, opened 2018-03-27
fix: Add time before onChange and allow input to be empty

ryshah:
Closes https://github.com/carbon-design-system/carbon-components-react/issues/736

When user deletes the input page number (pressing backspace), this change updates the component to use current page for rendering and onChange is not called. 

Also, a timer has been added for `onChange` calls. The default value is set at `250ms` and so refresh (page fetching being most likely scenario) doesn't happen until after that time. 

#### Changelog

**Changed**

* Adds timer before `onChange`
* Allows user to delete input page text field.


zgrossbart:
I’m seeing one bug when running the code in this PR.  If I enter a page number of zero then I get an invalid label.  

screen shot 2018-03-27 at 4 20 22 pm

I'm also seeing an issue with the timer.  If I try to type page 11 it loads page one and then loads page 11 no matter how fast I type.  When I look in the browser console I can see that the `onChange` messages have a delay from the timer, but the page X of Y label updates immediately in the UI.

ryshah:
Fixed the issue for invalid label with `0` input. Reduced the `onChangeTimeout` to `300` in the commit. Need to synch up with @tw15egan or some one from carbon team to discuss page rendering concern that @zgrossbart has. 

Current implementation refreshes UI when user types `1` (while typing 11 or 101). However, `onChange` gets called when user is done typing due to timeout. Thoughts?

asudoh:
Thanks @ryshah for jumping in! Before doing deep dive on the details in the code, my quick reaction is that this looks a typical case where we want to debounce the input. `carbon-components`/`carbon-components-react` codebase use `lodash.deboucne` library to address this problem space and great to see the new code here adhere the pattern. Thanks!

asudoh:
@ryshah `lodash.debounce` is for preventing the function from running if subsequent call happens in the given duration. If you find it's not the case, it'll be great if I can take a look at what you come up with and figure out the best solution. Thanks!

ryshah:
@asudoh Solution in PR is to use `setTimeout` which triggers `onChange` every Nth second (specified by `onChangeTimeout` prop). If another event happens before specified duration, previous event trigger will be cancelled and new one is created.

asudoh:
I see - That’s exactly what `lodash.debounce` does.

src/components/Pagination/Pagination.js
```diff
@@ -57,6 +64,12 @@ export default class Pagination extends Component {
 
   componentWillMount() {
     this.uniqueId = `${Math.floor(Math.random() * 0xffff)}`;
+    this.pageInputDebouncer = debounce(
```
asudoh: This code seems to require the scenario of re-mounting.

src/components/Pagination/Pagination.js
```diff
@@ -180,8 +221,7 @@ export default class Pagination extends Component {
           ) : (
             <TextInput
               id={`bx-pagination-input-${inputId}`}
-              placeholder="0"
```
asudoh: Did you want to kill the `placeholder` here?

src/components/Pagination/Pagination.js
```diff
@@ -82,11 +95,11 @@ export default class Pagination extends Component {
   handlePageInputChange = evt => {
     const page = Number(evt.target.value);
     if (
-      page > 0 &&
+      page >= 0 &&
```
asudoh: @joshblack Would you agree with the change here?

src/components/Pagination/Pagination.js
ryshah: Are you suggesting `flush` the debouncer when component unmounts?

src/components/Pagination/Pagination.js
asudoh: OK - Seems that this prop is for `<input>` placeholder, which won't interfere user input?

src/components/Pagination/Pagination.js
asudoh: Checked `flush` - It doesn't seem something that cancels the debouncer. `cancel()` of the debouncer is much closer, which we can run at `cDU`.

src/components/Pagination/Pagination.js
ryshah: No - placeholder is for when `<input>` component is empty and doesn't interfere with user input.

ryshah:
Any other feedback @joshblack @asudoh ?

src/components/Pagination/Pagination.js
ryshah: Placeholder was not getting used before since page # had to be > 0... We don't want placeholder since if user deletes page # it will put 0 in `input` instead of keeping it empty

src/components/Pagination/Pagination.js
asudoh: OK is it possible for you to add what you described here to the code as a comment? Thanks!

src/components/Pagination/Pagination.js
asudoh: 👍 Thank you for articulating!

ryshah:
Any other suggestions before we can merge this in?

asudoh:
Thanks @ryshah for making the series of changes - IMV the change is fine now, but I’d like review from @joshblack who better knows about the area of data table.

src/components/Pagination/Pagination.js
joshblack: Do we need to create this class instance property in `cWM`? Especially since this method is deprecated in 16.3+. Seems like we could just have:

```js
onPageInput = debounce(page => {
  if (page > 0) {
    this.props.onChange({ page, pageSize: this.state.pageSize });
  }
}, this.props.onChangeInterval)
```

src/components/Pagination/Pagination.js
```diff
@@ -82,11 +99,14 @@ export default class Pagination extends Component {
...
-      page > 0 &&
+      // Added props for defaultPageText and defaultItemText to be
+      // used for case when page # is 0 or empty. For other cases
+      // existing props will be used.
+      page >= 0 &&
       page <= Math.ceil(this.props.totalItems / this.state.pageSize)
     ) {
       this.setState({ page });
-      this.props.onChange({ page, pageSize: this.state.pageSize });
+      this.pageInputDebouncer(page);
```
joshblack: Should we invoke this in the callback to `setState`?

src/components/Pagination/Pagination.js
```diff
@@ -102,25 +122,68 @@ export default class Pagination extends Component {
     this.props.onChange({ page, pageSize: this.state.pageSize });
   };
 
+  getItemsText = () => {
```
joshblack: Are these extra class methods necessary for #736 or are they extra on top of that issue?

src/components/Pagination/Pagination.js
asudoh: Wrt `cWM` - Given JavaScript has no notion of "destructor", we need to pair the lifecycle callbacks for things that need clean-up. I'm guessing that `cDM`/`cWU` is the way to go.

src/components/Pagination/Pagination.js
ryshah: Like @asudoh mentioned in other comment, we might need it to `cancel` debounce with `cWU`

src/components/Pagination/Pagination.js
ryshah: Added the callback.

src/components/Pagination/Pagination.js
ryshah: They are just extracted from inline rendering to its separate method. This was done since adding the logic for page === 0 would need one more ternary operator to existing ternary operator. I felt this would improve readability of the cod.e

src/components/Pagination/Pagination.js
joshblack: @ryshah am fine with having it be cancellable, the main issue is placing it in `cWM` since that lifecycle method is deprecated after 16.3.

ryshah: @joshblack I see your concern. After reading the documentation, looks like React recommends adding subscriptions in `cDM` method instead of `cWM`. I will make that change instead.

ryshah:
@joshblack Changes requested have been addressed. It would be amazing if you can take another pass at this.

src/components/Pagination/Pagination.js
joshblack: Totally understand! More-so was just curious if they are needed to close out #736, or if they are additional behavior added on.

shinytoyrobots:
@asudoh - could you take a look at this PR and resolve/confirm @joshblack's requested changes.

PR #760, opened 2018-03-28
fix(DataTable): removed hardcoded id, resolves #735

### LINKED ISSUE #735: Remove Hardcoded id in TableToolbarSearch
## Detailed description

This is a bug in the component itself. Having the id hardcoded is causing 'non-unique id' errors.
This issue appears to have been fixed in the Search component but not in the TableToolbarSearch (for reference) 

## Additional information
screen shot 2018-03-20 at 7 55 27 pm

bellsworth13:
Fixes #735 

**New**

TableToolBarSearch can be passed id prop

**Removed**

Hardcoded Id


src/components/DataTable/TableToolbarSearch.js
```diff
@@ -34,6 +35,7 @@ TableToolbarSearch.propTypes = {
   /**
    * Provide an optional class name for the search container
    */
+  id: PropTypes.string,
```
asudoh: Please add a comment for the new `id` prop here so that the prop definitions and the comments are in sync.

joshblack: Also, is this a prop type definition that needs to be required, or are we providing a fallback value?

joshblack:
Thanks for working on this @bellsworth13!

joshblack:
cc @asudoh to re-review 🙂

Issue #708, opened 2018-03-13
MultiSelect does not have an accurate count when receiving items


joshblack:
From Yajiao Zhang:

> And another problem about `MultiSelect` with `initialSelectedItems`:
> - I wrote my own components to show the selected items in the tags
> - I implement my `onChange` method of `MultiSelect` to change the state of the page, when a new item is added, the state will be updated then the page will re-rendered and the new tags shows
> - but i found the selection number become doubled when the render method is called

![5c0244c830916f13](https://user-images.githubusercontent.com/3901764/37346836-6c6777d8-269e-11e8-9455-d2928b9fdac8.jpg)

joshblack:
Closed by #761

PR #761, opened 2018-03-29
fix(MultiSelect): handle case of MultiSelect with initialSelectedItems

### LINKED ISSUE #707: [Regression] MultiSelect does not un-select properly
Seems like we might have had a regression recently with the way MultiSelect handles selection. Including a gif to reproduce below:

kapture 2018-03-13 at 9 08 48

rezak-otmani:
Closes #707  and #708

`componentWillReceiveProps` re-insert the removed items systematically

#### Changelog

**New**

**Changed**
**Removed**
`componentWillReceiveProps`


joshblack:
Hi @rezak-otmani! 👋 Thanks for taking the time to make this PR! One last note before reviewing, could you add test cases for the two issues to help protect against these regressions in the future? Thanks!

asudoh:
Hi @rezak-otmani thank you for trying to address the problem, one quick question - Did you intend to kill `prop` -> `state` sync for the component?

rezak-otmani:
Hi @asudoh not sure understand what you mean, but I think that I do not intend to do it

asudoh:
@rezak-otmani OK - Then I believe `cWRP` code you removed should come back. Do you want to work on this? Thanks!

rezak-otmani:
Hi @asudoh I modifed the PR, just changing the condition `(itemsToAdd.length> 0)` which is still true as long as there are the initialSelectedItems 
this condition is modifed in a way that `cWRP` does not re-insert the elements de- inserted

asudoh:
@joshblack Would you agree with the latest fix? Looks simple by scratching the surface, but need expertise from one who deeply understands the codebase. Thanks!

src/internal/Selection.js
```diff
@@ -23,7 +23,7 @@ export default class Selection extends React.Component {
       return selectedItems.indexOf(item) === -1;
     });
 
-    if (itemsToAdd.length > 0) {
+    if (selectedItems.indexOf(itemsToAdd) !== -1) {
```
joshblack: In this block, would `itemsToAdd` be an array? Seems like this branch would be unreachable as a result. Do we need it then?

rezak-otmani: yes, `itemsToAdd` is an array that contains `initialSelectedItems` de-inserted, 
so the contents of `itemsToAddd` should not be re-inserted,
I think adding items here must be removed,  but what does do `cWRP` then

joshblack: Potentially nothing. I think the goal here originally was to select things that were added to `initialSelectedItems` but weren't in `selectedItems`. If that branch isn't one that makes sense, can totally remove this 😄

asudoh:
@rezak-otmani Seems that `cWRP` is removed again? Would you want to take another look to see what happened with the merging? Thanks!

rezak-otmani:
@asudoh `cWRP` has no impact on other components except the MultiSelect, and it does not have any sense even for this component, so I do not see the usefulness of keeping it

asudoh:
OK thanks @rezak-otmani for clarifying! It's fine as long as @joshblack is good with it, but another thing that makes me wonder is removal of `cWRP` is the only change here...?

rezak-otmani:
yes,  cWRP is the only cause of the issue

asudoh:
Thanks @rezak-otmani for more clarification! My last question is - Will we support changing list of selected items via `props`?

rezak-otmani:
I think it's already done via  state

asudoh:
Just double-checking - Do you think the new code covers the following scenario `props.initialSelectedItems` being `[{ id: 'id-foo', label: 'label-foo', value: 'value-foo' }, { id: 'id-bar', label: 'label-bar', value: 'value-bar' }]` initially and then changed to `[{ id: 'id-foo', label: 'label-foo', value: 'value-foo' }, { id: 'id-baz', label: 'label-baz', value: 'value-baz' }, { id: 'id-baz', label: 'label-bar', value: 'value-bar' }]`? You said it's done via state, but wanted to see how. Thanks!

joshblack:
@asudoh if I understand correctly, I think the intent behind the `cWRP` code was to update selected items with what was in `initialSelectedItems` if it was updated. I think the naming of that prop suggests that it should only apply to first use (e.g. the initial stage) which makes this patch potentially valid in addition to fixing the current issue people are facing with MultiSelect. Definitely was an error on my side with the initial component logic here, I think 😅

asudoh:
OK @joshblack thank you for clarifying!!

PR #765, opened 2018-03-29
fix(RadioTile): Accepts className in props

ahsannaveed007:

`label` in `RadioTile` now accepts className via Props

#### Changelog

**New**

```
static propTypes = {
          checked: PropTypes.bool,
          className: PropTypes.string,
          defaultChecked: PropTypes.bool,
          id: PropTypes.string,
          name: PropTypes.string,
          onChange: PropTypes.func,
          value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
  };
```

**Changed**

* `className: PropTypes.string`

**Removed**

* N/A

### DIFF
--- a/src/components/RadioTile/RadioTile-test.js
+++ b/src/components/RadioTile/RadioTile-test.js
@@ -65,8 +65,8 @@ describe('RadioButton', () => {
         expect(label.text()).toMatch(/test label text/);
       });
 
-      it('should have the class passed through props', () => {
-        expect(label.hasClass(label.props.className)).toEqual(true);
+      it('should not have the className when no class is passed through props', () => {
+        expect(label.hasClass(label.props.className)).toEqual(false);
       });
     });
 
--- a/src/components/RadioTile/RadioTile-test.js
+++ b/src/components/RadioTile/RadioTile-test.js
@@ -64,6 +64,10 @@ describe('RadioButton', () => {
         wrapper.setProps({ labelText: 'test label text' });
         expect(label.text()).toMatch(/test label text/);
       });
+
+      it('should have the class passed through props', () => {
+        expect(label.hasClass(label.props.className).toEqual(true));
+      });
     });
 
     describe('wrapper', () => {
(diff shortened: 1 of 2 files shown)


src/components/RadioTile/RadioTile.js
```diff
@@ -16,6 +17,7 @@ export default class RadioTile extends React.Component {
 
   static defaultProps = {
     onChange: () => {},
+    className: null,
```
joshblack: Don't need this in `defaultProps` since `classNames` will have similar behavior with `undefined` 😄

src/components/RadioTile/RadioTile.js
```diff
@@ -29,9 +31,14 @@ export default class RadioTile extends React.Component {
   render() {
     const { children, ...other } = this.props;
 
-    const classes = classNames('bx--tile', 'bx--tile--selectable', {
-      'bx--tile--is-selected': this.props.checked,
-    });
+    const classes = classNames(
+      this.props.className,
```
joshblack: Could add `className` to the `const { children, ...other } = this.props` de-structure statement above 🙂

src/components/RadioTile/RadioTile-test.js
```diff
@@ -64,6 +64,10 @@ describe('RadioButton', () => {
         wrapper.setProps({ labelText: 'test label text' });
         expect(label.text()).toMatch(/test label text/);
       });
+
+      it('should not have the className when no class is passed through props', () => {
```
joshblack: Could you add a test for a custom class name as well? Thanks! 🙌

src/components/RadioTile/RadioTile.js
ahsannaveed007: Thanks for reviewing the PR much appreciated - will make the changes 👍

src/components/RadioTile/RadioTile.js
ahsannaveed007: will do 👍

src/components/RadioTile/RadioTile-test.js
ahsannaveed007: will do 👍

ahsannaveed007:
You are very welcome - looking forward to making some more contributions 😄

ahsannaveed007:
Thanks @asudoh  and no problem hope to submit more PR's in future 👍

PR #767, opened 2018-03-30
fix(Modal): change ariaLabel to aria-label

joshblack:
#### Changelog

**New**

**Changed**

- Change `ariaLabel` to `aria-label` in `Modal`

**Removed**


loganmccaul:
LGTM 👍

PR #768, opened 2018-03-30
fix(Modal): update ariaLabel to be aria-label and add it to the story

loganmccaul:
Aria is actually kinda strange in react. It is one of the few html attributes that is supported in it's original state and not camelCased. https://reactjs.org/docs/accessibility.html

This means that react doesn't translate ariaLabel to be `aria-label` in the html element. You must instead define it as `aria-label` in react.

This goes for all `aria-*` attributes.


loganmccaul:
Duplicate of #767

PR #769, opened 2018-03-31
fix(floating-menu): positioning in scrollable container maintained

psachs21:
Override style of default render to position offscreen to the right
to prevent the window.scrollY from being affected right before
measuring the size of the menu.

This seems to only be an issue in the latest version of react (16.3) probably
due to the asynchronous rendering changes.

### DIFF
--- a/src/internal/FloatingMenu.js
+++ b/src/internal/FloatingMenu.js
@@ -318,25 +318,26 @@ class FloatingMenu extends React.Component {
   _getChildrenWithProps = () => {
     const { styles, children } = this.props;
     const { floatingPosition: pos } = this.state;
-    const style = pos
-      ? Object.assign(
-          {
-            left: `${pos.left}px`,
-            top: `${pos.top}px`,
-            position: 'absolute',
-            right: 'auto',
-            margin: 0,
-            opacity: 1,
-          },
-          styles
-        )
+    // If no pos available, we need to hide the element (offscreen to the right)
+    // This is done so we can measure the content before positioning it correctly.
+    const positioningStyle = pos
+      ? {
+          left: `${pos.left}px`,
+          top: `${pos.top}px`,
+          right: 'auto',
+        }
       : {
           right: `${window.innerWidth}px`,
           top: '0px',
-          position: 'absolute',
         };
     return React.cloneElement(children, {
-      style,
+      style: {
+        ...styles,
+        ...positioningStyle,
+        position: 'absolute',
+        margin: 0,
+        opacity: 1,
+      },
     });
   };
 
--- a/src/internal/FloatingMenu.js
+++ b/src/internal/FloatingMenu.js
@@ -327,7 +327,7 @@ class FloatingMenu extends React.Component {
           right: 'auto',
         }
       : {
-          right: `${window.innerWidth}px`,
+          left: `${window.innerWidth}px`,
           top: '0px',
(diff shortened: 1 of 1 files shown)


asudoh:
Thanks alot @psachs21 for finding the issue and trying to fix! The fix signals that we need to better understand the problem ourselves. Meanwhile, if you have more details on your findings it’ll very nice. Thanks!

psachs21:
@asudoh the issue is pretty simple: when the menu first renders, it's rendered as is (without position: absolute) in the regular flow of the document. So depending on the layout of the document, it may adjust the scrollY position (if there are enough elements). This was theoretically a bug before react 16.3, but I suspect was only exposed due to async rendering changes in core React. I think the fix of positioning the element of screen is the easiest solution.

src/internal/FloatingMenu.js
```diff
@@ -318,21 +318,26 @@ class FloatingMenu extends React.Component {
...
-              top: `${pos.top}px`,
-              position: 'absolute',
-              right: 'auto',
-              margin: 0,
-              opacity: 1,
-            },
-            styles
-          ),
-        });
+    const style = pos
```
joshblack: As a code style comment, could we re-write this in a way that could communicate why these values are needed? Comments are fine as well 🙂

As an added note, why in this ternary do we apply `styles` when `pos` is defined but not when it isn't defined? Trying to catch up on this component implementation 😄

psachs21: @joshblack updated. Should be a little clearer now.

I was thinking I didn't want to apply stylings if we're just hiding it off to the side but you're right, if those stylings change the size of the element, that should be included. Updated.

psachs21:
@asudoh I've also been able to reproduce the issue in the playground, you just need to set some css:

```css
html, body, #root, [role='main'] {
  height: 100%;
}
```

And make sure your example story has a top level div that also has height: 100%.

I haven't submitted an update to the story because I don't think it serves much of a purpose other than to demonstrate the bug.

src/internal/FloatingMenu.js
```diff
@@ -318,21 +318,27 @@ class FloatingMenu extends React.Component {
...
+    // If no pos available, we need to hide the element (offscreen to the left)
+    // This is done so we can measure the content before positioning it correctly.
+    const positioningStyle = pos
+      ? {
+          left: `${pos.left}px`,
+          top: `${pos.top}px`,
+          right: 'auto',
+        }
+      : {
+          right: `${window.innerWidth}px`,
```
asudoh: Thanks @psachs21 for explaining your finding in detail! Your fix makes sense basically, just one thing - Given you want to make the temporally menu location absolute and off-screen - Did you mean `left` here?

psachs21: It's currently positioned off the the left of the screen, but we can put it off to the right as well. Both are roughly equivalent.

asudoh: Yeah I thought the same - Like `right: `${window.innerWidth}px` or `left: 0`.

asudoh: ^ I guess we'd choose either of those. Thanks!

psachs21: hmm, your right, I had a brain fart. Left would be the natural fit for me. I'll update the PR.

PR #772, opened 2018-04-02
feat(ComboBox,DropdownV2): support string values in 'initialSelectedItem' prop

Yohanna:
Closes carbon-design-system/carbon-components-react#718

The PR adds support for string values in `initialSelectedItem` prop in `ComboBox` & `DropboxV2`. The rationale behind this changes in the linked issue.

`defaultItemToString` in both component will now check first if the `item` is of type `string` and in that case will return the item as it is. Otherwise, it will use the old behavior where it will assume it's an Object and returns the `label` property in it.

I also added tests for the `initialSelectedItem` prop in `DropdownV2` since they didn't exist before, however, I wasn't sure how to access the value of each selected item, so I had to use a different way (using `.text()`) than the tests in `ComboBox`

#### Changelog

**New**

* Support for string values in 'initialSelectedItem' proptype & updated tests for `ComboBox` & `DropdownV2`

**Changed**

* Methods docs style (for better intellisense)


src/components/ListBox/test-helpers.js
```diff
@@ -39,23 +39,33 @@ export const assertMenuClosed = wrapper => {
...
 };
 
-// `GenericItem` corresponds to an item in a collection that is passed to
-// MultiSelect that is in a predictable shape and works with the default
-// `itemToString` out of the box.
+/**
+ * `GenericItem` corresponds to an item in a collection that is passed to
+ * MultiSelect that is in a predictable shape and works with the default
+ * `itemToString` out of the box.
+ * @param {number} index
```
joshblack: Want to add the `@returns` annotation in these JSDoc blocks as well? 😄

src/components/DropdownV2/DropdownV2-test.js
```diff
@@ -57,4 +57,33 @@ describe('DropdownV2', () => {
...
     });
   });
+
+  describe('should display initially selected item found in `initialSelectedItem`', () => {
+    it('using an object type for the `initialSelectedItem` prop', () => {
+      const wrapper = mount(
+        <DropdownV2 {...mockProps} initialSelectedItem={mockProps.items[0]} />
+      );
+
+      expect(
```
joshblack: As a code formatting comment, could we do:

```js
expect(wrapper.find('span.bx--list-box__label').text()).toEqual(mockProps.items[0].label);
```

Yohanna:
No problem, I'll work on the changes now.

src/components/ListBox/test-helpers.js
Yohanna: For some reason the `return` annotation is not detected for `generateItems` when I do `@returns {Object[]}` or `@returns {Array<Object>}`. I'm using VSCode.

Should I just add `@returns {Array<Object>}`?

src/components/DropdownV2/DropdownV2-test.js
Yohanna: Good point.

src/components/ListBox/test-helpers.js
Yohanna: Done ✔️

src/components/DropdownV2/DropdownV2-test.js
Yohanna: Done ✔️

PR #777, opened 2018-04-02
fix(CodeSnippet): added prop to CodeSnippet for CopyButton props

davidmenendez:
Closes carbon-design-system/carbon-components-react#776

Adds a new prop `copyProps` to the CodeSnippet component that's passed to the CopyButton component inside.

#### Changelog

**Changed**

* `CodeSnippet`
  * added the `copyProps` prop to contain the props for the `CopyButton` component that it uses for additional customization.


src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -7,7 +7,7 @@ const CodeSnippet = ({
   className,
   type,
   children,
-  onClick,
+  copyProps,
```
asudoh: Thanks @davidmenendez for improving the code to make allow consumers more control of copy button in code snippet! Looks that the change breaks the original `<CodeSnippet onClick={onClickHandler}>` case. Is it possible for you to update this PR to keep the backward-compatibility? Thanks!

davidmenendez: yeah for sure. it had occurred to me this would break it, but i thought my proposed solution was a good way to go just to have a clear separation between the two sets of props.

asudoh: @davidmenendez I see your solution is better, too - The consumers of this project has become sensitive for breaking changes, so we want to make sure we avoid breaking changes.

davidmenendez:
updated based on feedback 👍

joshblack:
It seems like for the corresponding issue we just need to add support for the `feedback` prop versus adding a `copyProps` `prop`. Is this the case?

davidmenendez:
@joshblack i had considered just adding that, but I figured since tooltip has a few other props you can pass it why not just allow it to be consumed with all the options instead of just two?

joshblack:
@davidmenendez I think that could easily be the case, for sure. I do feel like that could prompt a different solution where we'd want to invert the control of this component, e.g. instead of:

```jsx
<CodeSnippet prop1 prop2 />
```

We could do:

```jsx
// Allow groups to compose components
<CodeSnippet prop1>
  <CodeSnippetContent>
    ...
  </CodeSnippetContent>
  <CopyButton prop2 />
</CodeSnippet>

// Allow custom render of copy
<CodeSnippet prop1 renderCopy={() => ...} />

// etc.
```

I think there could also be a case of YAGNI here. While it might be useful in the future, I think we could defer to supporting only what is needed at this moment. If this changes in the future, then we could pivot and re-evaluate with a stricter sense of requirements for this component.

davidmenendez:
I can see that argument. When I wrote this I was just trying to cover all my bases. While I only need that feedback updated, I definitely can't say others wouldn't need to use `iconDescription` or `feedbackTimeout` in this scenario. Plus both components can use `className`. To me it was the easiest way I could think of to separate the props from one another. If everyone would prefer I just add the feedback prop by itself to codeSnippet then i'm fine with that since that's all I personally need.

joshblack:
Let's add in `feedback` for now, if we get requests for additional props for CopyButton then I think we can look to what you've done here to update the component 🙂 Thank you again for taking the time to do this, really appreciate it! 🙌

davidmenendez:
updated

joshblack:
Sorry, I meant literally the propType would be:

```
CodeSnippet.propTypes = {
  feedback: PropTypes.string,
  // or could do this
  copyFeedback: PropTypes.string
};
```

Versus `copyProps`. In the case of `copyProps`, we would want the prop type to mirror the prop types for `CopyButton`, e.g.

```
CodeSnippet.propTypes = {
  copyProps: PropTypes.shape(CopyButton.propTypes),
};
```

Which just kind of feels like if we want folks to control all the props, then we should just expose `CopyButton` in some way. This is the motivation behind just supporting the prop for this one use-case versus opening it up to the idea of full customization 👍

davidmenendez:
makes sense

PR #782, opened 2018-04-03
fix(list-box): Prevent focus on icon inside selection

psachs21:
This prevents a weird issue in IE where focus on the svg causes an
exception in downshift.

The issue occurs only in IE. Basically, downshift tries to manage focus after the input loses focus, and if the element is an SVG, downshift has a fit. This quick fix prevents that on our side.

### DIFF
--- a/src/components/ListBox/ListBoxSelection.js
+++ b/src/components/ListBox/ListBoxSelection.js
@@ -42,7 +42,7 @@ const ListBoxSelection = ({
       onKeyDown={handleOnKeyDown}
       title={description}>
       {selectionCount}
-      <Icon name="close" description={description} focusable={false} />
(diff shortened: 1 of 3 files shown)


joshblack:
@psachs21 could you help me understand where `focusable` comes from and how it addresses the underlying issue here? First time learning about that attribute for SVG elements 😮

psachs21:
@joshblack it's a little new to myself as well, I only recently ran into this issue on IE. The issue occurs technically in downshift, where it's accessing `document.activeElement.dataset.toggle` which blows up if the activeElement is an SVG in IE. focusable="false" prevents the svg from becoming the activeElement.

psachs21:
FYI, i've also logged an issue against downshift: https://github.com/paypal/downshift/issues/409

PR #783, opened 2018-04-03
feat(checkbox/overflow-menu): Support wrapping classes

psachs21:
It's more useful to be able to set the classname on the wrapping element
(which can be used to address internal elements)

This will go some way to enable more dynamic theming options.

#### Changelog

**New**

* `wrapperClassName` prop on OverflowMenuItem and Checkbox.


joshblack:
Could this be considered a breaking change since we're changing the location of where certain props are applied in the inner components?

psachs21:
@joshblack no, I've made sure to keep the existing locations of where className was applied. It's a little inconsistent accross components (some apply className to the wrapper, some to some inner element).

PR #788, opened 2018-04-06
feat(Dropdown): stop event propagation when ESC is pressed, resolves #600

### LINKED ISSUE #600: ListBox components should stop event propagation when `ESC` is pressed
Dave Beshero brought this up in our Slack, but currently on our website you can see this behavior demonstrated when you do the following:

- Visit the component page for Dropdown
- Make sure the carbon menu is open
- Focus an individual dropdown component and open it
- Press escape to close the menu
- Notice that both menus are closed because of event propagation

This issue is to add in the behavior to stop propagation after ESC is pressed so this doesn't happen.

rezak-otmani:
Closes carbon-design-system/carbon-components-react#600

add the `handleKeydown` function to retrieve the event press ESC and stop the spread
add `componentDidMount` in which the event is applied

#### Changelog

**New**
the propagation is stoped after ESC is pressed

**Changed**

**Removed**


joshblack:
Hi @rezak-otmani! #600 is actually for DropdownV2 versus this Dropdown component 😄

rezak-otmani:
Hi @joshblack, yes I mixed things up a bit, it just adds the Dropdown closure with esc

src/components/Dropdown/Dropdown.js
```diff
@@ -50,6 +50,10 @@ export default class Dropdown extends PureComponent {
     }
   }
 
+  componentDidMount() {
```
joshblack: Could we add the accompanying `componentWillUnmount` to remove the listener?

rezak-otmani: yes,  I will add this

src/components/Dropdown/Dropdown.js
```diff
@@ -92,6 +100,12 @@ export default class Dropdown extends PureComponent {
     }
   };
 
+  handleKeydown = evt => {
+    if (evt.which === 27 && this.state.open) {
```
joshblack: Should we use `keyCode` instead of `which` here since `which` is deprecated? Reference: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which

asudoh: Hi @joshblack though `which` is deprecated, IIUC it's the only option that works well with all support browsers (in other words, other options require some browser-specific code).

joshblack: Oh duh, good point @asudoh. `keyCode` is also deprecated too lol.

PR #789, opened 2018-04-06
fix(date-picker-input): default pattern is faulty

psachs21:
The default pattern for the input should be a valid regexp.

I don't know if the intention was to be `\d{1,2}\d{1,2}\d{4}` or if it was to be separated by /s or not.

#### Changelog

**Changed**

* Default `pattern` prop on DatePickerInput is now a valid regexp

### DIFF
--- a/src/components/DatePicker/DatePicker-story.js
+++ b/src/components/DatePicker/DatePicker-story.js
@@ -21,13 +21,13 @@ const datePickerInputProps = {
   onClick: action('onClick'),
   onChange: action('onInputChange'),
   placeholder: 'mm/dd/yyyy',
-  pattern: 'd{1,2}/d{1,2}/d{4}',
+  pattern: '\\d{1,2}\\/\\d{1,2}\\/\\d{4}',
   id: 'date-picker-input-id',
 };
 
 const simpleShortDatePickerInputProps = {
   placeholder: 'mm/yyyy',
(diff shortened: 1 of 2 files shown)


psachs21:
@joshblack turns out the pretty print is what's breaking this after all. What prettier wants for the default is not the same as what we actually want. Not sure how to get around this, to be honest. A jsfiddle of the broken behaviour is here: https://jsfiddle.net/zr4xtjnn/7/. Note that the validation only occurs on firefox.

psachs21:
I've added lint and prettier ignores so the default prop doesn't get formatted.

asudoh:
@joshblack Do you know if the regular expression is something that goes to `new RegExp()`? If so, I'm thinking that we'll need escape for string in addition to escape for regular expression.

joshblack:
So quick question @psachs21, can we pass the regex directly as a prop versus the string representation?

psachs21:
@joshblack @asudoh not that I can see. I tried it here: https://jsfiddle.net/n5u2wwjg/14768/

psachs21:
@asudoh I believe it's escaped properly, is there some other adjustment you wanted? Just need one more review.

asudoh:
Hi @psachs21 thank you for asking - I'm thinking it should be: `'\\d{1,2}\\/\\d{1,2}\\/\\d{4}'`

PR #791, opened 2018-04-07
fix(Slider): Provide sane defaults for onChange, step props

### LINKED ISSUE #775: Slider: Requires more props than are documented
## Detailed description
The Slider component requires both a `step` property (integer) and an `onChange` property (function), otherwise it simply doesn't work with mouse events. This is neither documented anywhere nor are those props marked as `.isRequired` in the component.

I think it would be helpful to either use reasonable defaults for those two, or to mark them as required in the code and the documentation, to keep other devs from wasting time debugging why their fresh-out-of-the-box Slider doesn't work.

Chrome

5.34.0

Check the Code Sandbox below; removing either `step` or `onChange` breaks slider dragging:
https://codesandbox.io/s/q9jvp9vj9

## Additional information
Happy to help resolve this with a PR if you guys don't have the time!

i-a-n:
Resolves #775

See issue for description and bug example. Slider needed better default props for `onChange` and `step`.

**Changed**

* Added default values ("`() => {}`" and "1", respectively) for the above props
* Made small cleanups to the test and storybook files
* Added a quick type check before executing the `onChange` callback function


src/components/Slider/Slider-test.js
```diff
@@ -15,13 +14,8 @@ describe('Slider', () => {
         min={0}
         max={100}
         step={1}
-        onChange={mockFn}>
-        <TextInput
```
asudoh: Any reason removing `<TextInput>` here...?

src/components/Slider/Slider.js
```diff
@@ -104,10 +104,12 @@ export default class Slider extends PureComponent {
 
   static defaultProps = {
     hideTextInput: false,
+    step: 1,
     stepMuliplier: 4,
     disabled: false,
     minLabel: '',
     maxLabel: '',
+    onChange: () => {},
```
asudoh: Do we need this default prop even though the code does `props.onChange` type check?

src/components/Slider/Slider-test.js
i-a-n: Maybe I'm wrong but I didn't see any code or documentation where that is the proper usage of `<Slider />`. It looks like the `TextInput` gets rendered [as part of the component](https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Slider/Slider.js#L382) and is not mentioned anywhere [in the Story](https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Slider/Slider-story.js#L16). I thought it was a mistake, but since I'm new to this project perhaps I misunderstood.

src/components/Slider/Slider.js
i-a-n: Good question! The reason I added the type check was in case someone had passed an incorrect prop type: if someone passes in a string for instance, _instead of a fatal error_, it would simply do nothing (except warn them in the console that the PropType was wrong). If you think this is unnecessary or too much complexity, I'll remove it. Was going back and forth on that one myself.

src/components/Slider/Slider-test.js
asudoh: @tw15egan Wondering if this is something you can chime in...? Thanks!

src/components/Slider/Slider.js
asudoh: Thank you for the thought @i-a-n! The type-check is definitely a good thing 👍 - And I think the type check you added makes the default prop here no longer needed. Thanks!

src/components/Slider/Slider-test.js
tw15egan: Yeah, probably just an oversight in the original creation of the component. The `TextInput` is rendered by the component itself, so not needed in the tests

PR #793, opened 2018-04-09
fix(TooltipSimple): handle case when the tooltip icon not displaying

### LINKED ISSUE #786: ToolTipSimple with no icon not displaying
## [Title]: 

ToolTipSimple with no icon not displaying

## Detailed description

The TooltipSimple without the icon does not display in the storybook. It looks like the before/after pseudo classes are not getting added on hover when there is no icon.

[TooltipSimple](http://react.carbondesignsystem.com/?selectedKind=TooltipSimple&selectedStory=no%20icon&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

rezak-otmani:
Closes #786

Add condition to handle case when the tooltip icon not displaying

#### Changelog
**New**
The TooltipSimple without the icon is displayed 

**Changed**

**Removed**

### DIFF
--- a/src/components/TooltipSimple/TooltipSimple.js
+++ b/src/components/TooltipSimple/TooltipSimple.js
@@ -27,9 +27,7 @@ const TooltipSimple = ({
             tabIndex="0"
             role="button"
             {...other}>
-            {showIcon && (
-              <Icon role="img" name={iconName} description={iconDescription} />
-            )}
+            <Icon role="img" name={iconName} description={iconDescription} />
           </div>
         </div>
       ) : (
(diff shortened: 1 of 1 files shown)


src/components/TooltipSimple/TooltipSimple.js
```diff
@@ -17,18 +17,33 @@ const TooltipSimple = ({
...
+      {showIcon ? (
+        <div className={tooltipWrapperClasses}>
+          {children}
+          <div
+            className={tooltipClasses}
+            data-tooltip-text={text}
+            tabIndex="0"
+            role="button"
+            {...other}>
+            {showIcon && (
```
asudoh: Seems that `showIcon` will never be `false` by the change...?

rezak-otmani: yes  this piece has to be removed since the condition is already verifying and `showIcon` is already true

Issue #720, opened 2018-03-16
DatePickerInput calls 'onClick' prop even if it's not passed to the component


Yohanna:
### Description

`DatePickerInput` component calls the `onClick` prop event if it was not passed in. This results in a `TypeError: _onClick is not a function` error in the console.

#### Is this a feature request (new component, new icon), a bug, or a general issue?
A bug.

#### Is this issue related to a specific component?
Yes. `DatePickerInput`

#### What did you expect to happen? What happened instead? What would you like to see changed?
`onClick` is not a required prop, it shouldn't be called if it wasn't passed to the component.

#### What version of the Carbon Design System are you using?
v`5.31.0`

#### What offering/product do you work on? Any pressing ship or release dates we should be aware of?
A new QRadar App. Initial version released in early April.

### Steps to reproduce the issue
https://codesandbox.io/s/k0wqqw8k27

### Additional information
Mari Johannessen looked into the issue on Slack and found a workaround by passing an empty `onClick`: `onClick={() => { }}`

tw15egan:
Could we simply add a default prop?  `onClick: () => {}`?

PR #794, opened 2018-04-10
fix(DatePickerInput): add default onChange prop

ahsannaveed007:

Add default `onChange` prop in `DatePickerInput` - This fix is similar to the one that has been done for #720 

#### Changelog

**New**
```
static defaultProps = {
    pattern: 'd{1,2}/d{1,2}/d{4}',
    type: 'text',
    disabled: false,
    invalid: false,
    labelText: '',
    onClick: () => {},
    onChange: () => {},
};
```

**Changed**

```
static defaultProps = {
    pattern: 'd{1,2}/d{1,2}/d{4}',
    type: 'text',
    disabled: false,
    invalid: false,
    labelText: '',
    onClick: () => {},
};
```

**Removed**
N/A

### DIFF
--- a/src/components/DatePickerInput/DatePickerInput.js
+++ b/src/components/DatePickerInput/DatePickerInput.js
@@ -15,6 +15,7 @@ export default class DatePickerInput extends Component {
     invalid: false,
     labelText: '',
     onClick: () => {},
+    onChange: () => {},
   };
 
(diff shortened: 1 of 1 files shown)


ahsannaveed007:
Thanks @asudoh  👍

PR #798, opened 2018-04-10
fix(floating-menu): Hide floating menu if no position available

psachs21:
Instead of hiding the menu off to the side (which, as it turns out, causes the text to wrap and extend the height of the element), just make it invisible.

### DIFF
--- a/src/internal/FloatingMenu.js
+++ b/src/internal/FloatingMenu.js
@@ -327,7 +327,7 @@ class FloatingMenu extends React.Component {
           right: 'auto',
         }
       : {
-          left: `${window.innerWidth}px`,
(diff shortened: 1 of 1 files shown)


psachs21:
Another way to do this would be to wrap the element in a position: absolute element and a relative positioned element:

```jsx
<div style={{position: 'absolute'}}>
   {React.cloneElement(child, { style: {position: 'relative'}})}
</div>
```

But this seemed less intrusive.

joshblack:
@asudoh I _think_ if a node has `visibility: hidden` as a style definition then the a11y tree will show `aria-hidden` as `true`.

PR #799, opened 2018-04-11
fix(date-picker): Provide examples for fully controlled datepicker

psachs21:
Also prevent onChange event from being placed on the any other element.

#### Changelog

**Changed**

* datepicker `onChange` prop triggers properly when provided the `value` prop. So the onChange event on the datepicker is only called when a date is selected and not on input change. Input change can be caught on datepicker-input.

### DIFF
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -52,7 +52,13 @@ export default class DatePicker extends Component {
      */
     value: PropTypes.oneOfType([
       PropTypes.string,
-      PropTypes.array,
+      PropTypes.arrayOf(
+        PropTypes.oneOfType([
+          PropTypes.string,
+          PropTypes.number,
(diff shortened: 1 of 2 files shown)


src/components/DatePicker/DatePicker.js
```diff
@@ -47,9 +47,15 @@ export default class DatePicker extends Component {
...
 
     /**
-     * The value of the `<input>`.
+     * The value of the date value provided to flatpickr, could
+     * be a date, a date number, a date string, an array of dates.
      */
-    value: PropTypes.string,
+    value: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.array,
```
asudoh: Wondering if `PropTypes.arrayOf()` can be used here...?

psachs21: done

PR #808, opened 2018-04-13
feat(Modal): pressing enter allows for submission of Modal

### LINKED ISSUE #803: No option to focus on a transactional Modal
## Detailed description

There is no property to focus on an element in a Modal.

Feature request

Modal

I expected the first element (TextInput, Dropdown, etc) in the Modal to be focused when opening, or the primary button to be focused if there are no other elements inside the Modal.

Chrome version 65

Latest

IBM Cloud Account. No pressing release dates.

Find an org in a Bluemix account, go to the overflow menu and click rename
In this case, the TextInput should have focus.

For a case without TextInput or other elements, go to the Remove Domain modal under Domains > click the overflow menu > Remove from Org
In this case, the Remove button (primary button) should have focus.

## Additional information

* I currently have a workaround solution in place locally, and as such would be willing to make the necessary changes on Carbon instead.

akim316:
Closes #803 

For Modals with Forms containing user input, for accessibility and convenience purposes, the Enter/Return press should submit the Modal's form. At first, I was considering having a prop to focus on the first element, but discovered there was no easy way to go about doing this. Since the Modal already had an onKeyDown prop, using ESC to close the modal, I simply applied the same pattern for Enter. If a Modal does not have onRequestSubmit set, then nothing will happen.


akim316:
Yeah, I was considering such use cases but couldn't think of any in particular. Could you chime in on what you had in mind?

asudoh:
@akim316 One example I’d rattle off is search with results, where hitting enter shows the result instead of closing modal.

akim316:
Added option to toggle whether or not enter will submit!

src/components/Modal/Modal.js
```diff
@@ -23,6 +23,7 @@ export default class Modal extends Component {
     primaryButtonDisabled: PropTypes.bool,
     onSecondarySubmit: PropTypes.func,
     danger: PropTypes.bool,
+    enterSubmit: PropTypes.bool,
```
joshblack: How about `shouldSubmitOnEnter` to follow the pattern in some other components? 😄

src/components/Modal/Modal.js
```diff
@@ -131,7 +137,7 @@ export default class Modal extends Component {
     return (
       <div
         {...other}
-        onKeyDown={this.handleKeyDown}
+        onKeyDown={event => this.handleKeyDown(event, enterSubmit)}
```
joshblack: Could we leave this as `this.handleKeyDown`, and then just read from props in that handler?

akim316: Yup, this was my original idea but the ESLint kept yelling about the prop not being used. I believe it is because I don't actually use that prop inside of the render function, just the helper.

PR #810, opened 2018-04-13
Feat: overflow menu item disable

bellsworth13:
Closes carbon-design-system/carbon-components-react#809

Disabled overlow menu item

#### Changelog

**New**

* Added `isDisabled` prop that allows a user to disable an OverflowMenuItem

Changed styles here: https://github.com/carbon-design-system/carbon-components/pull/699


src/components/OverflowMenuItem/OverflowMenuItem.js
```diff
@@ -7,6 +7,7 @@ const OverflowMenuItem = ({
   itemText,
   hasDivider,
   isDelete,
+  isDisabled,
```
alisonjoseph: Can we change this to `disabled` so it follows the pattern that other components are using.

src/components/OverflowMenuItem/OverflowMenuItem.js
```diff
@@ -41,6 +43,7 @@ const OverflowMenuItem = ({
         {...other}
         {...primaryFocusProp}
         className={overflowMenuBtnClasses}
+        disabled={isDisabled}
         onClick={handleClick}>
         {itemText}
       </button>
```
alisonjoseph: Can we add `tabindex="-1"` to the button if `disabled` is true.

PR #811, opened 2018-04-14
feat(click-listener): Able to change ref prop name if desired

psachs21:
Sometimes the children is an HoC whose element is not useful. Allowing
HoCs to pass the ref on to wrapped instances.

This change is made with the hopes that clicklistener can be exposed externally at some point.


joshblack:
Thanks so much for doing this @psachs21! Quick question, could this use-case be accomplished with [`InnerClickListener`](https://github.com/carbon-design-system/carbon-components-react/blob/master/src/internal/InnerClickListener.js)?

psachs21:
@joshblack Ah, indeed it does. I missed this one. Thanks!

PR #813, opened 2018-04-16
fix(OverflowMenu): prevent scroll-into-view behavior upon focus management

### LINKED ISSUE #792: Overflow menu inside Datatable having issues
## Detailed description

After updating to carbon 5.41.1 I saw strange behavior of the OverflowMenu inside the data table

overflow-bug

bug

OverflowMenu

I expected the menu to open normally and not jump the page or not show up at all

Firefox

Currently on 5.29.3, but this issue came up after trying to update to 5.41.1

Armada UI

1. Have table with overflow menu inside
2. Click on overflow menu

asudoh:
Fixes #792.

This change will ensure that focusing on overflow menu body happens after it's fully shown,
and focusing back on the trigger button does not happen when user clicks on another trigger button.

#### Changelog

**New**

* `onPlace` prop in `<FloatingMenu>`, called when the menu is mounted and positioned.

**Changed**

* Using above new `onPlace` callback for focusing on overflow menu body after the menu is shown
* A mechanism to prevent focusing back on the trigger button when user clicks on another trigger button


.storybook/polyfills.js
```diff
@@ -3,3 +3,4 @@ import 'core-js/modules/es6.array.fill';
 import 'core-js/modules/es6.string.includes';
 import 'core-js/modules/es6.string.trim';
 import 'core-js/modules/es7.object.values';
+import './polyfills/element-matches';
```
joshblack: Any way we could do this without using matches? If not, how should we suggest communicating the updated polyfill list here in order to not break teams?

asudoh: @joshblack Two options:

1. Embed the polyfill (Or an equivalent) instead of making it a dependency
2. Use our regular communication channels to advertise the change

asudoh: Took the former approach for now.

PR #819, opened 2018-04-17
feat(OverflowMenuItem): call onClick for Space/Enter key presses

asfordmatt:
Closes carbon-design-system/carbon-components-react#818

Make it easier for consumers to build accessible overflow menus - saves them having to add their own key handlers for each menu item.

#### Changelog

**New**

* To help with Accessibility, OverflowMenuItem now calls `onClick` function for Space/Enter key presses (unless an `onKeyPress` function is already provided)

### DIFF
--- a/src/components/OverflowMenuItem/OverflowMenuItem-test.js
+++ b/src/components/OverflowMenuItem/OverflowMenuItem-test.js
@@ -38,5 +38,63 @@ describe('OverflowMenuItem', () => {
 
       expect(wrapper.hasClass('bx--overflow-menu--divider')).toEqual(true);
     });
+
+    it('calls onClick function when clicked', () => {
+      const clickFn = jest.fn();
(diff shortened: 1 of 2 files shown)


asfordmatt:
@asudoh good point! Yes disabling that code does work, but of course we still want the menu to open on the key events too. Any suggestions as to how we can get both behaviours? Could the element that opens the menu be separate, i.e. not contain the menu items? As the menu items are contained inside the element with the key listener, it is also invoked for the menu items, but we only want it invoked for the open menu button. Perhaps the open menu button could be a sibling to the menu items instead of parent?

asudoh:
Hi @asfordmatt AFAIK opening overflow menu with keyboard should work, at least for the "floating" version - At the code we are discussing. IIRC I saw it worked before. We may want to try going to some older (recent) commits and see how those work.

PR #823, opened 2018-04-18
fix(DataTable): rerender when cell data changes

asfordmatt:
Closes carbon-design-system/carbon-components-react#784

- updates story and test for dynamic cell updates
- add deep compare for row data


asfordmatt:
@joshblack good spot, update made!

PR #825, opened 2018-04-19
fix(OverflowMenu): allow key events to be handled by menu items

asfordmatt:
Closes carbon-design-system/carbon-components-react#818

Allow menu items to be selected by pressing Space/Enter. Previously the key handler caused the menu to close, before the menu items had chance to handle the event. But the menu closes when tabbing out of the menu, so the handler is only needed to open the menu, not to close it.

The fix can be seen from the OverflowMenu storybook - without the fix, selecting menu items using the keyboard doesn't show anything in the Action Logger, but with the fix the onClick action is fired.

#### Changelog

**Changed**

* Accessibility fix: Allow overflow menu items to be selected by pressing Space/Enter

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu-test.js
+++ b/src/components/OverflowMenu/OverflowMenu-test.js
@@ -120,7 +120,21 @@ describe('OverflowMenu', () => {
       expect(rootWrapper.state().open).toEqual(true);
     });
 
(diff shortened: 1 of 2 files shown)


asfordmatt:
@asudoh Thanks for your comments on the previous PR, I think this is a cleaner fix now.

PR #831, opened 2018-04-24
feat(ComposedModal): close ComposedModal when the buttuns is clicked

rezak-otmani:
Close the ComposedModal when the Cancel or save buttun is clicked
#### Changelog

**New**
**Changed**

**Removed**


joshblack:
Hi @rezak-otmani! 👋 Thanks for taking the time to do this 🙂 While we're reviewing the PR, could you take a moment to fill out the changelog in the PR template? Would help a lot for reference looking back to this diff in the future.

marijohannessen:
What's the status on this @asudoh @rezak-otmani ?

PR #856, opened 2018-05-03
fix(ComposedModal): close ComposedModal when ESC is pressed

### LINKED ISSUE #851: ComposedModal's onKeyDown does not work
Currently working on accessibility issues on ICP.  We are not able to use the `esc` key to quickly close the ComposedModal.

---

I see that there is logic to listen for an `esc` key press and set the `open` state to false.

https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/ComposedModal/ComposedModal.js#L22

However, I think the `div` that is being returned by `ComposedModal` needs a `tabIndex='0'` in order to be able to listen to keyboard events in React.

https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/ComposedModal/ComposedModal.js#L75

---

I came across these two sources that mentioned `tabIndex` being required to use `onKeyDown`:

- https://stackoverflow.com/questions/43503964/onkeydown-event-not-working-on-divs-in-react?answertab=active#tab-top

- https://drkbl.com/div-not-response-to-keyboard-event-without-tabindex/

rezak-otmani:
Closes  #851

Close ComposedModal when ESC is pressed

#### Changelog

**Changed**

* Add `tabIndex={-1}`
* Put the `{...other}` before anothers props, so that the component can't override them.

**Removed**


src/components/ComposedModal/ComposedModal.js
```diff
@@ -74,11 +74,12 @@ export default class ComposedModal extends Component {
 
     return (
       <div
+        {...other}
         role="presentation"
         ref={modal => (this.modal = modal)}
         onKeyDown={this.handleKeyDown}
         className={modalClass}
-        {...other}>
+        tabIndex={-1}>
```
joshblack: Would this contradict the `role="presentation"` prop set beforehand?

rezak-otmani: it contradict the onKeyDown function

PR #858, opened 2018-05-04
Invert tooltip direction base on visibility from viewport

nelsonchen90:
Closes carbon-design-system/carbon-components-react#763

Calculate the tooltip visibility on the viewport and invert the position if the defined one doesn't allow user to see the tooltip fully on the window viewport

#### Changelog

**Changed**

* Added logic to detect the out of viewport scenario and invert the direction of the tooltip

### DIFF
--- a/src/components/Tooltip/Tooltip-story.js
+++ b/src/components/Tooltip/Tooltip-story.js
@@ -198,23 +198,23 @@ storiesOf('Tooltip', module)
     the tooltip will render above the element. The example below shows the default scenario.
   `,
     () => (
-      <div style={{ marginTop: '2rem' }}>
-        <Tooltip triggerText="Tooltip label" direction="top">
-          <p className="bx--tooltip__label">Tooltip direction is top</p>
-          <p>
-            But you will see the tooltip appear to the bottom of the trigger
-            element
-          </p>
-          <p>
-            Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do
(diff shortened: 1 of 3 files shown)


nelsonchen90:
@asudoh Can you take a look and see if I get it correctly?
Also I tried to fix that for `OverflowMenu`, but it doesn't seem to support different positions (you can define menuDirection, but the arrow position on the tooltip can't be changed and I think that's a bug. So this fix will only apply to Tooltip

asudoh:
Hi @nelsonchen90 - Thank you for trying to address this problem space!! One question before I do deep dive - Do you have a test case against your code where `<body>` scrolls? Thanks!

nelsonchen90:
@asudoh 
I tested if the body is scrollable `<body style="overflow: scroll;">` this will still invert the direction because this is only depending on the viewport. 
Do you think we should **NOT** invert if the body is scrollable so that user can actually scroll down to see the whole tooltip?

nelsonchen90:
@asudoh 
- fixed an issue the invert logic doesn't work after scroll. 
- added new prop on Tooltip to indicate whether it should use dynamic direction.
- also updated the story book with hopefully better demo

Can you take a look and let me know any changes needed, Thanks a lot!

marijohannessen:
@asudoh what's the status on this?

asudoh:
@nelsonchen90 is looking at addressing the problem without touching our floating menu/tooltip code. Let’s close this for now until the effort turns out that we actually need change in floating menu/tooltip code.

PR #859, opened 2018-05-04
fix(857): Support RHL proxied components in custom type checking

liviudobrea:
Closes carbon-design-system/carbon-components-react#857

This adds `react-hot-loader` as dependency, since new type checking is done via [areComponentsEqual](https://github.com/gaearon/react-hot-loader/blob/f3e1208878583136e9e14c0e2e6b7ffc7db94fea/src/utils.dev.js#L9-L11)
```javascript
export const areComponentsEqual = (a, b) =>
  getProxyOrType(a) === getProxyOrType(b)
``` 

#### Changelog

**New**

* react-hot-loader project dependency

**Changed**

* Updated childrenOf and childrenOfType type checking and tests

### DIFF
--- a/src/prop-types/childrenOf.js
+++ b/src/prop-types/childrenOf.js
@@ -1,4 +1,5 @@
 import { Children } from 'react';
+import { areComponentsEqual } from 'react-hot-loader/patch';
 import createChainableTypeChecker from './tools/createChainableTypeChecker';
 import getDisplayName from './tools/getDisplayName';
 
@@ -21,7 +22,13 @@ const childrenOf = expectedChildTypes => {
         return;
       }
       const childDisplayName = getDisplayName(child.type || child);
-      if (!expectedChildTypes.includes(child.type)) {
+      const expectedChildType = expectedChildTypes.find(
+        c =>
+          typeof c === 'string'
+            ? c === childDisplayName
+            : c.name === childDisplayName
+      );
+      if (!areComponentsEqual(child.type, expectedChildType)) {
         throw new Error(
           `Invalid prop \`children\` of type \`${childDisplayName}\` ` +
             `supplied to \`${componentName}\`, expected each child to be one ` +
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
     "lodash.omit": "^4.5.0",
+    "react-hot-loader": "^4.1.2",
     "warning": "3.0.0",
     "window-or-global": "^1.0.1"
   },
--- a/scripts/rollup.config.js
+++ b/scripts/rollup.config.js
@@ -75,6 +75,7 @@ module.exports = {
           'PropTypes',
           'createElement',
         ],
+        'node_modules/react-hot-loader/patch.js': ['areComponentsEqual'],
         'node_modules/react-dom/index.js': ['render'],
       },
     }),
--- a/src/prop-types/__tests__/childrenOf-test.js
+++ b/src/prop-types/__tests__/childrenOf-test.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import reactHotLoader from 'react-hot-loader/patch';
 import childrenOf from '../childrenOf';
 
 const StatelessComponent = () => <div />;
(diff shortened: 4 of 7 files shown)


src/prop-types/childrenOf.js
```diff
@@ -1,4 +1,5 @@
 import { Children } from 'react';
+import { areComponentsEqual } from 'react-hot-loader/patch';
```
asudoh: @joshblack Would this add `react-hot-loader` dependency to our main React component library?

joshblack: Looking at this, it seems so. Leaving a comment in the general thread.

joshblack:
Hi @liviudobrea! 👋 How hard would it be to embed the `areComponentsEqual` check inside of the utility method versus bringing in RHL as a dependency?

liviudobrea:
@joshblack That would be impossible, since the `areComponentsEqual` method uses an internal registry to check the type of the components you are requiring the equality check for. https://github.com/gaearon/react-hot-loader/blob/master/src/utils.dev.js and https://github.com/gaearon/react-hot-loader/blob/master/src/reconciler/proxies.js
The other method (that would not rely on having RHL as dep) is to check wether the components we're checking contains the `PROXY_KEY` prop https://github.com/gaearon/react-hot-loader/blob/df8ddaedbfd116e688ee7309a9b1f992318ebe41/src/proxy/constants.js#L1-L2.
```javascript
export const PREFIX = '__reactstandin__'
export const PROXY_KEY = `${PREFIX}key`
``` 
Personally, I would not wish for either (not for RHL as dependency, nor for checking for a propname that I'm uncertain of).

src/prop-types/childrenOf.js
```diff
@@ -21,7 +22,13 @@ const childrenOf = expectedChildTypes => {
...
       const childDisplayName = getDisplayName(child.type || child);
-      if (!expectedChildTypes.includes(child.type)) {
+      const expectedChildType = expectedChildTypes.find(
+        c =>
+          typeof c === 'string'
+            ? c === childDisplayName
+            : c.name === childDisplayName
+      );
+      if (!areComponentsEqual(child.type, expectedChildType)) {
         throw new Error(
```
asudoh: @joshblack Thinking more about this issue - Should the condition here *really* be considered fatal?

joshblack:
@liviudobrea I'm down for checking to see if the component is being proxied. Is there a common way to access the underlying proxied component, or is this info stored in that registry that you alluded to?

src/prop-types/childrenOf.js
joshblack: Good point! Should be returning the error instead of throwing it 🙂 Must have missed this in an earlier PR.

liviudobrea:
In RHL there's a helper method attached to every proxied component called
"__reactstandin__getCurrent()" that will return the original instance.

joshblack:
@liviudobrea makes sense! Is it realistic to add a check that basically is looking to see if `__reactstandin__getCurrent` is defined, and if so use that to get access to the instance?

asudoh:
@liviudobrea @joshblack As [this thread](https://github.com/gaearon/react-hot-loader/issues/304) discusses, I see a few possible solutions:

* Don’t treat non-matching React class fatal (Warning-only instead of error)
* `child.type` === `React.createElement(ExpectedComponentClass).type` (Or define a signature hash in the class and compare such signature)
* [RHL 4](https://github.com/gaearon/react-hot-loader/issues/304#issuecomment-353865002)?

liviudobrea:
@asudoh 
- This would be preferred
- This would be unnecessarily expensive for our use case
- That link is broken. [This is the original reference](https://github.com/gaearon/react-hot-loader/#checking-element-types), and as you can see it makes use of `areComponentsEqual`, which requires `RHL` as project dependency.

@joshblack Until this is done https://github.com/gaearon/react-proxy/issues/68 (if ever...), I would currently opt for warning instead of throwing when comparing types, in favour of checking for the magic `__reactstandin__getCurrent` prop.

joshblack:
Relevant PR in the Palantir codebase: https://github.com/palantir/blueprint/pull/2099

joshblack:
Another option that (seemingly) could work: https://github.com/gatsbyjs/gatsby/issues/3486#issuecomment-364167698 

Maybe we can add it in behind a `__DEV__` flag?

marijohannessen:
@liviudobrea @asudoh @joshblack we'll be closing this PR in 7 days if there is no more activity by then!

asudoh:
@liviudobrea My apologies in case you were waiting for my feedback. One of the links @joshblack referred to uses the 2nd approach. My preference is the 1st approach, unless RHL has solved the problem - Hope it makes things clear. Thanks!

marijohannessen:
Closing due to inactivity. Feel free to re-open if needed.

PR #860, opened 2018-05-04
Add Inline Text Cell Edit to the DataTable

joshblack:
Add Inline Cell Editing for Text Inputs in a DataTable.

Preview URL: [Link](http://860-add-inline-cell-edit-storybook-temptable-colobus.mybluemix.net/?selectedKind=DataTable&selectedStory=with%20inline%20cell%20edit&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

#### Changelog

**New**

* `EditableTextCell`
* `EditCellStatus`
* `EditCellActions`
* `EditCellField`
* Inline Edit story for DataTable

**Changed**

* `DataTable`
  * Now supports a global `isEditing` state field
  * Now `getCellProps` will return additional prop helpers for coordinating edit state

**Removed**


src/components/DataTable/EditCellField.js
```diff
@@ -0,0 +1,99 @@
...
+      'bx--data-table__edit-field': true,
+      'bx--data-table__edit-field--invalid': error,
+    });
+    return (
+      <React.Fragment>
+        <div className={className} ref={this.handleRef}>
+          <label className="bx--label" htmlFor={id}>
+            {labelText}
+          </label>
+          <input
```
Noodles006: I am wondering whether this is gonna to just support InputText editing? What about `Select`?

joshblack:
cc @aagonzales for visual review! There is a preview link to check it out in the storybook 👍

aagonzales:
The visuals look good! There's just a few interaction tweaks. 
- [ ] Is there a way to disable the row hover when the edit field is active?
- [ ] When it first opens can we already have the focus on the field? So the user doesn't have to click again to start typing?
- [ ] The validation message is supposed to open when the user focuses on the invalid field.

joshblack:
@aagonzales thanks for the feedback! Update the preview link with the requested changes 🙂

.storybook/polyfills.js
```diff
@@ -1,3 +1,7 @@
+// Used solely for storybook development
+import '@babel/polyfill';
```
asudoh: I’d suggest maintaining a separate polyfills list for Story-book env-specific polyfills, like [one in vanilla](https://github.com/carbon-design-system/carbon-components/blob/b282853/demo/polyfills/devenv.js). Also, I’d suggest staying away from importing the entire Babel polyfill as doing so will deprive our way to test against IE11 for the list of polyfills.

config/jest/setup.js
```diff
@@ -4,6 +4,7 @@ jest.unmock('promise');
 jest.unmock('whatwg-fetch');
 jest.unmock('object-assign');
 
+require('@babel/polyfill');
```
asudoh: Should use [(the original version) of `.storybook/polyfills.js`](https://github.com/carbon-design-system/carbon-components-react/blob/087604a/.storybook/polyfills.js) instead.

src/components/DataTable/EditCellField.js
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  state = {
+    isEditing: true,
+  };
+
+  componentDidMount() {
+    // Set cursor position to end of control
+    this.inputNode.value = '';
+    this.inputNode.value = this.props.value;
```
asudoh: Why two assignment here…?

src/components/DataTable/EditCellStatus.js
```diff
@@ -0,0 +1,33 @@
...
+const EditCellStatus = ({ isSaving, shouldDisplaySuccess }) => (
+  <div className="bx--data-table__edit-status">
+    {isSaving && <Loading small withOverlay={false} />}
+    {!isSaving &&
+      shouldDisplaySuccess && (
+        <Icon
+          className="bx--data-table__icon--success"
+          name="checkmark"
+          role="alert"
+          description="Successfully saved cell"
```
asudoh: Need to move this to a translatable string.

src/components/DataTable/EditableTextCell.js
```diff
@@ -0,0 +1,214 @@
...
+          ) : (
+            <div className="bx--data-table-cell__activity">
+              <button
+                className="bx--data-table-cell__edit"
+                title={`Edit Cell: ${initialValue}`}
+                onClick={this.handleOnEdit}>
+                <Icon
+                  name="edit--glyph"
+                  className="bx--data-table-cell__icon--edit"
+                  description={`Edit Cell: ${initialValue}`}
```
asudoh: Need to move this to a translatable string.

src/components/DataTable/_data-table.scss
```diff
@@ -0,0 +1,362 @@
+@import 'carbon-components/scss/globals/scss/helper-mixins';
```
asudoh: Should move this file to `carbon-components`.

src/components/DataTable/DataTable-story.js
```diff
@@ -98,81 +15,7 @@ storiesOf('DataTable', module)
...
-                    ))}
-                  </TableRow>
-                ))}
-              </TableBody>
-            </Table>
-          </TableContainer>
-        )}
-      />
-    )
+    require('./stories/default').default
```
asudoh: Should use ESM import.

joshblack: Any reason?

asudoh: Given the codebase is standardized on ESM, avoiding CJS `require()` helps our codebase look more coherent. Doing so avoids the kind of bugs e.g. one like a recent date picker PR introduced.

src/components/DataTable/EditCellField.js
joshblack: Will actually reset the input to have the correct cursor position at the end of the string.

config/jest/setup.js
joshblack: The big issue is that we need `regeneratorRuntime` for `async` support, which is why this is being brought in.

.storybook/polyfills.js
joshblack: > importing the entire Babel polyfill

It's important to note the `babel-preset-env` config set to `'usage'` to help out with this situation.

src/components/DataTable/DataTable-story.js
joshblack: Have a link to the bug for date picker? Would love to understand that, didn't realize something had happened 😮

.storybook/polyfills.js
asudoh: Babel is eager to bring in polyfills - I believe you have been surprised how many Babel polyfills was brought in for a simple case, when we discussed this topic before?

config/jest/setup.js
asudoh: Have you tried `babel-plugin-transform-runtime` with `polyfill` option turned off? You can refer to how vanilla test env is set up.

src/components/DataTable/DataTable-story.js
asudoh: Here it is: https://github.com/carbon-design-system/carbon-components-react/pull/869#issuecomment-387591240

config/jest/setup.js
joshblack: Definitely can use that instead!

Noodles006:
@joshblack Thanks for your awesome work. I checked your changes in this PR. And I just found inline editable TextInput, do you still implement inline editable Select?

joshblack:
Hi @Noodles006! 👋 This PR will just bring in TextInput. We 100% want to add in the other input types, just am having to go one type at a time since the scope of this work is so large. Let me know if you want to help contribute any of the other input types and I can set up something to go over how we're thinking inline editing is going to work with those types! 🙂

marijohannessen:
What's the plan for this PR @joshblack ?

joshblack:
Good question @marijohannessen! I'm honestly not too sure. Can close until this feature gets re-prioritized, I guess?

PR #862, opened 2018-05-04
fix(ListBox): prevent key events from propagating

joshblack:
Closes carbon-design-system/carbon-components-react#600

#### Changelog

**New**

**Changed**

* `package.json`
  * Updated `downshift` dependency
* `src/components/ListBox`
  * Now prevents event propagation for key events through `onKeyDown`

**Removed**

### DIFF
--- a/src/components/ListBox/ListBox.js
+++ b/src/components/ListBox/ListBox.js
@@ -6,6 +6,10 @@ import ListBoxMenu from './ListBoxMenu';
 import { ListBoxType } from './ListBoxPropTypes';
 import childrenOf from '../../prop-types/childrenOf';
 
+const handleOnKeyDown = event => {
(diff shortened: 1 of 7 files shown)


src/components/ListBox/ListBox.js
```diff
@@ -6,6 +6,10 @@ import ListBoxMenu from './ListBoxMenu';
 import { ListBoxType } from './ListBoxPropTypes';
 import childrenOf from '../../prop-types/childrenOf';
 
+const handleOnKeyDown = event => {
+  event.stopPropagation();
```
asudoh: Should we explicitly check for ESC key…?

joshblack: Good catch! Will add that in 🙂

Issue #868, opened 2018-05-08
flatpickr uses window object in isomorphic applications for DataPicker


codejamninja:
I get the following error when using [carbon-components-react](https://github.com/carbon-design-system/carbon-components-react) with [gatsbyjs](https://www.gatsbyjs.org).

The error is coming from [flatpickr](https://github.com/flatpickr/flatpickr) [HERE](https://github.com/flatpickr/flatpickr/blob/1882ea322e11bdfd5376bce6affb0d7d8562c38b/src/index.ts#L2013).

The error is a result of the window object not existing in node. Essentially this library is not compatible with isomorphic react frameworks unless the window objects are conditionally handled in cases where they do not exist. I noticed that you are doing this throughout most of your library with the [window-or-global](https://www.npmjs.com/package/window-or-global) npm module.

I'm assuming this was missed because flatpickr is a third party, and you don't have control over how they handle the window object. I suggest that you conditionally import this package, and in the case that the window does not exist, render an empty `<div />` or `<input />`.

In an isomorphic application, once the client javascript kicks in, the flatpickr will render.

```sh
error Building static HTML for pages failed

See our docs page on debugging HTML builds for help https://goo.gl/yL9lND

  1767 | 	position: "auto",
  1768 | 
> 1769 | 	animate: window.navigator.userAgent.indexOf("MSIE") === -1,
       | ^
  1770 | 
  1771 | 	/* if true, dates will be parsed, formatted, and displayed in UTC.
  1772 |  preloading date strings w/ timezones is recommended but not necessary */


  WebpackError: window is not defined
  
  - flatpickr.js:1769 Object.<anonymous>
    ~/flatpickr/dist/flatpickr.js:1769:1
  
  - DatePicker.js:14 Object.assign.i
    ~/carbon-components-react/lib/components/DatePicker/D    atePicker.js:14:1
  
  - index.js:13 Object.defineProperty.value
    ~/carbon-components-react/lib/components/DatePicker/i    ndex.js:13:1
  
  - index.js:777 Object.defineProperty.value
    ~/carbon-components-react/lib/index.js:777:1
  
  - header.js:3 Object.exports.__esModule
    src/components/header.js:3:1
```

joshblack:
[...]

PR #869, opened 2018-05-08
Fixed flatpickr isomorphic error

codejamninja:
Fixed issue #868

Fixed flatpickr isomorphic error

#### Changelog

**Changed**

* Only import flatpickr when window exists


joshblack:
Hi @codejamninja! 👋 Thanks for making this PR! Seems like the status check failed because of file formatting differences. Can you make sure to run `yarn format` in a new commit? There should be a git hook that is registered when you run `yarn install` as well, supposedly, which can help catch these kinds of things in the future 👍

codejamninja:
I'm sorry, but happens when FlatPickr is loaded. I get the error even when I don't use FlatPickr.

codejamninja:
Is there a way to conditionally determine if it is UMD?

asudoh:
@codejamninja The problem is *not* with "UMD" as the matter of fact - The problem happens whenever the code is built with Rollup. Rollup CommonJS plugin may help, but it breaks the coherency in our codebase where all modules are written in ESM.

Some options I can rattle off are:

* File an issue against https://github.com/flatpickr/flatpickr
* Codemod to wrap the entire FlatPickr code in a function so the code can be lazily-evaluated
* Let the FlatPickr class ref passed along via `props` (Breaking change)

marijohannessen:
What's the status here @codejamninja @asudoh ?

codejamninja:
@marijohannessen @asudoh, I wrapped flatpickr in a separate module.

https://github.com/codejamninja/isomorphic-flatpickr
https://github.com/codejamninja/isomorphic-flatpickr/blob/master/src/index.js

asudoh:
@codejamninja Thank you for your continued interest in this area - Would you have a chance to try latest Flatpickr?

codejamninja:
isomorphic-flatpickr is using the latest

codejamninja:
I think this pull request is not needed anymore.

Issue #868 was fixed by #880

PR #872, opened 2018-05-09
fix: Pagination component's range and page text for no items

### LINKED ISSUE #553: PaginationV2 page menu label blank when there are no items
image

The drop down menu label is blank. Probably this should just not be displayed if there are no pages to choose from. Or at least say "0" and be disabled.

McBrosa:
resolves #553 

**Summary**:
When the total items provided to the Pagination component is 0 the correct item and page ranges will be displayed now. (ie. 0-0 of 0 items and 1 of 1 pages, instead of 1-0 of 0 items and 1 of 0 pages)

**Additional Notes**:
Pagination V2
screen shot 2018-05-09 at 5 31 12 pm

Pagination
screen shot 2018-05-09 at 5 19 08 pm

#### Changelog

**Changed**
* Pagination, PaginationV2: Fix item and page ranges when there are no items.


asudoh:
Hi @McBrosa thank you for trying to address this problem space! Do you think your change has something to do with https://github.com/carbon-design-system/carbon-components-react/pull/755, or is it completely a separate issue?

McBrosa:
They are similar but I believe there are still aspects of #533 that are not resolved in #755.
For example:
```
<Pagination {...props} totalItems={0} />
```
Result:
screen shot 2018-05-14 at 10 00 49 am

PR #879, opened 2018-05-11
fix(Tooltip): fix missing dirty-check in cWRP

### LINKED ISSUE #876: Tooltip closes whenever parent rerenders
## Detailed description

Whenever the tooltip is open, if parent rerenders the tooltip closes. 

It looks like the tooltip has a default property of 'open' set to false. Whenever it receives new props (as part of parent rerender), the tooltip sets state 'open' to the value of prop.open (which is always false). This causes the tooltip to close.

Bug

Tooltip

Expected tooltip to stay open

Chrome & Safari macOS X

5.36.0

[Codesandbox](https://codesandbox.io/s/ox5m94lxj5)

1. Hover on tooltip
2. Wait 3 seconds, it will then close

asudoh:
Fixes #876.

#### Changelog

**New**

* Added missing prop dirty-check in `<Tooltip>`’s `cWRP` code.

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -167,7 +167,7 @@ export default class Tooltip extends Component {
 
   componentWillReceiveProps({ open }) {
     /**
-     * so that tooltip can be controlled programmatically thru this `open` prop
+     * so that tooltip can be controlled programmatically through this `open` prop
      */
     const { open: origOpen } = this.props;
     if (origOpen !== open) {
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -165,13 +165,16 @@ export default class Tooltip extends Component {
     });
   }
 
-  componentWillReceiveProps(newProps) {
+  componentWillReceiveProps({ open }) {
     /**
      * so that tooltip can be controlled programmatically thru this `open` prop
```
joshblack: Only nit would be sentence case and `thru` -> `through`

PR #880, opened 2018-05-12
fix(DatePicker): lazily evaluate FlatPickr module

### LINKED ISSUE #868: flatpickr uses window object in isomorphic applications for DataPicker

asudoh:
Fixes #868.

#### Changelog

**New**

* Mechanism to codemod Flatpickr AMD module, transforming:

From:

```javascript
define([], function () {
  (Flatpickr code containing window object access reported in #868)
  …
  return flatpickr;
});
```

To:

```javascript
define([], function () {
  return function () {
     (Flatpickr code containing window object access reported in #868)
     …
     return flatpickr;
  };
});
```

* `<DatePicker>` code to evaluate such function-wrapped version of Flatpickr module. It automatically detects if the module is function-wrapped or not, to support applications importing Carbon modules directly (where our codemod is not automatically applied).

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,34 +1,47 @@
 const path = require('path');
 
-module.exports = {
-  module: {
-    rules: [
-      {
-        test: /\.scss$/,
-        use: [
-          { loader: 'style-loader' },
-          {
-            loader: 'css-loader',
-            options: { importLoaders: 2 },
+module.exports = storybookBaseConfig => {
+  storybookBaseConfig.module.rules.push(
+    {
+      test: /flatpickr\.js$/,
+      use: [
+        {
+          loader: 'babel-loader',
+          options: {
+            babelrc: false,
+            plugins: [
+              require.resolve('../scripts/babel-plugin-wrap-amd-factory'),
+            ],
           },
-          {
-            loader: 'postcss-loader',
-            options: {
-              plugins: () => [
-                require('autoprefixer')({
-                  browsers: ['last 1 version', 'ie >= 11'],
-                }),
-              ],
-            },
+        },
+      ],
+    },
+    {
+      test: /\.scss$/,
+      use: [
+        { loader: 'style-loader' },
+        {
+          loader: 'css-loader',
+          options: { importLoaders: 2 },
+        },
+        {
+          loader: 'postcss-loader',
+          options: {
+            plugins: () => [
+              require('autoprefixer')({
+                browsers: ['last 1 version', 'ie >= 11'],
+              }),
+            ],
           },
-          {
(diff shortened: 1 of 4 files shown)


joshblack:
Related issue on main project: https://github.com/flatpickr/flatpickr/issues/1363

Do we believe they'll be timely in a fix upstream?

asudoh:
Probably this is it? https://github.com/flatpickr/flatpickr/commit/9618cbd8e476a2db38523ca827a296c68afb6efe

joshblack:
I believe this as well, correct? https://github.com/flatpickr/flatpickr/commit/1882ea322e11bdfd5376bce6affb0d7d8562c38b

PR #883, opened 2018-05-14
feat(code-snippet): Add updated code snippet variations

alisonjoseph:
Ref https://github.ibm.com/carbon/issues/issues/658
#535

#### Changelog

**New**

* `Copy` component
*  Added inline code snippet option (Added a `light` prop for the alternate background color, not sure how we want to standardize this across components that will have this option?)

**Changed**

* Renamed terminal -> single and code -> multi (left in support for the old names)

**TODO: Remove scss files from PR before merging**


src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,144 @@
...
+      /*
+      showMoreText,
+      showLessText,
+      */
+      ...other
+    } = this.props;
+
+    const codeSnippetClasses = classNames(className, {
+      'bx--snippet': true,
+      'bx--snippet--single': (type === 'single') | (type === 'terminal'),
```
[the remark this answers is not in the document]
alisonjoseph: Yes, thanks for catching that 👍

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,144 @@
...
+    const codeSnippetClasses = classNames(className, {
+      'bx--snippet': true,
+      'bx--snippet--single': type === 'single' || type === 'terminal',
+      'bx--snippet--multi': type === 'multi' || type === 'code',
+      'bx--snippet--inline': type === 'inline',
+      'bx--snippet--expand': this.state.expandedCode,
+      'bx--snippet--light': light,
+    });
+
+    const expandCodeBtnText = this.state.expandedCode ? 'less' : 'more'; // need the show/more less text props to show here
```
alisonjoseph: @joshblack I was struggling with the best way to show these text values, it wouldn't let me use the props. Any ideas?

src/components/CodeSnippet/CodeSnippet.Skeleton.js
```diff
@@ -1,34 +1,51 @@
...
-    </div>
-  );
-};
+export default class CodeSnippetSkeleton extends Component {
+  static propTypes = {
+    /**
+     * The type of code snippet
+     * can be inline, single or multi
+     */
+    type: PropTypes.string,
```
joshblack: Could we further restrict this prop type definition to:

```js
PropTypes.oneOf(['single', 'inline', 'multi']);
```

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
-  const wrapperClasses = classNames('bx--snippet', className, snippetType);
-  return (
-    <div className={wrapperClasses} {...other}>
+export default class CodeSnippet extends Component {
+  static propTypes = {
+    /**
+     * The type of code snippet
+     * can be inline, single or multi
+     */
+    type: PropTypes.string,
```
joshblack: Could we further restrict this prop type definition to:

```js
PropTypes.oneOf(['single', 'inline', 'multi']);
```

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
+  };
+
+  static defaultProps = {
+    type: 'single',
+    showMoreText: 'Show more',
+    showLessText: 'Show less',
+  };
+
+  state = {
+    showBtn: false,
```
joshblack: What is this state bit toggling? Might help to standardize around the `shouldShowX` convention, e.g. `shouldShowExpandTrigger` (or something way less verbose 😂), to help de-couple the state value from the UI.

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
+    expandedCode: false,
+  };
+
+  componentDidMount = () => {
```
joshblack: Can just have `cDM` as a regular class method since it already has access to instance 👍 

```diff
-  componentDidMount = () => {
+  componentDidMount() {
```

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
+    const codeSnippetClasses = classNames(className, {
+      'bx--snippet': true,
+      'bx--snippet--single': type === 'single' || type === 'terminal',
+      'bx--snippet--multi': type === 'multi' || type === 'code',
+      'bx--snippet--inline': type === 'inline',
+      'bx--snippet--expand': this.state.expandedCode,
+      'bx--snippet--light': light,
+    });
+
+    const more = showMoreText;
```
joshblack: What do we need to alias these values for?

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
+
+    const expandCodeBtnText = this.state.expandedCode ? less : more; // need the show/more less text props to show here
+
+    const moreLessBtn = (
+      <button
+        className="bx--btn bx--btn--ghost bx--btn--sm bx--snippet-btn--expand"
+        type="button"
+        onClick={this.expandCode}>
+        <span className="bx--snippet-btn--text">{expandCodeBtnText}</span>
+        <Icon
```
joshblack: Should we add `aria-hidden` here since the icon is just for presentation?

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -1,43 +1,159 @@
...
+          aria-label={copyLabel}
+          feedback={feedback}>
+          <code>{children}</code>
+        </Copy>
+      );
+    }
+
+    if (type === 'single') {
+      return (
+        <div className={codeSnippetClasses} {...other}>
```
joshblack: Should we re-order this to:

```diff
-        <div className={codeSnippetClasses} {...other}>
+        <div {...other} className={codeSnippetClasses}>
```

Just to make sure `className` doesn't get overridden?

src/components/CodeSnippet/CodeSnippet.js
alisonjoseph: we don't need it, I'll remove

alisonjoseph:
@joshblack I think I addressed all the changes you requested, can you double check when you get a chance? 😄

PR #888, opened 2018-05-15
feat(Icon): introduce pass-thru option

asudoh:
Continuation of https://github.com/carbon-design-system/carbon-components-react/pull/588 - With a bit of breaking change for the sake of `v9` major version.

The breaking change is that app needs to call `setIconsList()` API for older usage to work, like below:

```javascript
import iconsList from 'carbon-icons';
import Icon, { setIconsList } from '../Icon';

setIconsList(iconsList);

...

<Icon name="icon--add" ... />
```

Wrt the newer pattern - The new `icon` prop allows icon data directly passed, and using it along with specific icon data from tree-shakable ESM module of `carbon-icons` beginning `6.3.0` release, app can optimize the bundle size.

The change also allows you to use an icon that is not in `carbon-icons`.

#### Changelog

**New**

* `<IconV2>` component, as mentioned above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -286,7 +286,7 @@
       "/cjs/"
     ],
     "transformIgnorePatterns": [
-      "[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$"
+      "[/\\\\]node_modules[/\\\\](?!(carbon-icons)).+\\.(js|jsx)$"
     ],
     "moduleFileExtensions": [
       "js",
--- a/src/components/Accordion/Accordion.Skeleton.js
+++ b/src/components/Accordion/Accordion.Skeleton.js
@@ -1,13 +1,14 @@
 import React from 'react';
 import Icon from '../Icon';
 import SkeletonText from '../SkeletonText';
+import { iconChevronRight } from 'carbon-icons';
 
 export default class AccordionSkeleton extends React.Component {
   render() {
     const item = (
       <li className="bx--accordion__item">
         <button type="button" className="bx--accordion__heading">
-          <Icon className="bx--accordion__arrow" name="chevron--right" />
+          <Icon className="bx--accordion__arrow" icon={iconChevronRight} />
           <SkeletonText className="bx--accordion__title" />
         </button>
       </li>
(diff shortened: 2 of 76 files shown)


joshblack:
At a high level, what advantages do we have for this approach in comparison to generating the components at build time from the SVG source?

asudoh:
@joshblack Thank you for asking!

> generating the components at build time from the SVG source

Could you please articular what are you looking for when you say this? Thanks!

asudoh:
BTW more background on this PR - Before this change, the icon being used was resolved run-time (by `name` prop passed along to `<Icon>`) which means no static analysis can determine the list of icons in use. This change is for passing along the icon data directly from tree-shakable `carbon-icons` module which allows static analysis to determine what icons are in use.

joshblack:
> Could you please articular what are you looking for when you say this? Thanks!

The way I understand it, we have ~2 options for doing icons in a way that can be DCE'd, namely:

- Import and use the icon data/information and supply it to a static Icon component
- Use the SVG source and generate components per-Icon

Was just curious about the former approach and its benefits/drawbacks. The linked article goes in detail in the second method, I believe.

When thinking about how folks might want to use import/use the icon, the best API is definitely the one that we have right now, e.g. `<Icon name="add" />`, but like you said it introduces a sense of dynamic behavior that makes it impossible to DCE the result.

We can achieve a similar API by doing that first approach with this PR, like with `<Icon icon={iconAdd} />`, but this requires importing both `carbon-components-react` and `carbon-icons`. 

To simplify this a bit, we could offer the icon exports directly from `carbon-components-react`, and have `carbon-icons` as a build-time dependency. One could then import icons like `import { AddIcon } from 'carbon-components-react'`, or `import AddIcon from 'carbon-components-react/icons/Add'`, etc. This has the added advantage of eliminating a dependency for a consumer to concern themselves with, and allows us to re-use the work we did to setup the react repo to properly DCE itself as a big module in newer versions of webpack.

asudoh:
Thanks @joshblack for articulating this!

> The linked article goes in detail in the second method, I believe.

Could you please put a link here? Thanks!

joshblack:
Of course 👍 http://nicolasgallagher.com/making-svg-icon-libraries-for-react-apps/

marijohannessen:
@asudoh can you fix the conflicts in this PR?

asudoh:
@marijohannessen Thank you for the ping! Did the rebase, re-targetted to `master`, and removed the breaking change as we have shipped `v9` already.

PR #889, opened 2018-05-15
feat(OverflowMenu): added onClose & onOpen to props to notify caller Menu's open state

kiett:

Current OverflowMenu doesn't notify parent component it's closed.

#### Changelog

**New**

this.props.onClose


kiett:
@asudoh can you please review again?  I moved the call to componentDidUpdate, so it should address the issue.  Thanks

kiett:
sure

kiett:
@asudoh added onOpen per your request, please kindly review again.  Thanks!

asudoh:
Hi @kiett thank you for your update - One point is that the `wrappedMenuBody` stuff is *not* fully placed in DOM as soon as it's rendered. So the `onOpen()` call should happen [when the menu is placed](https://github.com/carbon-design-system/carbon-components-react/blob/dddd7f6/src/components/OverflowMenu/OverflowMenu.js#L295). Thanks!

asudoh:
@kiett My apologies if it was not clear, but here's where I'd like `onOpen()` should happen: https://github.com/carbon-design-system/carbon-components-react/blob/dddd7f6/src/components/OverflowMenu/OverflowMenu.js#L295

Don't hesitate to ping me if you have further questions. Thanks!

kiett:
@asudoh,  Thanks for a quick review.  Don't you think placing onClose & onOpen call in the componentDidUpdate where the **component already updated** is the most sure place?  I know you mentioned put the onOpen in _handlePlace, but I think put it in componentDidUpdate where the menu is already either opened or closed, is the best bet.

asudoh:
Good question @kiett - The reason why I brought up the specific place is that it’s where the menu is fully placed (positioned) and `cDU` is too early AFAICT - Feel free to ping me if you have further questions - Thanks!

kiett:
@asudoh from official React site https://reactjs.org/docs/react-component.html#componentdidupdate:

> componentDidUpdate() is invoked immediately after updating occurs. This method is not called for the initial render.

cDU is called right after updating occurs, so in the floatingmenu, the Menu must have been shown already.  Unless you have other idea.  Therefore, I believe cDU is the best place for these hooks.  I'm fine to place onOpen in _handlePlace, if you think otherwise!

asudoh:
Thanks @kiett for thinking this through! You are right that `cDU` is called right after updating occurs, but floating menu goes through multiple state updates until the menu is fully shown, and more importantly, placed into the correct place. That’s why I brought up the line of https://github.com/carbon-design-system/carbon-components-react/blob/dddd7f6/src/components/OverflowMenu/OverflowMenu.js#L295. Don’t hesitate to speak up if you need more clarification - Thanks!

kiett:
Thanks @asudoh!  Another point I forgot to mentioned,  _handlePlace get called only when **floatingMenu** is true https://github.com/carbon-design-system/carbon-components-react/blob/dddd7f6/src/components/OverflowMenu/OverflowMenu.js#L376.  So, if we place hooker call in there, we may miss the case floatingMenu = false.  Thanks!

asudoh:
Thanks @kiett for bringing this up - Great point! Given this, `onOpen()` should be called in `cDU` like what you have for non-floating mode, and should be called at https://github.com/carbon-design-system/carbon-components-react/blob/dddd7f6/src/components/OverflowMenu/OverflowMenu.js#L295 for floating mode.

BTW - It appears that your recent rebase seems to remove recent changes in `master`. It'd be great if you can double-check - Thanks!

kiett:
@asudoh:  I feel it's redundant to call in both places.  Again, I want to refer back to react site https://reactjs.org/docs/react-component.html#componentdidupdate

> componentDidUpdate() is invoked immediately after updating occurs.

If this.state.open turns to true, then react will perform updating the component.  Once done, it will call componentDidUpdate, at this time the Menu **must** be fully shown already.  Assuming, cDU get called and Menu is still performing updating as you mentioned earlier, then there is something wrong in the react logic don't you think!?  In this case, this.state.open may not have the true correct value.

asudoh:
@kiett Let me make it clearer - In floating mode, `state.open` in `<OverflowMenu>` is *not* the only state we should take care of. We should also monitor `state.floatingPosition` in `<FloatingMenu> to see if the menu is fully placed. Hope this helps removing the confusion in our conversation here. Don't hesitate to ping me if you have any further questions. If you're an IBMer, we can talk over our internal Slack, too. Thanks!

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -257,14 +276,13 @@ export default class OverflowMenu extends Component {
...
-    if (!this._menuBody || !this._menuBody.contains(evt.target)) {
-      this.closeMenu();
-    }
+  handleClickOutside = () => {
+    this.closeMenu();
   };
 
   closeMenu = () => {
     this.setState({ open: false });
+    this.props.onClose();
```
joshblack: Could we move this into the callback position for the state updater? e.g.

```js
closeMenu = () => {
  this.setState({ open: false }, () => {
    this.props.onClose();
  });
}
```

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -293,6 +308,7 @@ export default class OverflowMenu extends Component {
   _handlePlace = menuBody => {
     if (menuBody) {
       this._menuBody = menuBody;
+      this.props.onOpen();
```
joshblack: Could we invoke this after the work has been done to open the menu?

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -257,14 +276,13 @@ export default class OverflowMenu extends Component {
     }
   };
 
-  handleClickOutside = evt => {
-    if (!this._menuBody || !this._menuBody.contains(evt.target)) {
-      this.closeMenu();
-    }
+  handleClickOutside = () => {
+    this.closeMenu();
```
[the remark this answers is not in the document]
asudoh: Seems that there was an error in rebasing - The `if()` block was introduced very recently.

asudoh:
Thanks @kiett for making the updates!

kiett:
@joshblack made changes based on your request.  Please review again.  Thanks!

kiett:
@alisonjoseph I resolved the merge conflict.  When will the merge restart?  Thanks!

PR #900, opened 2018-05-18
feat(DataTable): Update table header to support node as proptype

kkomprakash:
Closes carbon-design-system/carbon-components-react#878

Adds support to DataTable Header for allowing `header` as type `node` instead of `string`. This helps us to have elements like Tooltip in the DataTable Header

#### Changelog

**Changed**

- `DataTable->headers->header` prop changed to `PropTypes.node` from `PropTypes.string`
- Converted the header element from `button` to `div` to fix the mouseover event of Tooltips in the firefox browser


tw15egan:
Removing the `button` elements is causing issues with keyboard accessibility, will need to add in extra role's and tabIndex to ensure users are still able to tab through the headers

tw15egan:
It seems like the tabbing a11y issues have been fixed, but now I'm noticing I'm not able to toggle the sorting by keypress. I am able to do this in the current version: [see here](http://react.carbondesignsystem.com/?selectedKind=DataTable&selectedStory=with%20sorting&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

joshblack:
Going to close this for now, feel free to re-open if you still want to take this on!

PR #902, opened 2018-05-18
feat(data-table-v2): add i18n support to batch actions

joshblack:
Closes carbon-design-system/carbon-components-react#780

Add i18n support to Table Batch Actions

#### Changelog

**New**

**Changed**

* `TableBatchActions`
  * `translateWithId` now offers a `state` value to interpolate values properly

**Removed**


joshblack:
Note before anyone merges this in, this might break folks given the addition of the id for cancel. Might be worth introducing in v9.

PR #909, opened 2018-05-21
chore(package): update to React v16.4

joshblack:
Updates the project to use React v16.3, including new features like `forwardRef` and `StrictMode`, in addition to updating the usage of deprecated lifecycle hooks by prefixing them with `UNSAFE_`.

Most likely would be a candidate for v9, or a release post-v9.

### DIFF
--- a/.storybook/Container.js
+++ b/.storybook/Container.js
@@ -7,16 +7,18 @@ export default class Container extends Component {
     const { story } = this.props;
 
     return (
-      <div
-        role="main"
-        style={{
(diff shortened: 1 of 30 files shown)


asudoh:
@joshblack My apologies not noticing the update sooner - Seems that the PR needs remove some irrelevant commits - This may have happened due to the re-targetting. Also, should we go with React 16.4? Thanks!

joshblack:
Updated @asudoh!

PR #912, opened 2018-05-22
feat: Add light version to all input types

alisonjoseph:
Ref https://github.ibm.com/carbon/issues/issues/736

Add light prop to the following components

- Date Picker
- Number Input
- Search
- Select
- Textarea
- Text Input
- Time Picker
- Data Table Search (updated to light)

### DIFF
--- a/.storybook/_container.scss
+++ b/.storybook/_container.scss
@@ -1,4 +1,4 @@
 $css--font-face: true;
 $css--reset: true;
 
-@import 'node_modules/carbon-components/src/globals/scss/styles';
+@import '~carbon-components/scss/globals/scss/styles.scss';


.storybook/_container.scss
```diff
@@ -1,4 +1,4 @@
 $css--font-face: true;
 $css--reset: true;
 
-@import '~carbon-components/scss/globals/scss/styles.scss';
+@import 'node_modules/carbon-components/src/globals/scss/styles';
```
joshblack: Is this something that we'll need to change after PR reviews or is this a permanent change?

alisonjoseph: Oh, good catch, I'll swap it back before we merge.

src/components/Select/Select.js
```diff
@@ -47,6 +44,11 @@ const Select = ({
           className="bx--select__arrow"
           description={iconDescription}
         />
+        {!inline ? (
```
asudoh: Is moving the `<label>` here intentional?

alisonjoseph: Yes, the html was updated in the vanilla repo, so this is just matching the changes to the v9 component there.

PR #922, opened 2018-05-24
feat:(DatePickerInput) — add defaultDate option

zerostatic:
Closes carbon-design-system/carbon-components-react#905

Added `defaultDate` in props so that we can preload dates.

#### Changelog

**New**

Added `defaultDate` in props


tw15egan:
Agree with @asudoh, this seems like it can already be done like so: `<DatePickerInput value="07/25/2018" />`

zerostatic:
Ah yes, I overlooked that. value works. I removed and just added value to proptypes so it shows up in the Storybook documentation. What are your thought's on that? Or is there another way to add it to the docs?

tw15egan:
Perfect. Thanks for adding that documentation in!

PR #929, opened 2018-05-25
732 search

aledavila:
## Overview

Resolves 
https://github.ibm.com/carbon/issues/issues/732

Deprecated search with icons.

### Removed

Removed search with icons

### Changed

Changed close button offset to default 12px because Search with icons is deprecated.


src/components/Search/Search.js
```diff
@@ -124,13 +114,6 @@ export default class Search extends Component {
           aria-label={closeButtonLabelText}>
           <Icon name="close--glyph" description="close" />
         </button>
-        {children}
```
asudoh: Did you mean removing support for `children` altogether...?

PR #935, opened 2018-05-29
[WIP] feat(tooltip): add Definition and Icon variants

joshblack:
Adds in the new behavior for the Tooltip component, in addition to supporting the new Definition and Icon variants for v9.

#### Changelog

**New**

* `TooltipDefinition`
* `TooltipIcon`

**Changed**

* `Tooltip`
  * now only expands tooltip when the node is interacted with

**Removed**


joshblack:
cc @asudoh went through and did the tooltip changes quickly, so I think I messed up the refactor for the click. Wanted to ping you just to get another eye on this 👍

alisonjoseph:
Looks good 👍a couple styling things I noticed

Tooltip is missing the caret
screen shot 2018-05-31 at 3 20 03 pm

Definition tooltip custom direction position is off
screen shot 2018-05-31 at 3 19 37 pm

joshblack:
> Definition tooltip custom direction position is off

Yeah, was wondering if this is an issue upstream for styles?

alisonjoseph:
I think there may be html changes.

joshblack:
@alisonjoseph for definition tooltip I just copied and pasted from what's in v9: https://github.com/carbon-design-system/carbon-components/blob/v9/src/components/tooltip/tooltip--definition.hbs

And am just swapping out the direction class name. Is there something else?

joshblack:
Going to use this PR just to introduce the new Tooltip variants. The tooltip code seems like it'll need more attention to update without regressions

joshblack:
@alisonjoseph just verified it's working as intended in v9 upstream, will debug locally in this PR and see what's going on 👍

joshblack:
@alisonjoseph after digging it, everything is identical so I think it's related to usage around `position: absolute` here given that the parent nodes aren't using `position: relative` to capture a new layout context. Seems like something that needs to go in upstream as a fix

alisonjoseph:
tagging @aledavila to take a look since she worked on the tooltip styles

alisonjoseph:
We are going to have to use javascript to position the definition tooltip `top` unless it is limited to a single line. We do need to add in a fix that has position relative to the container though so that `bottom` will work correctly in all places.

alisonjoseph:
I created an issue in `carbon-components` to track what needs to get done. https://github.com/carbon-design-system/carbon-components/issues/842

joshblack:
@alisonjoseph could we just do `transform: translateY(-100%)` for that dynamic case? Can investigate in the vanilla repo, for sure 👍

Issue #933, opened 2018-05-29
ExpandableTile: Chevron a11y description


ZachStowellIBM:
## Detailed description

The description for the chevron is listed as "Tile chevron" on the ExpandableTile component.   For improved accessibility it would be beneficial to have the description be "Expand" when the tile is collapsed, and "Collapse" when the tile is expanded.

This issue appears in all browsers, and is currently reproducible on the master branch.

## Steps to reproduce the issue

1.  Navigate to the [storybook](http://react.carbondesignsystem.com/?selectedKind=Tile&selectedStory=Expandable&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) for the ExpandableTile component.

2.  When the tile is collapsed, hover over the chevron icon.  The tooltip/description displays "Tile chevron"
3.  Click on the chevron icon to expand the panel.
4.  Hover over the chevron icon again.  The tooltip/description displays "Tile chevron".

## Additional information
Screenshots: 
<img width="380" alt="collapsed-tooltip" src="https://user-images.githubusercontent.com/39738800/40674883-0cab8784-6344-11e8-9dba-11f054f602f6.png">

<img width="337" alt="expanded-tooltip" src="https://user-images.githubusercontent.com/39738800/40674826-ea429732-6343-11e8-9901-ad6c5661e096.png">

PR #936, opened 2018-05-29
fix(ExpandableTile): improve the description for the chevron (#933)

ZachStowellIBM:
Closes carbon-design-system/carbon-components-react#933

This PR will update the chevron inside of an ExpandableTile component to display a description of "Expand" when the tile is collapsed, and "Collapse" when the tile is expanded.

#### Changelog

**Changed**

* ExpandableTile chevron Icon now changes description based on the `expanded` state of the component.

### DIFF
--- a/src/components/Tile/Tile.js
+++ b/src/components/Tile/Tile.js
@@ -225,6 +225,8 @@ export class ExpandableTile extends Component {
     className: PropTypes.string,
     expanded: PropTypes.bool,
     tabIndex: PropTypes.number,
+    tileCollapsedIconText: PropTypes.string,
+    tileExpandedIconText: PropTypes.string,
   };
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/components/Tile/Tile.js
```diff
@@ -326,7 +326,10 @@ export class ExpandableTile extends Component {
         onClick={this.handleClick}
         tabIndex={tabIndex}>
         <button className="bx--tile__chevron">
-          <Icon name="chevron--down" description="Tile chevron" />
+          <Icon
+            name="chevron--down"
+            description={this.state.expanded ? 'Collapse' : 'Expand'}
```
tw15egan: Taking this one step further, could we pass down something like `tileCollapseIconText` and `tileExpandIconText` as a prop, with 'Collapse' and 'Expand' as the defaults? Just thinking about future use-cases with regards to internationalization

ZachStowellIBM: Good call.  I'll update to allow for that change.

tw15egan: Thanks!

PR #938, opened 2018-05-30
chore(search): remove icons from search

aledavila:

#### Changelog

**Removed**

Removed icons from default search. Is now normal and small no icons. Left support for custom.

### DIFF
--- a/src/components/Search/Search.js
+++ b/src/components/Search/Search.js
@@ -88,8 +88,6 @@ export default class Search extends Component {
       'bx--search-close--hidden': !hasContent,
     });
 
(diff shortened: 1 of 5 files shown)


src/components/Search/Search.js
```diff
@@ -94,7 +88,7 @@ export default class Search extends Component {
       'bx--search-close--hidden': !hasContent,
     });
 
-    const renderButtons = !children && !small;
+    // const renderButtons = !children && !small;
```
asudoh: Let’s remove this line.

aledavila: done @asudoh

PR #941, opened 2018-05-31
Enhancement: Allow React elements in DataTable.TableHeader

AndySeymour2904:
Altered Prop~Types for DataTable.TableHeader to prevent warnings when a user supplies a React Element as the child of the TableHeader, currently it only allows a string and hence no custom styling


tw15egan:
Related: https://github.com/carbon-design-system/carbon-components-react/pull/900

asudoh:
Wondering - Do we want to do `PropTypes.node`?

AndySeymour2904:
I'll update to PropTypes.node

Issue #945, opened 2018-06-01
FileUploader - support setting the name attribute of an input element


nhayashida:
It is helpful if we can set the `name` attribute of the input element generated by the FileUploader component as a prop value.

## Detailed description

> Describe in detail the issue you're having.

As some server side components for handling an uploaded file like 
[multer](https://github.com/expressjs/multer) use the name attribute of an input element, client code can be simple if the FileUpload component supports setting the attribute like `<FileUploader name="fieldName" />.`

> Is this issue related to a specific component?

FileUploader

emyarod:
This issue should be resolved by #974 in v6.4.0. Forgot to tag it properly in my PR

PR #974, opened 2018-06-06
feat(FileUploader): Add optional `name` attribute for FileUploader

emyarod:
Closes carbon-design-system/carbon-components-react #945 

This PR will add an optional `name` attribute to the FileUploader and FileUploaderButton so that the input element will receive its value. If no value is provided, then the attribute value defaults to an empty string.

#### Changelog

**New**

* add `name` attribute for `FileUploader` and `FileUploaderButton` components

### DIFF
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -13,6 +13,7 @@ export class FileUploaderButton extends Component {
     labelText: PropTypes.string,
     listFiles: PropTypes.bool,
     multiple: PropTypes.bool,
+    name: PropTypes.string,
     onChange: PropTypes.func,
     onClick: PropTypes.func,
     role: PropTypes.string,
@@ -65,6 +66,7 @@ export class FileUploaderButton extends Component {
       tabIndex,
       buttonKind,
       accept,
+      name,
       ...other
     } = this.props;
     const classes = classNames({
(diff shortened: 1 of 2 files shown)


src/components/FileUploader/FileUploader.js
```diff
@@ -96,6 +98,7 @@ export class FileUploaderButton extends Component {
           type="file"
           multiple={multiple}
           accept={accept}
+          name={name || ''}
```
asudoh: AFAIK React doesn’t set a property whose value is undefined. So I guess default empty string here is not needed?

emyarod: I guess the only reason why I left default as empty string is because I declared the propType to be of type String. Just a habit from using TypeScript and having it prevent the value from being undefined. I can remove the default value if you'd like

joshblack: For this case, might be worthwhile to have the empty string be set as the value in the `defaultProp` then 👍

emyarod: My latest commit removed these default values, but I will leave it to the team to decide whether or not you'd like to set a default name. It makes sense to me not to have a default name, and it doesn't seem like PropTypes will yell about it. I will defer to you guys on the final decision

joshblack: Since it's `name` for a form element, I think you're right that it shouldn't have a default value. For now, I'd just go with `name={name}` instead of giving it a default 👍 In the future we should make this prop type required, most likely.

emyarod: Sounds good, thanks for the input @asudoh and @joshblack. This branch is now ready for review

PR #978, opened 2018-06-07
fix(export): add tooltip exports

tw15egan:
Add in exports for the new `TooltipIcon` and `TooltipDefinition`


nicoxxie:
@asudoh TooltipSimple will be deprecated according to the notes

@tw15egan thanks for the quick turnaround!

tw15egan:
It will be deprecated, but not until the next release so I'll add the export for `TooltipSimple`

asudoh:
Thanks @tw15egan for the quick change!

PR #980, opened 2018-06-07
fix(Modal): Focus primary actionable button on modal opening

emyarod:
Closes carbon-design-system/carbon-components-react#827

This PR fixes the issue where the primary actionable element is not focused when a Modal component is opened. To accomplish this, `ref`s are attached to the Modal and Button components so that the primary actionable element can be programmatically referenced and focused.

#### Changelog

**Changed**

* Add support for `ref`s in both the Modal and Button components. This will allow for autofocusing of input elements on render.
* Convert `Button` from a stateless functional component to a class component in order to support `ref`s

There are 2 options to solve this issue, and I would like to gather some more opinions about this implementation:

One method would be to added the `autoFocus` attribute to the primary button or the `modalButton` in `Modal.js` depending on whether or not the modal is passive. This may be detrimental to a11y and usability though, since the linter will show a warning if `autoFocus` is used (as mentioned in the ticket).

Another method (currently implemented in this PR) is to convert the Button component from a stateless functional component to a class component, so that we can attach a ref and call the focus method when the Modal component is mounted.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -42,7 +42,7 @@ const Button = ({
       {...commonProps}
       disabled={disabled}
       type={type}
-      ref={other.inputRef}>
+      ref={other.inputref}>
       {children}
       {buttonImage}
     </button>
@@ -54,7 +54,7 @@ const Button = ({
       {...commonProps}
       href={href}
       role="button"
-      ref={other.inputRef}>
+      ref={other.inputref}>
       {children}
       {buttonImage}
     </a>
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -3,55 +3,81 @@ import React from 'react';
 import Icon from '../Icon';
 import classNames from 'classnames';
 
-const Button = ({
-  children,
-  className,
-  disabled,
-  small,
-  kind,
-  href,
-  tabIndex,
-  type,
-  icon,
-  iconDescription,
-  ...other
-}) => {
-  const buttonClasses = classNames(className, {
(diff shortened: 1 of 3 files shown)


emyarod:
The diff for this PR should now be greatly reduced. I've refactored my changes so that I am no longer converting `Button.js` from a stateless function to a class component. I am keeping the Button component as a stateless functional component while still maintaining a ref to the DOM to focus the button when it is visible

emyarod:
thank you for the tip @asudoh. Since I am now listening to the `onTransitionEnd` event, I no longer need to use component lifecycle methods or timeouts to call `focus()` on a button. Now I check the name of the property that is transitioning to avoid calling `focus()` an unnecessary number of times after the `onTransitionEnd` event fires. If the property name is `opacity` or `visibility`, then I know the modal is visible and the elements within it can be focused. Please let me know what you think about my latest changes

asudoh:
Hi @emyarod thank you for trying to deal with the animation stuff! Fiddling with the code revealed that the code is capturing some events bubbled from inner elements. Also, using event details for determining if the modal is visible (and if the button is focusable), especially not looking at what the value is transitioned *to*, sounds a bit brittle to our possible future changes to modal animation. So I'd recommend more direct check if the modal is actually visible or not, like what our vanilla code does. Thanks!

emyarod:
@asudoh thanks for your help and explanations. Please let me know what you think about the latest changes!

emyarod:
understood @asudoh, I've gone ahead and made that change!

PR #983, opened 2018-06-08
fix(OverflowMenu): close  open floatingMenu when  click on trigger in OverflowMenu

### LINKED ISSUE #931: OverflowMenu: floatingMenu while open, click on trigger opens it again.
## Detailed description

While using floatingMenu prop on OverflowMenu, clicking the trigger opens it, clicking the trigger while open, opens it again. 

Bug

OverflowMenu

When using the floatingMenu option, I would expect a click on the button while the menu is open to close it. The same behaviour as the non-floatingMenu.

Chrome 

5.51.5

Product not launched yet but would like to see a fix sooner rather than later

Go to [react.carbondesignsystem.com](http://react.carbondesignsystem.com/?selectedKind=OverflowMenu&selectedStory=floating%20menu&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).  

1.  Look at the OverflowMenu storybook example
2. "Basic" example will work when you click to open, and click the button again to close.
3.  "Floating menu" example and the click opens it, and the click opens it again rather than closing.

rezak-otmani:
Closes #931

We  close the menu after checking the no existence  of the css selectors

#### Changelog

The problem is also solved for Toolbar component

**Changed**

**Removed**


marijohannessen:
@asudoh can you review this again and see if your comments were addressed?

asudoh:
Hi @rezak-otmani my apologies for getting back late on this - Could you explain why you made a change in a way your latest code does? Thanks!

rezak-otmani:
Hi @asudoh, It's just the solution that solves the problem without touching the fucus, which is as you say a fundamental part of the component,
the solution is: close the menu after checking for the no existence of css selectors, not before this check

asudoh:
I see, thank you for articulating @rezak-otmani! Given `if (this.menuEl && !matches(target, '.bx--overflow-menu,.bx--overflow-menu-options'))` is the only thing in `if (!menuBody.contains(target))` as the result of your change, could you please merge those two? Thanks!

marijohannessen:
Are we okay with merging this now @asudoh ?

PR #1000, opened 2018-06-13
fix(datepicker): unknown prop on div tag

DmitriWolf:
Fixes [Issue 997](https://github.com/carbon-design-system/carbon-components-react/issues/997) and [Issue 998](https://github.com/carbon-design-system/carbon-components-react/issues/988).

## This is a Fix for v5
My project has not migrated to React 16.4 so we cannot use your v6. Please merge to v5.

## To reproduce with v5 and React 15.x:
- The console error does not occur in Storybook (maybe because that uses React 16) but you can still observe the props added to the `<div>` tag with the React dev-tools.
- Start with an app that is successfully using DatePicker component with version "^5.44.0",`
- Add a `locale` prop to DatePicker such as `locale: {'fr'}` *OR* a `minDate` or `maxDate` prop. All break v5 withe React 15.4.
- See error in console
- DatePicker will not initialize and will not respond to input.
- Console error will appear as in first screenshot
- Using React dev-tools to inspect the children of DatePicker will reveal that there is a plain `<div>` element that is being passed a `locale` prop, just as the error states.

### DIFF
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -373,6 +373,7 @@ export default class DatePicker extends Component {
 
   render() {
     const {
+      appendTo, // eslint-disable-line
       children,
       className,
       short,
@@ -381,6 +382,9 @@ export default class DatePicker extends Component {
       maxDate, // eslint-disable-line
       dateFormat, // eslint-disable-line
       onChange, // eslint-disable-line
+      locale, // eslint-disable-line
+      value, // eslint-disable-line
+      ...other
     } = this.props;
 
     const datePickerClasses = classNames('bx--date-picker', className, {
@@ -430,7 +434,7 @@ export default class DatePicker extends Component {
     });
     return (
       <div className="bx--form-item">
-        <div className={datePickerClasses}>
+        <div className={datePickerClasses} {...other}>
           {childrenWithProps}
           {datePickerIcon}
         </div>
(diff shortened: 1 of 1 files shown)


asudoh:
Hi @DmitriWolf thank you for jumping in to fix this issue! AFAICT the `{...other}` stuff is for giving consumer more control of attributes in `div.bx--date-picker`. Therefore the fix should be eliminating *only* the offending props. Also, if we are making a fix to `v5` such change should be done in `master`, too. Thanks!

DmitriWolf:
Thank you, @asudoh. 
I updated the code to leave `other`. Instead I assigned `const`s to all the rest of the props we expect (from looking at `propTypes`).

DmitriWolf:
@asudoh 
Thank you for the approval.
Please merge and manually make another v5 release so we can use the updated code. 
I will work on making these updates on the current `master` branch for v6 version.

asudoh:
Hi @DmitriWolf this project takes two reviewers’ approval before getting a PR merged. Also - Could you please make the same PR against `master`, too? Thanks!

DmitriWolf:
@asudoh I believe this does not require that as this is a merge into a Branch, not into master. 
On this PR I see this:
screen shot 2018-06-15 at 4 02 20 pm
While on the PR to merge into master I see this:
screen shot 2018-06-15 at 4 02 38 pm

asudoh:
Hi @DmitriWolf though this repo does not seem to have enabled “two reviewers” system rule (somehow), Carbon core takes “two reviewers” rule as the team. Thank you for your patience and understanding!

marijohannessen:
This PR is included in version 5.52.2

PR #1003, opened 2018-06-14
fix(DataTable): allow TableSelect components to be disabled

ryanomackey:
Closes carbon-design-system/carbon-components-react#973

#### Changelog

**Changed**

Allows a `disabled` prop to be passed down to the following:

- `<TableSelectAll />`
- `<TableSelectRow />`
- `<InlineCheckbox />`


ryanomackey:
@joshblack So, this is a pretty barebones approach. A more fleshed-out feature probably requires some additional discussion. 

I was toying around with the idea of allowing rows to have a `isDisabled` key, but that felt like it went against the spirit of the component.

Let me know what y'all think.

marijohannessen:
@joshblack can you chime in here?

PR #1011, opened 2018-06-18
fix(OverflowMenu): emulate vanilla component behavior

emyarod:
Closes carbon-design-system/carbon-components-react#991

Unlike its vanilla counterpart, the React basic OverflowMenu does not automatically focus the first menu item when the menu is opened. This PR will add a ref to the first menu item so that focus can be applied when the overflow menu is opened.

#### Changelog

**New**

* add `floatingMenu` prop to `OverflowMenuItem`s based on whether or not the `OverflowMenu` is a floating menu
* attach corresponding primary focus data attribute to menu item based on `primaryFocus` and `floatingMenu` props
* add conditional statement to focus primary element in basic `OverflowMenu` based on the new `data-overflow-menu-primary-focus ` prop

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -236,7 +236,7 @@ export default class OverflowMenu extends Component {
     if (this.state.open) {
       if (!floatingMenu) {
         (
-          this.menuEl.querySelector('[data-basic-menu-primary-focus]') ||
+          this.menuEl.querySelector('[data-overflow-menu-primary-focus]') ||
           this.menuEl
         ).focus();
         onOpen();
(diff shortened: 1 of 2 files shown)


emyarod:
Hi @asudoh is this more in line with what you are looking for? I am now using the existing `menuRef` and querying the DOM for the `data-floating-menu-primary-focus` attribute

emyarod:
@asudoh I think it would be better to change the attribute name only for the non-floating menu, to avoid potential confusion. In order to check whether or not the menu is floating, I added a `floating` prop to each menu item. I then check for `primaryFocus` && `floating` before adding the correct data attribute to the menu item. Let me know what you think of this implementation method

emyarod:
thanks for the suggestions @asudoh, I've gone ahead and made those changes to the data attribute and prop names

PR #1012, opened 2018-06-18
feat(InlineLoading): Create new component for InlineLoading

loganmccaul:
See issue here: https://github.com/carbon-design-system/carbon-components/issues/912

inlineloadingincontext

### Summary

The Inline Loading component is for any quick Create, Update, or Delete actions. It can work inline in a table, or in context of a primary button as pictured above.

### Props

Name  | type | description
------------- | ------------- | -------------
className | string | Additional classes that can be added to the parent for positioning
success  | boolean | Whether to show the loading or success "state"
description | string | The text to display
onSuccess  | function | A callback function for a post success action. This could be redirecting to a different page, closing a modal, or calling a refetch API function

### Why?

The Inline Loading component comes from a month of user research about loading patterns in the IBM Cloud Platform. We found that user's specifically liked having the description to help understand what action is taking place as well as the checkmark to have reassurance that the action succeeded.

Styles: https://github.com/carbon-design-system/carbon-components/pull/904

### DIFF
--- a/src/components/InlineLoading/InlineLoading.js
+++ b/src/components/InlineLoading/InlineLoading.js
@@ -9,14 +9,23 @@ export default class InlineLoading extends React.Component {
     success: PropTypes.bool,
     description: PropTypes.string,
     onSuccess: PropTypes.func,
+    successDelay: PropTypes.number,
   };
 
   static defaultProps = {
     success: false,
+    successDelay: 1500,
   };
 
   render() {
-    const { className, success, description, onSuccess, ...other } = this.props;
+    const {
+      className,
+      success,
+      description,
+      onSuccess,
+      successDelay,
+      ...other
+    } = this.props;
 
     const loadingClasses = classNames('bx--inline-loading', className);
 
@@ -26,7 +35,7 @@ export default class InlineLoading extends React.Component {
           if (onSuccess) {
             onSuccess();
           }
-        }, 1500);
+        }, successDelay);
 
         return (
           <svg
--- a/src/components/InlineLoading/InlineLoading.js
+++ b/src/components/InlineLoading/InlineLoading.js
@@ -1,0 +1,58 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import classNames from 'classnames';
+import Loading from '../Loading';
+
+export default class InlineLoading extends React.Component {
+  static propTypes = {
+    className: PropTypes.string,
+    success: PropTypes.bool,
+    description: PropTypes.string,
+    onSuccess: PropTypes.func,
+  };
+
+  static defaultProps = {
+    success: false,
+  };
+
+  render() {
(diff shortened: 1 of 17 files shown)


asudoh:
Thanks @loganmccaul for trying to contribute! CC @carbon-design-system/design just to ensure that the UX pattern has been discussed with the team. My specific interest is that if the UX pattern should be common beyond Cloud BU.

loganmccaul:
@asudoh Absolutely! We have been in sync with @tay-aitken throughout the process, but I'm happy to discuss with the larger team as well.

Right now, I would like to get it into a component library so we can test the API easier in context throughout platform. If that means the carbon-addons-cloud then I'm happy with that as well.

I can create a PR with the same component in carbon-addons-cloud and an Issue in the design board to discuss it's proper location.

tw15egan:
Thanks for adding this in @loganmccaul! A few things:

- If there is an error state, should we also add an error story for reference?
- I'm seeing an issue with the `success` icon, but this could also just be due to the styles not being set yet (see below) 

screen shot 2018-06-19 at 10 02 16 am

loganmccaul:
@asudoh Design issue created here: https://github.com/carbon-design-system/carbon-components/issues/912

loganmccaul:
@tw15egan The visual issue with success is because of the missing styles. I had to load both locally to get everything working.

There was originally a planned error state, but after communication with our design team and adding it into context we realized that error states are already handled within the form or by a notification. So instead the best action would be just revert back to the button if an error occurred. I'll update the storybook to remove Error

src/components/InlineLoading/InlineLoading.js
```diff
@@ -0,0 +1,58 @@
...
+
+    const loadingClasses = classNames('bx--inline-loading', className);
+
+    const getLoading = () => {
+      if (success) {
+        setTimeout(() => {
+          if (onSuccess) {
+            onSuccess();
+          }
+        }, 1500);
```
asudoh: Can we make this number as a prop?

loganmccaul: Yup! I wasn't sure if we would want to the success time to be consistent or not, but I'm happy for it to be a prop.

tay-aitken:
I was hoping to add this into our main library, unless you guys would prefer to put it in add-ons first then move it over eventually.

src/components/InlineLoading/InlineLoading.js
asudoh: Great - I think we can make `1500` the default. Thanks @loganmccaul!

marijohannessen: Did you want to make this a prop before merging @loganmccaul ?

loganmccaul: Oh, must've never pushed. My fault 😄

PR #1015, opened 2018-06-19
feat(DataTable): allow users to pass placeholder text into table tool…

### LINKED ISSUE #999: TableToolbarSearch: Cannot pass through placeHolderText to Search
## Detailed description

I want to change the placeholder text of the search bar in the DataTable so we can indicate to the user the functionality of this search (which in this case would be to filter tags)

Update the component with a placeHolderText prop just like the regular `Search` component 

DataTable > TableToolbarSearch

## Additional information

screen shot 2018-06-13 at 2 11 22 pm

screen shot 2018-06-13 at 2 06 01 pm

`<Search
        light
        className={className}
        {...rest}
        small
        id={id}
        labelText={t('carbon.table.toolbar.search.label')}
        placeHolderText={t('carbon.table.toolbar.search.placeholder')}
        onChange={onChange}
      />`

bakerTX:
…bar search

Closes #999 

This adds an optional prop for `TableToolbarSearch`, controlling the placeholder text of the search bar. If the prop is not present, then the placeholder text falls back to the previous implementation, which was relying on a translate function as such: `t('carbon.table.toolbar.search.placeholder')` (which I did not fully understand, but would be curious to know more about)

#### Changelog

**New**

* `placeHolderText` optional prop

**Changed**

* `Search` prop `placeHolderText` now points to a custom prop first

### DIFF
--- a/src/components/DataTable/TableToolbarSearch.js
+++ b/src/components/DataTable/TableToolbarSearch.js
@@ -5,22 +5,14 @@ import Search from '../Search';
 import setupGetInstanceId from './tools/instanceId';
 
 const getInstanceId = setupGetInstanceId();
-const translationKeys = {
-  'carbon.table.toolbar.search.label': 'Filter table',
-  'carbon.table.toolbar.search.placeholder': 'Search',
-};
-
-const translateWithId = id => {
-  return translationKeys[id];
-};
 
 const TableToolbarSearch = ({
   className,
   searchContainerClass,
   onChange,
-  translateWithId: t,
   id = `data-table-search-${getInstanceId()}`,
   placeHolderText,
+  labelText,
   ...rest
 }) => {
   const searchContainerClasses = cx(
@@ -35,10 +27,8 @@ const TableToolbarSearch = ({
         {...rest}
         small
         id={id}
-        labelText={t('carbon.table.toolbar.search.label')}
-        placeHolderText={
-          placeHolderText || t('carbon.table.toolbar.search.placeholder')
-        }
+        labelText={labelText || 'Filter table'}
+        placeHolderText={placeHolderText || 'Search'}
         onChange={onChange}
       />
     </div>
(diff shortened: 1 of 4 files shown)


joshblack:
Hi @bakerTX! The `translateWithId` prop is basically aimed at this (which is that `t` method you mentioned).

The idea is that you provide a function that takes in a message id, and then return the localized string that you want rendered for the user.

In the case of custom placeholder text, this would look like:

```
const translateWithId = id => {
  if (id === 'carbon.table.toolbar.search.placeholder') {
    return 'Placeholder';
  }

  if (id === 'carbon.table.toolbar.search.label') {
    return 'Filter table';
  }
};

<TableToolbarSearch translateWithId={translateWithId} />
```

Hope that helps!

bakerTX:
Thanks for the clarification, @joshblack!  I was able to get it working this way on my machine. However, it feels a little strange to be passing a simple string down this way. Is there a reason for this extra abstraction? I'm guessing just for the convenience of grouping multiple custom texts here (placeholder, label) But it feels a bit counter intuitive to me! 🤔 

Totally fine with leaving the current implementation in place, but let me know if you'd want a hand making these props a little more user-friendly 👍

marijohannessen:
What's the status on this @bakerTX @joshblack ?

bakerTX:
Looking through the code base, I was seeing a lot more usage of the `translateWithId` function than I initially expected. It might be appropriate to conduct some kind of multi-component audit to determine the actual necessity of this particular abstraction! 

However - regarding this specific component and issue, see my recent commits for a refactoring that removes the abstraction locally and addresses #999 at least.

joshblack:
> Looking through the code base, I was seeing a lot more usage of the translateWithId function than I initially expected. It might be appropriate to conduct some kind of multi-component audit to determine the actual necessity of this particular abstraction!

This method is used w hen we have a Component that renders a lot of children that also need to be translatable. While this is super frustrating from an isolated component perspective, this is very useful for the happy path of using the larger component directly and having that handle this pattern. 

Definitely think there is some happy path that we could provide that satisfies both use-cases, am open to suggestions! 😄

joshblack:
Hi @bakerTX! Going to close this due to inactivity, but feel free to re-open if you still want to tackle this!

PR #1019, opened 2018-06-21
Add invalid prop to Select

emyarod:
Closes carbon-design-system/carbon-components-react#1017

This PR will add an `invalid` prop to `Select` to apply data validation styles to the component. Depends on https://github.com/carbon-design-system/carbon-components/pull/919

#### Changelog

**New**

* Add `invalid` prop to `Select` component

**Changed**

* If boolean values are falsy, set value to `undefined` so that they are not attached as string `"false"` to the rendered DOM elements
* Fix tests and typos

**Removed**

* Removed conditional logic for determining location of `Select` label. Reflects changes in https://github.com/carbon-design-system/carbon-components/pull/919

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,7 @@
     "es/**/*",
     "umd/**/*"
   ],
-  "contributors": [
-    {
+  "contributors": [{
       "name": "Brian Han",
       "email": "bthan@us.ibm.com"
     },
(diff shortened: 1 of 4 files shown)


tw15egan:
Can you update the `package.json` to reflect the version of Carbon Components that contains the select updates when it is released?

emyarod:
good call @tw15egan I will take care of that when the next version is released

tw15egan:
Should be included in `v9.1.1`

emyarod:
@tw15egan https://github.com/carbon-design-system/carbon-components/issues/921 Just noticed another issue with those changes

emyarod:
updated the `package.json` to include the new required version of `carbon-components` after https://github.com/carbon-design-system/carbon-components/pull/926 was merged

PR #1022, opened 2018-06-22
fix(FileUploader): `onChange` method called twice

emyarod:
Closes carbon-design-system/carbon-components-react#998

The `onChange` method supplied to `FileUploader` is called twice by the component

#### Changelog

**Removed**

* Remove duplicate method call
* Remove redundant and unused props

### DIFF
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -210,7 +210,6 @@ export default class FileUploader extends Component {
         [...evt.target.files].map(file => file.name)
       ),
     });
-    this.props.onChange(evt);
   };
 
   handleClick = (evt, index) => {
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -175,7 +175,6 @@ export default class FileUploader extends Component {
     labelTitle: PropTypes.string,
     multiple: PropTypes.bool,
     name: PropTypes.string,
-    onChange: PropTypes.func,
     onClick: PropTypes.func,
     className: PropTypes.string,
     accept: PropTypes.arrayOf(PropTypes.string),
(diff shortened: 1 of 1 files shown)


emyarod:
@asudoh I've stopped the event from propagating and added the `onChange` method call back in the `FileUploader` change handler

asudoh:
Thanks @emyarod for the update! One question - Do you think we should keep attaching `onChange` event handler on [the top-level `<div>` in `<FileUploader>`](https://github.com/carbon-design-system/carbon-components-react/blob/cd1258b/src/components/FileUploader/FileUploader.js#L250)?

emyarod:
It seems like we only want `<FileUploaderButton>` to have access to the `onChange` handler, so I would like to remove it from the top-level `<div>`. It also looks like it's being [spread into the `<span>`s that contain `<Filename>`](https://github.com/carbon-design-system/carbon-components-react/blob/cd1258b/src/components/FileUploader/FileUploader.js#L270) which is probably not necessary as well.

I typically avoid destructuring and spreading `this.props` unless I am sure that all of the props are being used. I prefer being more explicit about which props are attached to which elements. It also looks like `className` prop is being spread into those elements, even though it is already being defined before that.

What are your thoughts?

asudoh:
@emyarod Thank you for putting more thoughts here! I agree with you that `onChange` should be removed from the top-level `<div>` in `<FileUploader>`.

IIUC `{…others}` pattern we see in the codebase is for allowing consumers better control on attributes in the key element, and I see changing the approach would largely affect consumers. But it’d be nice looking through our codebase and seeing if some properties should be swallowed by component code or some other elements. Thanks!

itdove:
Maybe something I didn't get but today I rebuild my app from travis and my package json is "^6.7.4" however I still see the issue. As it was from travis and no packages were installed I guess it took the latest one.
Should I do something special to get the real latest 6.7.4.... with the bug fix?

PR #1037, opened 2018-06-27
chore(package): update carbon-componets to 9.1.6

loganmccaul:
Bumps the version of carbon components to fix the Inline Loading component style in the storybook

### DIFF
--- a/.storybook/Container.js
+++ b/.storybook/Container.js
@@ -9,6 +9,7 @@ export default class Container extends Component {
     return (
       <React.StrictMode>
         <div
+          data-floating-menu-container
           role="main"
           style={{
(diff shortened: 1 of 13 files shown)


tw15egan:
Should the updated `yarn.lock` file also be included with this PR?

asudoh:
Seems that `yarn.lock` has more changes than the change in `package.json` dictates. One of the extra changes to `yarn.lock` is to `prettier`, and which seems to have caused Travis failure.

@loganmccaul Would you try picking `yarn.lock` from the latest `master` and run `yarn install` again? This should limit `yarn.lock` change. Thanks!

PR #1043, opened 2018-06-29
fix(overflow-menu): fix for long OverflowMenuItem text

tw15egan:
Adds in a `title` prop to `OverflowMenuItem` that will add the browser default tooltip on text that is expected to be truncated. I felt this was less intrusive than adding `title` on all `OverflowMenuItem` and can be added on a case by case basis. 

Requires https://github.com/carbon-design-system/carbon-components/pull/947 to be merged first

#### Changelog

**New**

- `title` prop that will add the browser default tooltip on the text that is expected to be truncated

**Testing**

- Check option 2 of the first `OverflowMenu` example once the PR is the main repo has been merged and pulled in


src/components/OverflowMenuItem/OverflowMenuItem.js
```diff
@@ -115,6 +117,11 @@ OverflowMenuItem.propTypes = {
    * `true` if this menu item belongs to a floating OverflowMenu
    */
   floatingMenu: PropTypes.bool,
+
+  /**
+   * `true` if this menu item has long text and requires a browser tooltip
+   */
+  title: PropTypes.bool,
```
asudoh: Just a nitpick - Wondered if we can change the prop name to `requireTitle` or something... I naïvely thought `title` goes directly to the tooltip when I saw the new prop name.

tw15egan: Yeah after thinking about it this weekend, I agree that it should be a bit more semantic. Will update

asudoh:
@tw15egan Would you fix the prettier thing...? Thanks alot!

PR #1060, opened 2018-07-09
feat(Checkbox): add `title` attribute to checkbox label

emyarod:
Closes carbon-design-system/carbon-components-react#1059

This PR adds a `title` attribute to the Checkbox component so that users can add alt text to their checkbox labels.

#### Changelog

**New**

* `title` attribute on `<Checkbox>`

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -10,6 +10,7 @@ const Checkbox = ({
   indeterminate,
   hideLabel,
   wrapperClassName,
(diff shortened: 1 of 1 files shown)


tw15egan:
https://github.com/carbon-design-system/carbon-components-react/issues/1059

emyarod:
updated the description @tw15egan thanks!

emyarod:
@alisonjoseph I've now added logic to set `title` to `null` if no value is passed in

PR #1062, opened 2018-07-09
Adopt tooltip

vidyakhadsareibm:

{{Tooltip needs to adopted by other component as a part of labelText to avoid CSS changes in the label alignment}}

#### Changelog

**Changed**

* NumberInput, Select, TextArea, TextInput prop types - Label Or LabelText changed from String to node.


tw15egan:
Closes https://github.com/carbon-design-system/carbon-components-react/issues/1035

Issue #1050, opened 2018-07-04
CodeSnippet single line component generates accessibility violation


natmasse:
## Detailed description

> Describe in detail the issue you're having.

Including a CodeSnippet (single line) component generates the following accessibility violation:
"A WAI-ARIA widget must have an accessible name (see message help text for details)"
for resolution see: 
https://aat.w3ibm.mybluemix.net/token/c61212f1-cc70-42bd-b57f-ad335cb02d77/3e48d0cf-c395-4356-998a-a40de772e4f3/latest///doc/w3/help/en-US/idhi_accessibility_check_g1156.html 

> Is this a feature request (new component, new icon), a bug, or a general issue?

bug

> Is this issue related to a specific component?

CodeSnippet - single line

> What did you expect to happen? What happened instead? What would you like to see changed?

No accessibility violation

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

carbon-components-react 6.7.4

<img width="1365" alt="screen shot 2018-07-04 at 5 53 59 pm" src="https://user-images.githubusercontent.com/40840777/42294961-aa832576-7fb3-11e8-9755-6a1bd2790800.png">

PR #1081, opened 2018-07-12
fix(code-snippet): add aria label to textbox

aledavila:
Closes IBM/carbon-components-react #1050 

Adds aria label to codesnippet text box  div. Cannot reproduce issue using DAP or AI but should fix it.

#### Changelog

**New**

* add aria label


tw15egan:
Can we allow users to set this through a prop (`ariaLabel`), and have a default string of 'code-snippet'? Seems to be how we do it on other components

aledavila:
Yea will do

Issue #1078, opened 2018-07-11
TooltipSimple broken not showing text on hover 


nicoxxie:
## Detailed description
![screen shot 2018-07-11 at 3 52 40 pm](https://user-images.githubusercontent.com/17915105/42598674-7c7c125e-8522-11e8-80ae-534151395403.png)

http://react.carbondesignsystem.com/?selectedKind=TooltipIcon&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general issue?
bug

> Is this issue related to a specific component?
TooltipSimple

> What did you expect to happen? What happened instead? What would you like to see changed?

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
V6.10.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cloud

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

PR #1083, opened 2018-07-13
fix(tooltipIcon): show text on hover

tw15egan:
Closes https://github.com/IBM/carbon-components-react/issues/1078

`TooltipIcon` uses `aria-label` to show the text. We were placing the text inside the `title` attribute.

#### Changelog

**Changed**

- Use `aria-label` instead of `title`


aledavila:
@asudoh do you want to approve

PR #1093, opened 2018-07-17
fix(FloatingMenu): use right .matches() function for different browsers

### LINKED ISSUE #1092: The overflow menu show an error on the browser console in IE
Use the latest IE browser.  Note, this works in Edge.  It has to be IE.

Go here:  http://react.carbondesignsystem.com/?selectedKind=OverflowMenu&selectedStory=floating%20menu&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Click on the overflow menu.  In IE you you will see the following error on the browser console.  This error doesn't appear to cause any problems.

image

asudoh:
Fixes #1092.

#### Changelog

**Changed**

* `Element.matches()` usage with the right API call for different browsers.


asudoh:
@joshblack You are right that the bug cannot be tested with our test setup which does not use actual browser.

Issue #1023, opened 2018-06-22
FileUploader differences from vanilla implementation


emyarod:
## Detailed description

while working on #1022 I noticed some divergence in the `FileUploader` from the vanilla components. Looking for some clarification on how to proceed

> Describe in detail the issue you're having.

When uploading files, the React `FileUploader` has some logic that will modify the label text. If the user is uploading multiple files, it will change the label text from `Add files` to the number of files currently being uploaded, and if the user is uploading a single file, the label text will be changed to the name of the file being uploaded. A couple things about the relevant code though:

1. it doesn't seem to be present in the vanilla components
2. it is unreachable in the current React component

should I remove it completely or is this a desirable feature extension from the vanilla implementation?

> Is this a feature request (new component, new icon), a bug, or a general issue?

general issue

> Is this issue related to a specific component?

`FileUploader`

> What did you expect to happen? What happened instead? What would you like to see changed?

Right now I am working around this issue while refactoring the component, but it might not even be worth keeping since it is different from the vanilla implementation. Would like some input on this matter though.

> What version of the Carbon Design System are you using?

6.6.3

## Steps to reproduce the issue

current Vanilla and React `FileUploader` behavior:

![kapture 2018-06-22 at 11 59 18](https://user-images.githubusercontent.com/8265238/41789269-4f258104-7614-11e8-9c3b-bda3c6cbc71d.gif)

logic present within React `FileUploader` but currently unreachable

![kapture 2018-06-22 at 12 02 57](https://user-images.githubusercontent.com/8265238/41789281-60ea944c-7614-11e8-937f-8afecc026f5d.gif)

asudoh:
Thanks @emyarod for reporting! I feel I’m kind of one the fence - The React one shows a nice UI wrt what is being uploaded, but loses a context of additional files can be uploaded. CC @carbon-design-system/design to see which one is more desirable.

emyarod:
hi @asudoh I am leaning towards the current behavior found in the vanilla component (not changing the text within the button), but I am fine with either option. I just want more clarification on the direction we want to proceed with as I fix the dead code issue.

One question though: is the `FileUploaderButton` ever used by itself (without the `FileUploader`) or is it always a child of a `FileUploader` component? I am assuming that `FileUploaderButton` should not be used by itself since the vanilla Carbon documentation does not show this usage pattern.

[...]

PR #992, opened 2018-06-12
chore(components): cWRP -> gDSFP migration for basic components

asudoh:
This PR makes a switch from `cWRP` -> `gDSFP` for basic components.

#### Changelog

**Changed**

* Switch from `cWRP` to `gDSFP`.

PR #1105, opened 2018-07-19
[WIP] feat(FileUploaderV2): add FileUploaderV2

emyarod:
Closes IBM/carbon-components-react#1023 

Closes IBM/carbon-components-react#1166

This PR will add a `FileUploaderV2` which shows individual file upload statuses rather than applying one status to all files in the upload queue. This will provide a better indication of overall status and upload progress. `FileUploaderV2` also avoids making use of deprecated React life cycle methods to handle updates. An example application using this component is included in the storybook for reference.

This PR will also fix existing errors logged by invalid custom prop names in `FileUploader` (V1), remove unused props, and refactor the `Filename` and `FileUploader` components to be a stateless function rather than a class component.

**TODO**

- [x] ~~deduplicate shared child components~~
- [x] ~~remove deprecated React life cycle methods from V1 FileUploader~~ addressed in #992 thanks to @asudoh 
- [ ] add an example application using the V1 FileUploader  to the storybook (?)
- [x] ~~mock `fetch` call in V2 FileUploader example app~~
- [x] ~~add knobs storybook addon to V2 story (related: #1260)~~

#### Changelog

**New**

* add `FileUploaderV2`
* add `FileUploaderItem`

**Changed**

* convert `Filename` to stateless functional component
* redesign component to be primarily display-only and move change handler and upload logic to parent application
* fix FileUploader behavior when `multiple` prop is falsy (fix only implemented in V2 component to prevent introducing breaking changes to V1) #1166

**Removed**

* remove unused props in `FileUploader` (V1)
* remove inline styles on loading SVG (https://github.com/IBM/carbon-components/pull/1016)
* remove invalid custom attribute names in `FileUploader` (V1) logging errors to the console


src/components/FileUploader/FileUploader-story.js
```diff
@@ -6,6 +6,38 @@ import FileUploader, { FileUploaderButton } from '../FileUploader';
 import FileUploaderSkeleton from '../FileUploader/FileUploader.Skeleton';
 import Button from '../Button';
 
+class App extends React.Component {
+  upload = ({ file }) =>
+    fetch('https://jsonplaceholder.typicode.com/posts/', {
```
asudoh: Rather than creating a service dependency here, we may want to consider something like [`fetch-mock`](https://www.npmjs.com/package/fetch-mock).

src/components/FileUploader/FileUploader.js
```diff
@@ -12,12 +12,10 @@ export class FileUploaderButton extends Component {
     className: PropTypes.string,
     disableLabelChanges: PropTypes.bool,
     id: PropTypes.string,
-    labelText: PropTypes.string,
-    listFiles: PropTypes.bool,
+    labeltext: PropTypes.string,
```
asudoh: The prop list/names changes look breaking change. To make it clear, we cannot make any breaking changes.

src/components/FileUploader/FileUploader.js
```diff
@@ -112,25 +107,14 @@ export class FileUploaderButton extends Component {
...
-    status: 'uploading',
-    style: {},
-    tabIndex: 0,
-  };
-
-  render() {
-    const { iconDescription, status, style, ...other } = this.props;
-
-    if (status === 'uploading') {
+export function Filename({
```
asudoh: We cannot make a breaking change tp `<Filename>` either - Can we make it non-breaking, or do you want to do `<FilenameV2>`?

src/components/FileUploader/FileUploader.js
emyarod: The name change was just for removing console errors. I can keep this change exclusive to V2 if that's preferable

src/components/FileUploader/FileUploader.js
emyarod: This change should be a non-breaking change. I am converting the class component to a stateless function. Have I mistakenly removed a feature that should be present?

src/components/FileUploader/FileUploader.js
```diff
@@ -13,11 +13,9 @@ export class FileUploaderButton extends Component {
     disableLabelChanges: PropTypes.bool,
     id: PropTypes.string,
     labelText: PropTypes.string,
-    listFiles: PropTypes.bool,
     multiple: PropTypes.bool,
     name: PropTypes.string,
     onChange: PropTypes.func,
-    onClick: PropTypes.func,
```
asudoh: Why did you remove the prop type for `onClick` here…?

emyarod: correct me if I'm wrong but I don't think this prop is being used at all since a click handler is already hard coded [here](https://github.com/IBM/carbon-components-react/pull/1105/files/8fb14765a5ce8307bf2b7edfab49ffdec075da9d#diff-b82c23ac01d092b1d134837688198215L106)

asudoh: @emyarod As far as I read the original code correctly, `onClick` of `<FileUploaderButton>` goes [here](https://github.com/IBM/carbon-components-react/blob/3677e8a/src/components/FileUploader/FileUploader.js#L94), but don’t hesitate to let me know if you think otherwise. Thanks!

emyarod: My mistake @asudoh you're right, I overlooked the spread props again. I've made the correction

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,285 @@
...
+                <span
+                  key={uid()}
+                  className="bx--file__selected-file"
+                  ref={node => (this.nodes[index] = node)}
+                  {...other}>
+                  <p className="bx--file-filename">{file.name}</p>
+                  <span className="bx--file__state-container">
+                    <FilenameV2
+                      iconDescription={iconDescription}
+                      status={file.status}
```
asudoh: Seems that `filenameStatus` is not used in `<FileUploaderV2>` but still I see `filenameStatus` in https://github.com/emyarod/carbon-components-react/blob/9576d6b/src/components/FileUploaderV2/FileUploaderV2-story.js#L68. This seems to be causing React’s unknown prop warning.

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,285 @@
...
+      files.push({
+        uuid,
+        name: file.name,
+        size: file.size,
+        status: 'uploading',
+      });
+      const index = files.findIndex(file => file.uuid === uuid);
+      this.props
+        .upload({ file })
+        .then(() => {
```
asudoh: The code here seems to assume `props.upload()` returns a `Promise`, but the corresponding `defaultProps` (https://github.com/emyarod/carbon-components-react/blob/c4a18be/src/components/FileUploaderV2/FileUploaderV2.js#L171) doesn’t? I don’t know if it’s the direct cause or not, but I see `Uncaught TypeError: Cannot read property 'then' of undefined` when I tried to upload a file in `<FileUploaderV2>`’s story.

emyarod: I've made the default `upload` prop value a function that returns an empty Promise to remove the console warning.

Is there a scenario where the `FileUploader` would be used standalone without providing an `upload` function? It might be worth completely removing the standalone `FileUploader` story and just leaving the example application story, if that would be a better illustration of how to use the component

src/components/FileUploaderV2/FileUploaderV2.js
emyarod: I've changed the `filenameStatus` and `labelText` props to all lowercase which should remove the console warnings

asudoh: Thanks @emyarod for addressing this! `filenameStatus` to `status` looks a good change. Could you please explain why the change from `labelText` to `labeltext` addresses the console warnings?

src/components/FileUploaderV2/FileUploaderV2.js
asudoh: Great point @emyarod! This leads me to a question - Could an application simply listen to `onChange`, upload the file, and set file uploading status via props once uploading is complete? In this way we can relegate the application-specific file uploading procedure to application code and let our file uploader component focus on visuals and user interactions.

src/components/FileUploaderV2/FileUploaderV2.js
emyarod: React wants to differentiate DOM-related props from [custom props](https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html), so they try to [keep DOM-related props in camelCase and custom props in lowercase](https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes)

This is the warning that is given if you pass a custom camelCase attribute

screen shot 2018-07-26 at 8 09 00 am

asudoh: Thank you for explaining @emyarod! This is caused by `labelText` going to `<label>`’s attribute in addition to `<label>`’s children, caused by `{…other}` thing. Basically, what you need to do is listing everything that goes other than `<label>`’s attribute in [here](https://github.com/emyarod/carbon-components-react/blob/67f26d3/src/components/FileUploaderV2/FileUploaderV2.js#L35-L42).

emyarod: got it @asudoh, I've reverted the prop name change and am now explicitly declaring the prop instead of spreading it like before

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,257 @@
...
+  style = {},
+  tabIndex = 0,
+  ...other
+}) {
+  switch (status) {
+    case 'uploading':
+      return (
+        <div
+          className="bx--loading"
+          style={{ ...style, width: '1rem', height: '1rem' }}
```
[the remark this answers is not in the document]
emyarod: would it be better to modify [these CSS rules](https://github.com/IBM/carbon-components/blob/master/src/components/file-uploader/_file-uploader.scss#L75-L76) in `_file-uploader.scss` rather than defining a new CSS class?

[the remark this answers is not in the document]
asudoh: Great question @emyarod! Sounds that it makes sense unless change in style there breaks vanilla.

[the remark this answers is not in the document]
emyarod: hi @asudoh, I've made a PR in the vanilla repo so that we can remove these inline styles https://github.com/IBM/carbon-components/pull/1016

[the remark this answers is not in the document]
asudoh: Hi @emyarod thank you for doing this! Seems that the PR contains a breaking change - Tapping designers to see if such breaking change can be done now. Thinking that introducing React-specific CSS class may be an option.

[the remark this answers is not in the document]
emyarod: the inline styles are removed now that the vanilla repo has been modified @asudoh

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,240 @@
...
+};
+FilenameV2.defaultProps = {
+  onKeyDown: () => {},
+  status: 'uploading',
+  style: {},
+  tabIndex: 0,
+};
+
+export default function FileUploaderV2(props) {
+  const handleClick = ({ evt, index }) => props.onClick({ evt, index });
```
asudoh: Seems that we can do `const handleClick = props.onClick`?

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,240 @@
...
+        multiple={multiple || undefined}
+        buttonKind={buttonKind}
+        onChange={onChange}
+        accept={accept}
+        name={name}
+      />
+      <div className="bx--file-container">
+        {files.length
+          ? files.map((file, index) => (
+              <span key={uid()} className="bx--file__selected-file" {...other}>
```
asudoh: Instead of re-generating `key` for every render (will cause React-level re-rendering), can we assign a unique ID for each `file`?

emyarod: @asudoh would setting `key={file.uuid}` be a solution to this issue?

src/components/FileUploaderV2/FileUploaderV2.js
emyarod: @asudoh I left that in case I needed to add some logic within the component itself I have now refactored it to directly call the `onClick` prop

src/components/FileUploaderV2/FileUploaderV2.js
shinytoyrobots: @asudoh - Do you have an answer to this question? Looks like one of the last things we need to do to resolve.

emyarod: @asudoh thank you for reviewing! and yes @shinytoyrobots as @asudoh mentioned, I will be wrapping up work on this component according to the TODO list in the description of this PR, which may change as needed

src/components/FileUploadStatus/FileUploadStatus.js
```diff
@@ -0,0 +1,85 @@
...
+  style,
+  tabIndex,
+  ...other
+}) {
+  switch (status) {
+    case 'uploading':
+      return (
+        <div
+          className="bx--loading"
+          style={{ ...style }}
```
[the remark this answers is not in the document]
emyarod: since it doesn't seem like `FileUploadStatus` receives any additional props, is it worth having `{...others}` in the `FileUploadStatus` component?

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,170 @@
...
+    onClick: () => {},
+    accept: [],
+  };
+  uid = this.props.id || uid();
+
+  render() {
+    const {
+      className,
+      multiple,
+      role,
```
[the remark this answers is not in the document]
emyarod: do you mean we no longer want to destructure these props @asudoh ?

src/components/FileUploader/FileUploader.js
```diff
@@ -170,67 +165,6 @@ export class FileUploaderButton extends Component {
   }
 }
 
-export class Filename extends Component {
```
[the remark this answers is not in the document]
emyarod: [is this what you meant?](https://github.com/IBM/carbon-components-react/pull/1105/commits/d809ba2ce0103571023169980e8b0c2c3e67418a)

src/components/FileUploaderV2/FileUploaderV2.js
```diff
@@ -0,0 +1,170 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import uid from '../../tools/uniqueId';
+import { ButtonTypes } from '../../prop-types/types';
+import FileUploadStatus from '../FileUploadStatus/FileUploadStatus';
+
+export class FileUploaderButtonV2 extends Component {
+  static propTypes = {
+    className: PropTypes.string,
```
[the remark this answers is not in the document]
emyarod: I also added some annotations to FileUploaderV1, since it looks like only FileUploaderButton was annotated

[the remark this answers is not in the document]
emyarod: @asudoh would it be worth considering moving this information to the documentation rather than within the codebase itself?

src/components/FileUploaderV2/FileUploaderV2.js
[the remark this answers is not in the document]
asudoh: @emyarod Thank you for asking - `role` and `tabIndex` seem to be set by `{...others}`, and thus we don't need to set those explicitly. Does it make sense? Don't hesitate to ping me otherwise!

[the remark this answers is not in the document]
emyarod: got it @asudoh , and just to make sure, is that what you meant in [this comment](https://github.com/IBM/carbon-components-react/pull/1105#discussion_r215134786) as well?

src/components/FileUploaderV2/FileUploaderV2.js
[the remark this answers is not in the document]
asudoh: @emyarod If you click on Show Info button at the top right, you’ll see the prop descriptions for some of the components. It comes from the comments from `propTypes`. Please see https://github.com/IBM/carbon-components-react/pull/1195 for more details, and don’t hesitate to ping me if you have any questions/concerns. Thanks!

src/components/FileUploaderV2/FileUploaderV2.js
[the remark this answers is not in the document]
asudoh: @emyarod Yes - I’d say any props that are covered by `{…other}` don’t have to be set explicitly (except `propTypes`).

src/components/FileUploadStatus/FileUploadStatus.js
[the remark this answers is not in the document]
asudoh: @emyarod Thank you for asking! `{…others}` serves two purposes. One is as a shortcut for propagating props to the primary React node. Another is giving consumers flexibility wrt control over the props on primary React node. Does it make sense? Don’t hesitate to ping me if you have any questions/concerns. Thanks!

src/components/FileUploader/FileUploader.js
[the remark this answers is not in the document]
emyarod: sorry [here](https://github.com/IBM/carbon-components-react/blob/d9b14658b1a4dd4825cec8589c266c2dea0e6f9a/src/components/FileUploader/FileUploader.js#L7-L9) is what I was referring to in the link @asudoh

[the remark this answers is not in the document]
asudoh: Great - My apologies for missing that @emyarod!

emyarod:
my point of view is that if current users are already using the V1 component, this story may not be beneficial to them, and if a new user wants to use `<FileUploader>`, then the V2 component would be a natural starting point. We could then separate this storybook example into a new ticket. What do you think @asudoh?

asudoh:
> my point of view is that if current users are already using the V1 component, this story may not be beneficial to them, and if a new user wants to use `<FileUploader>`, then the V2 component would be a natural starting point. We could then separate this storybook example into a new ticket.

I may have missed a discussion @emyarod - Did you have any discussion this PR triggering your thought above? Otherwise do you want to elaborate this further?

BTW another consideration for this PR is https://github.com/IBM/carbon-components-react/issues/1339, which I'm trying to get in place soon. Let me work with you on the integration once it happens - Thanks!

emyarod:
I haven't had any other discussion about this @asudoh but I was just under the impression that the V1 component would be on a deprecation path, so we would want to funnel users towards V2

and thanks for the heads up on #1339! Looking forward to integrating feature flags into this component

PR #1108, opened 2018-07-19
Added polyfill for Array.findIndex()

zerostatic:
The MultiSelect component uses method: findIndex() This requires a polyfill  for older browsers (IE11 etc...)
#### Changelog

**New**

added `import 'core-js/modules/es6.array.find-index';` to .storybook/polyfills.js

giphy

PS, great library!


joshblack:
Thanks for the catch! We might want to actually try and revert the usage of `findIndex` to something that is IE11-compat, instead of bringing on another dependency. What do you think?

zerostatic:
Sounds good. I changed findIndex() to use a a combination of map and indexOf. Let me know if that works or if you know of a better solution. Thanks!

src/components/MultiSelect/MultiSelect.js
```diff
@@ -194,18 +194,17 @@ export default class MultiSelect extends React.Component {
...
+                      var itemIndex = selectedItem
+                        .map(function(selected) {
+                          return selected.id;
+                        })
+                        .indexOf(item.id);
                       const isChecked =
-                        selectedItem.findIndex(
-                          selected => item.id === selected.id
-                        ) !== -1;
+                        typeof selectedItem[itemIndex] !== 'undefined' || false;
```
asudoh: Thank you for your quick change @zerostatic! Given you mimicked `findIndex`, do you want to keep `!== -1` comparison?

zerostatic:
Ah, good point! I changed to use `!== -1` Much cleaner.

zerostatic:
Nice catches...Sorry, old habits!  These have been updated.

Issue #874, opened 2018-05-10
The ModalWrapper doesn't have a keyboard trap


zgrossbart:
1. Go here:  http://react.carbondesignsystem.com/?selectedKind=ModalWrapper&selectedStory=transactional%20modal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

2. Click the *Transactional Modal* button to open the dialog.

3. Press the tab key.  At this point the focus will be on the close button in the upper right.

4. Press the tab key twice.  At this point the focus will be on the Save button.

5. Now press the tab again.  The focus will leave the dialog and end up on the *Show Info* button.

If you keep pressing the tab key the focus will leave the modal and end up on different items in the page.  For accessibility it should stay within the modal dialog until it is dismissed.

zgrossbart:
There is also a Jira issue to track this problem.  cloudlab-jira.canlab.ibm.com:8080/browse/UI-1510

asudoh:
Thank you for repoting @zgrossbart! Having contributions porting [`carbon-components/src/components/modal/modal.js#L140-L153`](https://github.com/carbon-design-system/carbon-components/blob/3e9bc26/src/components/modal/modal.js#L140-L153) to this React repo would be nice.

aledavila:
Closing fixed in above.

PR #1115, opened 2018-07-20
fix(Modal): add keyboard trap

Kyle-Cooley:
Adds a keyboard trap to fix issue #874


Kyle-Cooley:
Hello @asudoh. That was a good catch. I have amended my commit to check if the modal is open when trapping focus.

asudoh:
Hi @Kyle-Cooley thank you for your quick change! Just a couple of additional things:

1. Vanilla version moves the focus (back) to modal itself, instead of preventing the focus from moving. Can we align the behaviors?
2. Vanilla version makes focus to floating menu an exception, due to https://github.com/IBM/carbon-components/issues/280. Can we address this in this React variant, too?

FYI here's the corresponding code in vanilla: https://github.com/IBM/carbon-components/blob/3e9bc26/src/components/modal/modal.js#L140-L153

Kyle-Cooley:
I addressed the first, but I'm struggling with the second. I'm using an overflow menu (floating) in a modal. The items in the menu can be tabbed to, but when the overflow menu closes, focus is not returned to the modal.

Kyle-Cooley:
I think I've managed to address the issue, let me know if there is something else.

src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -323,7 +323,11 @@ export default class OverflowMenu extends Component {
   };
 
   closeMenu = () => {
+    let wasOpen = this.state.open;
     this.setState({ open: false }, () => {
+      if (wasOpen) {
+        this.focusMenuEl();
```
asudoh: Did you want to remove: https://github.com/Kyle-Cooley/carbon-components-react/blob/742a24bb891752975242ccc388f49f22515c0948/src/components/OverflowMenu/OverflowMenu.js#L383-L386

src/components/Modal/Modal.js
```diff
@@ -36,10 +43,21 @@ export default class Modal extends Component {
     iconDescription: 'close the modal',
     modalHeading: '',
     modalLabel: '',
+    selectorsFloatingMenus: ['.bx--overflow-menu-options__btn'],
```
asudoh: Do you think it's possible to align the list with vanilla? https://github.com/IBM/carbon-components/blob/3e9bc26/src/components/modal/modal.js#L194

joshblack:
Hi @Kyle-Cooley! Looks great, mind fixing the merge conflicts and then we can merge?

Kyle-Cooley:
Should be good now.

PR #1119, opened 2018-07-22
fix(OverflowMenu): support decimal point calculating the size of pointng arrow

asudoh:
Refs IBM/carbon-components#979.

#### Changelog

**New**

* Support for decimal support calculating the size of pointng arrow of overflow menu.


asudoh:
@joshblack Good point - Seems that we need to wait for https://github.com/speedskater/babel-plugin-rewire/issues/209 to get that function tested.

Issue #1123, opened 2018-07-24
Cannot override default focus behavior on modal open


emyarod:
## Detailed description

> Describe in detail the issue you're having.

Originally reported by [Garrett Darley](https://ibm-cloudplatform.slack.com/archives/C2K6RFJ1G/p1532015414000217). The modal does not allow the user to focus on other elements on modal open. Instead, it auto focuses on the primary actionable element within the modal itself no matter what.

PR #1124, opened 2018-07-24
fix(Modal): add support for primary focus selector prop

emyarod:
Closes IBM/carbon-components-react#1123

This PR will allow users to provide a selector that will be focused when the modal is opened. This behavior will be in line with the vanilla component's capabilities.

#### Changelog

**New**

* add `selectorPrimaryFocus` prop on `<Modal>`
* include `<ModalWrapper>` Storybook example showing `selectorPrimaryFocus` prop behavior

**Changed**

* modify auto focus behavior based on `selectorPrimaryFocus` value, or use default primary actionable item


IOIO72:
I suggest, that the data attribute should also be possible to set in other carbon components like ComboBox. They contain text fields, which should also be `focusable` in modal views.

PR #1125, opened 2018-07-24
feat: sets a default value for date picker on mount

ricardo-jsx:
This feature lets the user set a default value to date-picker on the mount of the component.


ricardo-jsx:
Hi @asudoh, I didn't notice this. Removed, thanks ;)

PR #1132, opened 2018-07-27
fix(Icon): add prop type check for either icon or name prop exists

asudoh:
Follow-up fix of #1129.

CC @asfordmatt 

#### Changelog

**New**

* New prop type check in `<Icon>`, requiring either `icon` or `name` exists.
* An explicit error in addition to above, instead of swallowing the error in such condition.


asudoh:
@asfordmatt Let’s get back to the question I asked over Slack - Why did you want `<Icon />` (with neither `icon` or `name`)? In other words, why did you want that not to cause an error?

To address your question, an error will be thrown suggesting that your `<Icon />` misses either `icon` or `name` prop, instead of an internal error like you saw `Cannot read property 'split' of undefined`.

asfordmatt:
@asudoh If throwing an error is your preferred approach then go ahead - at least it's a helpful one instead of the internal one. We didn't find out how an Icon came to have an undefined name. So when this goes in it may break us, but we can revisit our investigations and it won't be so urgent as we can always pin on version 6.14.3 whilst we do that investigation.

src/components/DataTable/TableToolbarAction.js
```diff
@@ -2,6 +2,7 @@ import cx from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import Icon from '../Icon';
+import isRequiredThisOrOneOf from '../../internal/isRequiredThisOrOneOf';
```
joshblack: Want to put this utility in our `prop-types` sub-folder?

asudoh:
@joshblack Thank you for asking. This is *not* `oneOfType`, where a prop takes one of two prop types. This is instead about one of two props are required (for `<Icon>`, it’s `name` or `icon`). Does it make sense? Don’t hesitate to ping me otherwise - Thanks!

src/components/DataTable/TableToolbarAction.js
asudoh: @joshblack OK done!

joshblack:
@asudoh can kind of get it, I think the name of it itself is throwing me off. Basically it's saying that this prop has a proptype and has a dependency/requirement on another prop?

asudoh:
@joshblack My apologies for my wordings. [A StackOverflow question of a same topic](https://stackoverflow.com/questions/42284076/at-least-one-required-prop-in-react) expresses it as “at least one required prop” - Wondering if it makes more sense to you?

joshblack:
Nice! That SO answer was really helpful. Seems like, at the end of the day, this is just XOR yeah?

One comment on the usability would be that it'd be nice if we didn't have to manually define dependencies between props that have relationships with other props (like in `['name']`). It'd be nice if the helper followed a pattern like:

```js
type XORPropTypeChecker<T, K extends keyof T> = ({
  [key: K]: PropTypeChecker // like PropTypes.string
}) => ({
  [key: K]: NewPropTypeChecker // Would be your custom logic wrapped around PropTypeChecker
});
```

In other words, it's a function that takes in an object where the keys are the prop names and they're original prop type. It returns an object of keys with the wrapped prop type that you implemented already.

In practice, this could look like:

```js
static propTypes = {
  ...onlyOnePropIsRequired({
    icon: PropTypes.object,
    iconName: PropTypes.string,
  }),
};
```

Which becomes:

```js
static propTypes = {
  icon: OurCustomPropType, // Wraps `PropTypes.object`
  iconName: OurCustomPropType, // Wraps `PropTypes.string`
};
```

The wrapper here is what handles validating the relationship between prop types since only of the keys is required in the given object.

asudoh:
Makes sense - Thanks @joshblack for your suggestion! Made a change.

PR #1134, opened 2018-07-27
fix:  Moves the label before the select list for better accessibility…

### LINKED ISSUE #1131: Select: Accessibility failure due to label
## Detailed description

When using a Select component, the `<label>` element is being rendered after the `<select>` element, causing the following accessibility error: `Text inputs and select elements must have a label before the input control.`

bug

Select

Chrome

6.14.3

IBM Cloud

zgrossbart:
…, resolves #1131.

This is a small fix.  We are placing the label before the select list with CSS anyway.  We just need to tweak the DOM so we can pass accessibility tests for screen readers.


joshblack:
Hi @zgrossbart! 👋 Thanks for taking the time to make this PR. Just curious, what a11y violation occurs when having the input and label swapped? Was under the assumption that, given that the `id` was valid, that the ordering wouldn't matter.

zgrossbart:
This error came from one of our testers.  Let me get more information.

zgrossbart:
@joshblack We have posted more information about the a11y issue to the issue report here:  #1131

alisonjoseph:
I'm worried this may be a breaking change, the label was moved after the select with v9 style updates. @joshblack do you know why? 

I also do not believe this should be causing a violation. (although I see where DAP says it is 😕) The order of the form element and the label shouldn't make a difference as long as the correct `id` and `for` are used. 

https://www.w3.org/WAI/tutorials/forms/labels/

joshblack:
@alisonjoseph have been chatting in the accessibility-at-ibm channel and it seems like this criteria comes up in their testing spec due to:

> This is documented in the w3 test techniques for H44: https://www.w3.org/TR/WCAG-TECHS/H44.html
> 
> For all input elements of type text, file or password, for all textareas and for all select elements in the Web page:
> * Check that there is a label element that identifies the purpose of the control *before* the input, textarea, or select element
> * Check that the for attribute of the label element matches the id of the input, textarea, or select element
> * Check that the label element is visible.

alisonjoseph:
This should be good to go. I tested that it doesn't cause any styling issues.

Issue #363, opened 2017-11-21
Usage without ejecting from Create React App


Alex-Mann:
The carbon design system and kit is an awesome library of components. I just wanted to inquire if there was any discussion (internal or external) about supporting this in a non-ejected version of create-react-app.

I realize there would have to be some sort of migration or tooling to go from `.scss` files to using `.css`, but I feel like this would really benefit the community, especially newcomers to react who want to jump into just using React without having to worry about all the tooling involved in the ecosystem.

Maybe `styled-components` is a potential migration option?

joshblack:
Hi @Alex-Mann! 👋  Thanks so much for taking the time to make this issue 😄

We definitely would want to support CRA without having to eject, unfortunately the requirement on sass is something that caused the current example to eject. Since the Carbon Design System's styles need to be library agnostic, targets like `styled-components`, `glamor`, etc. are unfortunately not on the immediate roadmap as they are heavily tied to React (in some capacity).

Would love to hear if you have any ideas/suggestions that could help match our requirements yet still meet your expectations from Carbon when looking to use it in things like CRA 😄

Alex-Mann:
So these are just some very preliminary thoughts and might not be completely applicable, but I'll just throw it out there in case it spurs some thought from anyone else reading this issue.

[This](https://medium.com/styled-components/getting-sassy-with-sass-styled-theme-9a375cfb78e8) was an interesting article I came across regarding extracting `sass` styles and putting them into styled components.

With your current situation, the best approach is probably some kind of generation/build pipeline to make components out of your sass styles, similar to what has been done [here](https://github.com/kossnocorp/decss).

I will do some more thinking about this as well and update this thread as those thoughts come about.

joshblack:
Love all your ideas @Alex-Mann, and definitely feel like our team should try and push on being able to define styles in one language (preferably JS) and then target CSS/SCSS/etc. Would help increase safety of the styles themselves as we can enforce/codify certain standards, and makes our compilations cleaner 😄

psachs21:
Another possibility for css in js frameworks in [emotion](https://emotion.sh/) which actually doesn't tie itself to react but supports much the same syntax (with styled components like support provided via an optional babel plugin). I was looking to start using it on our project (within watson finance) and would love to synchronize with this project, if possible.

joshblack:
[...]

PR #1135, opened 2018-07-27
Add create-react-app example without ejecting

vpicone:
Closes IBM/carbon-components-react#363

Uses react-app-rewired to circumvent the need for ejecting create-react-app or forking react-scripts. The webpack override allowing for scss imports has been included.

#### Changelog

**New**

* Add example

### DIFF
--- a/src/components/DropdownV2/DropdownV2-test.js
+++ b/src/components/DropdownV2/DropdownV2-test.js
@@ -73,7 +73,7 @@ describe('DropdownV2', () => {
       );
 
       expect(wrapper.find('span.bx--list-box__label').text()).toEqual(
-        mockProps.items[0].label
+        mockProps.items[0].text
       );
     });
 


vpicone:
Could someone help me figure out why this build failed? Should it not have updated the yarn offline directory?

joshblack:
Hi @vpicone! 👋 Thanks so much for taking the time to make this PR 😄 

With respect to the build here, I believe you'll just need to run `yarn prettier` in order to format some of the files you added.

At a high level, for a non-eject CRA could we simply add a reference to our styles from the unpkg CDN in `public/index.html`? This is how our codesandbox is currently setup. Just another suggestions 👍

vpicone:
@joshblack Yeah I'm uncertain as to why the original example ejected, if sass isn't a requisite to using the library it seems weird to add sass functionality it arbitrarily. Simply adding the cdn like you mentioned or importing the minified css from carbon-components npm package would do just fine. It seems like some of the sass-specific features aren't needed in the majority of usecases. 

I'm experimenting and learning about carbon components in general so I'm probably missing something haha.

joshblack:
The original CRA example was ejected just because some teams wanted to do one of the following:

- Include only the component styles that they were using
- Use some of the primitives that we use in our sass in their own sass files

I don't think this is a majority use-case for teams picking up CRA just to try out Carbon, but we'll definitely still want to provide some production-specific use-case just in case.

Issue #1130, opened 2018-07-26
MultiSelect: Enhance to support invalid and invalidText props


scottdickerson:
## Detailed description
The Select and TextInput fields show invalid and invalidText beautifully, this enhancement request is to allow the MultiSelect to also support these invalid states.  Here's an example of Select
![image](https://user-images.githubusercontent.com/6663002/43283048-72a0e888-90dd-11e8-845a-a9c9e7d5c511.png)

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Feature Request

> Is this issue related to a specific component?

MultiSelect

> What did you expect to happen? What happened instead? What would you like to see changed?

MultiSelect feature parity with Select

> What browser are you working in?

> What version of the Carbon Design System are you using?

9.0.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

joshblack:
Hi @scottdickerson! 👋 Definitely a great suggestion for an enhancement. Will definitely accept PRs that help make this happen!

scottdickerson:
@joshblack I'm working it today and will submit soon.

PR #1139, opened 2018-07-30
feat(MultiSelect): Add invalid and invalidText props to show validation errors

scottdickerson:
Closes IBM/carbon-components-react#1130

Adds two new props to MultiSelect to show validation errors like the other input widgets.  Also requires this PR to be incorporated into carbon-components to handle the styling: https://github.com/IBM/carbon-components/pull/1004

#### Changelog

**New**

* Add(MultiSelect) added two new props invalid, and invalidText to show error messages if validation errors occurred.


scottdickerson:
if we get the official new published version # of carbon-components would be good to update the peer dependencies to point to it.

PR #1160, opened 2018-08-03
fix: Update stories to reflect required label property

vpicone:

The storybooks don't make it clear that the required property for defaultItemToString to function is 'label' not 'text'. This is especially confusing since storybook hides the custom itemToString supplied in the stories as itemToString={itemToString()}.

#### Changelog

**Changed**

* Switched item property from 'text' to 'label' to prevent confusion when relying on the defaultItemToString method.

### DIFF
--- a/src/components/ComboBox/ComboBox-story.js
+++ b/src/components/ComboBox/ComboBox-story.js
@@ -6,19 +6,19 @@ import WithState from '../../tools/withState';
 const items = [
   {
     id: 'option-1',
-    text: 'Option 1',
+    label: 'Option 1',
   },
   {
     id: 'option-2',
-    text: 'Option 2',
+    label: 'Option 2',
   },
   {
     id: 'option-3',
-    text: 'Option 3',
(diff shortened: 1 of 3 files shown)


joshblack:
Hi @vpicone! 👋 This is actually intentional to show usage of `itemToString`. Definitely understand how this is confusing. Do you have any other suggestions for how to show this without changing all instances of using `itemToString`?

vpicone:
@joshblack I think at its core this issue might be that this older version of Storybook isn't parsing the stories and their tables properly. Making the default itemToString function visible in the table would be ideal. I can totally see how this is an effective way of showing the flexibility of this function, but it's impossible to know what to put in the items prop just going off the story and its opaque functions. It seems to upgrade, a slight overhaul of the Storybook implementation would need to take place since the addWithInfo method is being deprecated. 

I believe the older version of storybook is also the cause of the double brackets in arrays of object props, this was fixed in a patch I believe and wouldn't need a major upgrade.

Double Brackets

PR #1168, opened 2018-08-08
feat(DataTable): introduce batch editing

asudoh:
This PR introduces several components based on Carbon `<DataTable>`’s sub-components, that achieves full table editing.

The usage of `<DataTable>` here is using it as a fully presentational component, relegating state management to application logic. Such approach stems from this fact; Given handling table, especially with batch editing, involves lots of application logic managing its data and managing states in `<DataTable>` itself causes mixed presentation logic and application logic and thus more complex codebase.

Here are the steps for quick testing (or you can use any approach to test this PR along with the styles from https://github.com/IBM/carbon-components/pull/1020):

```shell
> cd /path/to/carbon-components
> git fetch git@github.com:asudoh/carbon-components.git table-batch-edit
> git checkout table-batch-edit
> gulp sass:source
> cd /path/to/carbon-components-react
> git fetch git@github.com:asudoh/carbon-components-react.git table-batch-edit
> git checkout table-batch-edit
> yarn install
> cd node_modules/cabron-components/scss/components/data-table-v2
> cp /path/to/carbon-components/scss/components/data-table-v2/* .
> yarn storybook
```

#### Changelog

**New**

* `<InvalidIcon>`, the form validation UI specific to full table edit (Only for below components, and thus *not* exported)
* `<TableBatchAction>`, a button that best suites in table batch edit action bar (Basically a ghost button)
* `<TableBatchEditable>`, which supports graying out the table while the content is being saved
* `<TableBatchEditableContainer>`, which supports other editing style of the table
* `<TableBatchEditActions>`, the table toolbar content for editing, with Save/Cancel buttons
* `<TableBatchEditDeleteCell>`, a table cell tailored for putting delete icon in
* `<TableBatchEditDeleteHeader>` a table cell header for above
* `<TableBatchEditNumberInput>`, a number input to be used in full table edit
* `<TableBatchEditTextInput>`, a text input to be used in full table edit
* Example application-level components that renders table from application data, and manipulates application data

**Changed**

* Prop type of `invalidText` in form components, from `string` to `node` to make them more flexible
* `<TableBatchAction>` to make it more flexible, by replacing hard-coded props to `defaultProps`


joshblack:
Thanks for moving this over @asudoh! Will definitely take a look at this.

asudoh:
@joshblack While it’s not straightforward to create a staging link for this PR for now due to style dependency, there is a internal staging link that can demonstrate the UI. I’ll message you internally.

joshblack:
One quick note, would we want to separate out stories for DataTable similar to how we had it in this PR: https://github.com/IBM/carbon-components-react/tree/962e07c2b108bafd02c321b201976c7cf0e69d9e/src/components/DataTable/stories

Might help for this story given the setup it needs for Redux. Could be done in a separate PR, as well.

asudoh:
Makes sense @joshblack! Will do once yours is merged.

joshblack:
Going through this a bit more, would it be helpful to separate out the `invalidText` contributions to components in another PR? Also, just PR'd that one for the stories that I was talking about over in #1189.

src/components/DataTable/TableBatchEditable.js
```diff
@@ -0,0 +1,37 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import classnames from 'classnames';
+import Table from './Table';
+
+/**
+ * `<Table>` supporting batch editing UI.
+ */
+const TableBatchEditable = ({ children, className, saving, ...other }) => {
```
joshblack: Would this read better as `BatchEditableTable` potentially?

joshblack:
One note on prop naming for consistency, would be nice if our state booleans followed a convention of `isSaving`, `isValid`, etc. Also down for other conventions, just want to make sure we consolidate on one 👍

asudoh:
Hi @joshblack thank you for reviewing and putting your thoughts on this! Naming changes are good and I'll go ahead with them unless anybody disagrees. As you imagined, I relegated the form validation stuff to consumer apps (Redux code in the Storybook example) to provide flexibility to consumers and to keep our codebase simple.

joshblack:
> As you imagined, I relegated the form validation stuff to consumer apps (Redux code in the Storybook example) 

I don't mean that we handle the validation stuff, more so that we give the controls needed to satisfy the different validation steps. This is why I brought up formik, as the API provided will let you configure how things are validated or what to do when they are submitted, but it doesn't actual handle/write these steps for you.

Definitely agreed that we don't want to do everything for them, but I do feel like the components themselves need to be aware of the different states of validation or submission which is why I suggested that API direction.

asudoh:
Thank you for putting it more thoughts @joshblack! While I’m inclined more toward the flexibility with the approach of simply letting Redux store define the form validation state, the centralized form validator approach you suggested is something we may want to think about in future.

alisonjoseph:
I'm having trouble getting this PR to run locally, getting a bunch of errors. 😕 

screen shot 2018-08-16 at 10 09 06 am

joshblack:
>  While I’m inclined more toward the flexibility with the approach of simply letting Redux store define the form validation state

I might be mis-explaining myself here, all I meant by my comment is that our component needs to be aware of this state for lines like:

https://github.com/IBM/carbon-components-react/blob/a356326a89fdec3562e8a01ec46fbc510917ee19/src/components/DataTable/TableBatchEditActions.js#L26-L29

Meaning that certain components we publish need to be aware of the following states:

- When a field is editable
- When a field is being edited
- When a field is being validated
- When a field has passed validation
- When a field has failed validation
- When a field is being saved
- When a field has been saved
- When a field has failed being saved

I think the field perspective here is more tailored towards inline editing, but I think in batch editing it can also stand true with respect to the table, so that the above becomes:

- When a table is editable
- When a table is being edited
- When a field in a table is being edited
- When a field has passed validation
- When a field has failed validation
- When a table is being saved
- When a table is saved
- When a table has failed being saved

If we have components that are aware of this, and we want to help orchestrate the coordination of these states (but not perform the actions themselves for validating/saving) then we should provide an API that surfaces this underlying state. This is why I brought up the idea of formik, since it surfaces things like `initialValues`, `validate` hook for validating the "form" (which would be our table), an `onSubmit` handler for submission, etc.

If we take our approach to `DataTable` and apply the `BatchEdit` model onto this, it could become another state container like:

```jsx
<BatchEdit
  onValidate={myCustomValidate}
  onSubmit={myCustomSubmit}
  rows={myCustomRows}
  initialValues={myCustomInitialEditableCellValues}>
  {({
    getEditProps,
    ...rest,
  }) => (
    <DataTable {...rest}>
      {({ rows, headers, getCellprops }) => (
        {/* other data table stuff */}
        {rows.map(row => (
          <TableRow {...getRowProps({ row })}>
            {cells.map(cell => (
              <TableCell {...getCellProps({ cell })} {...getEditProps({ cell })} />
            )}
          </TableRow>
        )}
      )}
    </DataTable>
  )}
</BatchEdit>
```

Just wanted to bring this up and see what everyone thought about it 👍

asudoh:
@alisonjoseph This PR introduces some new dependencies. `react-redux`/`redux-logger` are ones of them. Could you try re-running `yarn install`? Thanks!

joshblack:
I think it might be helpful to take this to a planning issue and try and work backwards from what we think the desired API would be here for consumers. There are also some great QOL changes in here that I think could be pulled out into smaller PRs that we can get in quickly while we chat about what we're hoping to expose here. What do you all think?

asudoh:
@joshblack If you take a close look at this PR, all new components that are introduced are simple presentational components that project props to user interface. No change to existing data table code while it’s used for batch editing story, too. If you want to engineer more, you can write ones on top of them. BTW don’t hesitate to ping me if it’s not clear enough on how to use the new batch editing API - Thanks!

src/components/DataTable/TableBatchEditable.js
asudoh: OK made a change as such!

shinytoyrobots:
Feels like we're at an impasse here in terms of the discussion. What are next steps?

Can we break up the QOL pieces that Josh suggested and feature flag where necessary?

asudoh:
@shinytoyrobots Agreed that the QOL stuffs should be tracked in a separated issue - Changes in this PR doesn’t prevent that QOL improvement in any way.

joshblack:
Going to close this due to inactivity. We should make sure to revisit this post-X so we should definitely keep the branch around 👍

Issue #1166, opened 2018-08-07
FileUploader allows choosing multiple files with `multiple={false}`


k-nut:
The `FileUploader` has a prop called `multiple`. If this is set to `false` (the default), the users cannot select multiple files to be uploaded at once.

It does allow choosing one file, closing the browser popup and the clicking the button again to upload another file though. I would assume that after one file is selected the button changes to a disabled state until the user clears the file that they previously selected.

Am I misunderstanding how this component should behave and it behaves as planned indeed? If so, maybe the prop should be renamed to `multipleAtOnce` or something like that?

[CodeSandbox](https://codesandbox.io/s/q8m5o1o4q9)

asudoh:
Hi @k-nut thank you for reporting! I think `multiple` being `false` should cause selected file being replaced when user selects another file in a separate pop-up. CC’ing @emyarod who is working on the next version of file uploader.

emyarod:
This issue should be resolved with [this commit](https://github.com/IBM/carbon-components-react/pull/1105/commits/778c2b20dfeadb1e945fe3c3f41d0a23120cea40) in the V2 component after [this PR](https://github.com/IBM/carbon-components-react/pull/1105) is merged in

PR #1172, opened 2018-08-08
fix(FileUploader): limit `multiple="false"` to a single upload

emyarod:
This PR will limit `FileUploader` to a single upload when the `multiple` prop is `false`. Closes IBM/carbon-components-react#1166

#### Changelog

**Changed**

* behavior of `onChange` handler in `FileUploader` when `multiple` prop is false

### DIFF
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -207,11 +207,12 @@ export default class FileUploader extends Component {
   }
   handleChange = evt => {
     evt.stopPropagation();
-    this.setState({
-      filenames: this.state.filenames.concat(
-        [...evt.target.files].map(file => file.name)
-      ),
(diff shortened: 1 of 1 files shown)


asudoh:
Thank you for doing this @emyarod!! Given this is a breaking change, wondering if we can do this in your V2…?

emyarod:
yes @asudoh I can make this change for V2 only, sorry about the misunderstanding

asudoh:
My apologies for not being clear and thank you for understanding @emyarod! 🙏

emyarod:
@asudoh this should be resolved with [this commit](https://github.com/IBM/carbon-components-react/pull/1105/commits/778c2b20dfeadb1e945fe3c3f41d0a23120cea40) in the V2 branch

PR #1177, opened 2018-08-10
 fix(MultiSelect): IE11 and menu closing fix

zerostatic:
Two issues fixed - I don't believe these are in the GitHub issues

For: **MultiSelect & FilterableMultiSelect**
1. In IE 11, the check boxes were not selecting properly. 

2. The follow these steps created the menu to close:
• Open menu
• Select an item
• Close Menu
• Open Menu
• Select another item -> Here it used to close the menu when you selected an item, This is what was fixed - it no longer closes the menu

#### Changelog

**New**

I added this to the Listbox click event to fix:
```
event.preventDefault();
event.stopPropagation()
```

**Changed**
Added `false` to this line: `let nextIsOpen = changes.isOpen || false;`
Because `isOpen` was sometimes returning undefined, and this was causing the menu to close when selecting checkboxes after following the steps in #2 above

giphy


joshblack:
Looks great! Mind updating the snapshot tests? 🙏

zerostatic:
My tests are passing locally now...not sure what the error is in Travis. Any tips to resolve this?

joshblack:
Just need to run `yarn prettier` I think on the files! 👍

PR #1189, opened 2018-08-14
refactor(data-table): update story structure

joshblack:
Updates the story structure for DataTable. cc @asudoh 

#### Changelog

**New**

**Changed**

* `DataTable` now has a `stories` directory where all the stories have been separated into

**Removed**


joshblack:
@asudoh honestly just convenience rather than doing static imports and figuring out a way to name all the stories 👍

PR #1190, opened 2018-08-14
Add prune config for Yarn in the offline mirror

joshblack:
Add prune config for Yarn in the offline mirror. Also updates our Yarn version in TravisCI

#### Changelog

**New**

**Changed**

* `.yarnrc` now has prune option enabled
* `.travis.yml` now uses Yarn v1.9.4

**Removed**


asudoh:
@joshblack Do you want to possibly update the offline mirror, or other stuffs that are needed for Travis to go through? Thanks!

PR #1194, opened 2018-08-15
fix(DropdownV2): improves dropdown keys to not rely only on their tex…

bakerTX:
…t values

This simply changes the key prop for DropdownV2 entries to be a little more robust, to not rely only on their text values but also on a unique index. This is in response to the following bug when I had 2 entries with the same primary text but they were, in fact, different objects.
screen shot 2018-08-14 at 7 44 43 pm

#### Changelog

**Changed**

ListBox.MenuItem `key` prop from:

`key={itemToString(item)}` 

to

```key={`dropdown-item-${itemToString(item)}-${index}`}```


joshblack:
Agreed with what @asudoh was saying above, we kind of need the unique ids since the order of items may not be guaranteed. Unfortunately, in the worst case we could end up with multiple of the same item just with different indices appended to the `id`.

I think this is a great proposal though for making it easier to work with this component. By default, the keys should work well enough. The assumption with `itemToString` as the `key` was that a consumer would not want two of the same option visible to the user in the list of options.

joshblack:
@bakerTX the id bit might be misleading here. The core issue, I think, is that `itemToString` is used to render options in the list. The assumption is that items in the list should be unique, otherwise consumers see duplicate of the same options. As a result, we use the output of that function as the `key` since if there are duplicate keys, then there duplicate options rendered and that would violate that initial assumption we had.

joshblack:
Let's go ahead and close this PR for now. Would love to continue this discussion over in an issue! That way when it comes to PR'ing any changes we know what to expect and can merge it in ASAP 😄

Issue #326, opened 2017-11-10
Update Storybook addWithInfo due to dep


eddiemonge:
The addWithInfo addon is deprecated. It has new usage using withInfo

```
@storybook/addon-info .addWithInfo() addon is deprecated, use withInfo() from the same package instead. 
See https://github.com/storybooks/storybook/tree/master/addons/info
```

PR #1195, opened 2018-08-15
fix(storybook): Add docgen babel plugin to allow for proptype/comment parsing in prop tables

vpicone:
This adds a dependency on the docgen plugin that parses components for improved prop tables in Storybook. The version provided by the info addon doesn't work with multiple named imports and this dependancy circumvents it. This helps with the opaque defaults as well as the missing requirements, Proptypes, and descriptions in the prop tables. Given that the storybook effectively serves as the documentation, this could go a long way towards making the library more accessible to new-comers.

**Before:**

Before

**After:**

After

#### Changelog

**New**

* babel-plugin-react-docgen

**Changed**

* added babel-plugin-react-docgen to parse Proptypes and documentation for storybook tables.


vpicone:
I think migrating to the withInfo syntax would also make defining modifying these tables' settings globally a little bit easier.  #326

Issue #1199, opened 2018-08-15
FileUploader storybook examples failing


emyarod:
## Detailed description

> Describe in detail the issue you're having.

while looking at examples to add storybook `addons-knobs` configurations to https://github.com/IBM/carbon-components-react/pull/1105 I found that the FileUploader storybook examples are failing or logging errors

> Is this a feature request (new component, new icon), a bug, or a general issue?

storybook issue

> Is this issue related to a specific component?

`<FileUploader>`

> What did you expect to happen? What happened instead? What would you like to see changed?

The 

> What browser are you working in?

Chrome 68+, Firefox 63+, Safari 11

> What version of the Carbon Design System are you using?

v6.20.0

## Steps to reproduce the issue

[storybook](http://react.carbondesignsystem.com/?knob-Label%20text%20%28labelText%29=Add%20files&knob-Accepted%20file%20extensions%20%28accept%29%5B0%5D=.jpg&knob-Accepted%20file%20extensions%20%28accept%29%5B1%5D=.png&knob-Supports%20multiple%20files%20%28multiple%29=true&knob-Status%20for%20file%20name%20%28filenameStatus%29=edit&knob-The%20title%20%28title%29=Section%201%20title&knob-ARIA%20role%20of%20the%20button%20%28role%29=&knob-The%20label%20description%20%28labelDescription%29=only%20.jpg%20files%20at%20500mb%20or%20less&knob-Form%20item%20name%3A%20%28name%29=&knob-Button%20kind%20%28buttonKind%29=none&knob-Tab%20index%20%28tabIndex%29=-1&knob-The%20label%20title%20%28labelTitle%29=Upload&knob-The%20button%20label%20%28buttonLabel%29=Add%20files&selectedKind=FileUploader&selectedStory=FileUploader&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

## Additional information

screenshots

`<FileUploaderButton>` story:

<img width="633" alt="screen shot 2018-08-15 at 9 19 16 am" src="https://user-images.githubusercontent.com/8265238/44153003-9755c31c-a06c-11e8-9021-2cbc54a4b116.png">

`<FileUploader>` story:

<img width="1440" alt="screen shot 2018-08-15 at 9 12 01 am" src="https://user-images.githubusercontent.com/8265238/44152871-39dc2adc-a06c-11e8-8fa2-81f5b736cc58.png">

PR #1200, opened 2018-08-15
docs(FileUploader): fix failing storybook example

emyarod:
Closes IBM/carbon-components-react#1199

#### Changelog

**Changed**

* modified `<FileUploader>` storybook examples to render and avoid logging errors to console

### DIFF
--- a/src/components/FileUploader/FileUploader-story.js
+++ b/src/components/FileUploader/FileUploader-story.js
@@ -17,9 +17,12 @@ import FileUploaderSkeleton from '../FileUploader/FileUploader.Skeleton';
 import Button from '../Button';
 
 const buttonKinds = {
-  none: 'None ()',
   primary: 'Primary (primary)',
   secondary: 'Secondary (secondary)',
+  danger: 'Danger (danger)',
+  ghost: 'Ghost (ghost)',
+  'danger--primary': 'Danger Primary (danger--primary)',
+  tertiary: 'Tertiary (tertiary)',
 };
 
 const filenameStatuses = {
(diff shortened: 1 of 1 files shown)


src/components/FileUploader/FileUploader-story.js
```diff
@@ -36,7 +39,7 @@ const props = {
       labelText: text('Label text (labelText)', 'Add files'),
       name: text('Form item name: (name)', ''),
       multiple: boolean('Supports multiple files (multiple)', true),
-      buttonKind: buttonKind === 'none' ? '' : buttonKind,
+      buttonKind: buttonKind || 'primary',
```
asudoh: Hi @emyarod any reason behind this change…? Thanks!

emyarod: hi @asudoh I wanted to consult you about this change but this is what I saw: I removed the 'none' option entirely since the value of `buttonKind` was defaulting to empty string [here](https://github.com/emyarod/carbon-components-react/blob/09b98f8fabcbb612c25dce086ef6cc56f5bbb230/src/components/FileUploader/FileUploader-story.js#L36) I believe. So it looked like the value of `buttonKind` was never `'none'`, and I changed it to default to `'primary'` if the value of `buttonKind` was falsy

asudoh: Thank you for the background @emyarod! My bad that I put an empty string there, which should be `none` aligned to the 1st selection my original code. I made it simply because empty string as an object key it not natural (though the spec seems to say it’s valid). Great to keep the selection of empty `buttonKind` in either way - Thanks!

emyarod: would you like me to modify it to be `'none'` instead of empty string, in line with your original intention @asudoh ? maybe `null` would work better than empty string too, since that it be just like if the user did not pass in any value for the `buttonKind` prop

asudoh: Thank you for asking @emyarod! Let’s go with `none` for now, given `null` won’t work as a object key (which has to be a string).

emyarod: oh ok so just to clarify @asudoh, is [this](https://github.com/IBM/carbon-components-react/pull/1202) more in line with what you were looking for?

PR #1212, opened 2018-08-21
fix: Added an aria label to the list box, resolves #1192.

### LINKED ISSUE #1192: MultiSelect: Accessibility failure due to widget name
When using the MultiSelect component, we're seeing the following error in DAP that's causing our accessibility tests to fail:

image

After version `6.18.4` was released last Monday we were passing all of our accessibility tests with the same usage of MultiSelect, and since then this error has surfaced. We're currently on version `6.20.0`.

zgrossbart:
Closes #1192

This change adds a default aria label to list boxes so we can pass the DAP tests.

This change also adds that new aria label to unit tests.

### DIFF
--- a/src/components/ListBox/ListBox.js
+++ b/src/components/ListBox/ListBox.js
@@ -95,7 +95,7 @@ ListBox.defaultProps = {
   innerRef: () => {},
   disabled: false,
   type: 'default',
-  ariaLabel: 'ListBox',
+  ariaLabel: 'Choose an item',
 };
 
 export default ListBox;
--- a/src/components/ListBox/ListBox.js
+++ b/src/components/ListBox/ListBox.js
@@ -22,6 +22,7 @@ const handleClick = event => {
  * container class name in response to certain props.
  */
 const ListBox = ({
+  ariaLabel,
   children,
   className: containerClassName,
   disabled,
@@ -42,6 +43,7 @@ const ListBox = ({
       <div
         {...rest}
         role="listbox"
+        aria-label={ariaLabel}
         tabIndex="0"
         className={className}
         ref={innerRef}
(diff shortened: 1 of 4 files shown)


src/components/ListBox/ListBox.js
```diff
@@ -82,12 +84,18 @@ ListBox.propTypes = {
...
+   * Specify the "aria-label" of the ListBox.
+   */
+  ariaLabel: PropTypes.string,
 };
 
 ListBox.defaultProps = {
   innerRef: () => {},
   disabled: false,
   type: 'default',
+  ariaLabel: 'ListBox',
```
joshblack: Should we have a default value for this, or just leave it blank?

zgrossbart: Having a default value here doesn't hurt anything and it will help users that are using screen readers.

zgrossbart: It's also worth noting that many other Carbon components have a default value for the aria-label.

joshblack: @zgrossbart totally understand, it might help if the label itself is more descriptive then if the intent is to aid the user here by default. List Box is something that could cause confusion as that term is not very common, or at least that's my assumption.

zgrossbart: I based the defaults on the other defaults which were very simple.  Just identifying what the element was.  I'm happy to change it.  Do you have another suggestion?

joshblack: How about "Choose an item" or "Choose an option"? Just basing off of: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

zgrossbart: Sounds good to me.

Issue #1220, opened 2018-08-22
Date Picker: allow selector to be used for the appendTo prop


samrose3:
## Detailed description

The [`flatpicker`](https://flatpickr.js.org/) created by the Date Picker component allows the `appendTo` option to be configured using the component props. The `PropType` for the `appendTo` prop only supports a DOM element to be passed to the `flatpicker` configuration.

**TL;DR**
Allow the `appendTo` prop to use a `String` selector or DOM element to allow for situations where the selector is known but the DOM element doesn't exist yet.

**Specific issue scenario**
When restyling the calendar in a Storybook component environment, I wanted to pass a DOM element used to scope the CSS styling as the `appendTo` prop. However, when Storybook is compiling stories to be rendered in the `addStoryWithInfo` function, the DOM element I want to append to doesn't exist yet.

To get around this, I want to pass in the known DOM selector and allow the `DatePicker` component to grab the DOM element on mount if the prop is a string. This doesn't break the current implementation of the `appendTo` prop and allows for delayed DOM selection in cases where it is needed such as above.

## Additional information

Ideal Storybook story declaration
```jsx

const datePickerProps = {
  id: 'date-picker',
  appendTo: '.story',    // allow selector here since DOM element doesn't exist yet
  onChange: datePickerOnChangeActions('onPickerChange'),
};

storiesOf('DatePicker', module)
  /* ... previous stories ... */

  .addWithInfo(
    'single with calendar',
    `
      A single Date Picker consists of an input field and a calendar.
    `,
    () => (
      <DatePicker
        {...datePickerProps}
        datePickerType="single"
        dateFormat="m/d/Y">
        <DatePickerInput {...datePickerInputProps} />
      </DatePicker>
    )
  )
```

PR #1222, opened 2018-08-22
feat(date-picker): append calendar to selector or DOM element

samrose3:
Allow the Date Picker calendar to be appended to either a passed DOM
element or a string selector

Closes IBM/carbon-components-react#1220

#### Changelog

**New**

**Changed**

* `appendTo` prop now accepts a string DOM selector as well as a DOM element which to append the `flatpickr` calendar

**Removed**


src/components/DatePicker/DatePicker.js
```diff
@@ -236,9 +236,13 @@ export default class DatePicker extends Component {
       const onHook = (electedDates, dateStr, instance) => {
         this.updateClassNames(instance);
       };
+      const appendToNode =
+        typeof appendTo === 'string'
+          ? document.querySelector(appendTo) || document.body
```
asudoh: Can we remove `|| document.body` so that Flatpickr itself can handle that scenario?

samrose3:
@asudoh great point!

Removed the extra code as suggested.  
Should be all good to go now! Thank you 👍

PR #1231, opened 2018-08-23
docs: add initial set of prop type descriptions

joshblack:
Add initial set of prop type descriptions for components

#### Changelog

**New**

**Changed**

* A lot of component props now have prop types

**Removed**


src/components/ComposedModal/ComposedModal.js
```diff
@@ -174,38 +231,58 @@ export class ModalHeader extends Component {
   }
 }
 
-export class ModalBody extends Component {
```
asudoh: Seems that you are cleaning up stuffs a bit, but I couldn’t find where `<ModalBody>` is gone - Any pointers?

joshblack: Haha yeah, totally missed this 😅 adding back in!

joshblack:
cc @lovemecomputer

src/components/FileUploader/FileUploader.js
```diff
@@ -115,17 +172,22 @@ export class FileUploaderButton extends Component {
...
     status: PropTypes.oneOf(['edit', 'complete', 'uploading']),
-    tabIndex: PropTypes.number,
-    onKeyDown: PropTypes.func,
   };
 
   static defaultProps = {
     onKeyDown: () => {},
     status: 'uploading',
     style: {},
-    tabIndex: 0,
```
asudoh: This prop seems to be included in `…other` portion. Changing default prop here seems to affect there?

joshblack: Good call!

Issue #500, opened 2018-01-17
Update Tabs to use new fragment syntax


joshblack:
This might be the case for multiple components, but it came up today that Tabs is currently using a containing `div` to wrap it's markup. With the update to React 16, we can instead return a React Fragment so that we don't need the containing `div` node. This issue is for the implementation of this update.

lovemecomputer:
addressed in https://github.com/IBM/carbon-components-react/pull/1232

PR #1232, opened 2018-08-23
Tabs fragment syntax

lovemecomputer:
Closes IBM/carbon-components-react#500

replaces `<div> … </div>` with `<React.Fragment> … </React.Fragment>`

#### Changelog

**New**

* new Fragment syntax

**Removed**

* the wrapping div


src/components/Tabs/Tabs.js
```diff
@@ -173,7 +173,7 @@ export default class Tabs extends React.Component {
     const selectedLabel = selectedTab ? selectedTab.props.label : '';
 
     return (
-      <div>
+      <React.Fragment>
```
asudoh: Can we use `Fragment` named import from `react` (as several other components do)?

lovemecomputer:
ah, just realized i messed up with branches and included a change from another branch! gotta remove that file from this PR

src/components/TimePicker/TimePicker-story.js
```diff
@@ -62,8 +62,8 @@ storiesOf('TimePicker', module)
             <SelectItem value="PM" text="PM" />
           </TimePickerSelect>
           <TimePickerSelect id="time-picker-select-2" {...selectProps}>
-            <SelectItem value="Timezone 1" text="Timezone 1" />
-            <SelectItem value="Timezone 2" text="Timezone 2" />
+            <SelectItem value="Time zone 1" text="Time zone 1" />
```
asudoh: I guess this change was for another PR…?

lovemecomputer: @asudoh that's correct, i made this branch off of that branch instead of master, so it got looped into here

src/components/Tabs/Tabs.js
lovemecomputer: @asudoh thanks! looking at other files to maintain consistency

lovemecomputer: @asudoh looking at other files, it looks like `import` with `<Fragment>` is only used once, but `<React.Fragment` is used 4 or 5 times. sticking with `<React.Fragment>` in this PR

lovemecomputer:
improper file removed from this PR, so all should be good to go!

src/components/Tabs/Tabs.js
asudoh: @lovemecomputer Thank you for doing the research!

joshblack:
🎉

lovemecomputer:
thanks for checking on those issues @asudoh !

PR #1233, opened 2018-08-24
RFC: chore(components): use named exports from React

asudoh:
This change goes away from `React.*` and `ReactDOM.*` and uses named exports from `react` and `react-dom` instead. The merits are:

* Ability of static analysis (e.g. ESLint will catch typo which cannot be done with object properties)
* Slightly better minification (Minifier cannot shorten object property names)

I understand above merits are minor and up to team’s taste. Anybody feel free to bring up your preference - The goal is code consistency rather than the specific choice this PR has.

#### Changelog

**Changed**

* From `React.*` and `ReactDOM.*` to named exports from `react` and `react-dom`

### DIFF
--- a/src/components/Accordion/Accordion.Skeleton.js
+++ b/src/components/Accordion/Accordion.Skeleton.js
@@ -1,9 +1,9 @@
-import React from 'react';
+import React, { Component } from 'react';
 import Icon from '../Icon';
 import SkeletonText from '../SkeletonText';
 import { iconChevronRight } from 'carbon-icons';
 
-export default class AccordionSkeleton extends React.Component {
+export default class AccordionSkeleton extends Component {
   render() {
     const item = (
       <li className="bx--accordion__item">
--- a/src/components/Breadcrumb/Breadcrumb.Skeleton.js
+++ b/src/components/Breadcrumb/Breadcrumb.Skeleton.js
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { Component } from 'react';
 
-export default class BreadcrumbSkeleton extends React.Component {
+export default class BreadcrumbSkeleton extends Component {
   render() {
     const item = (
       <div className="bx--breadcrumb-item">
--- a/src/components/BreadcrumbItem/BreadcrumbItem.js
+++ b/src/components/BreadcrumbItem/BreadcrumbItem.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Children, cloneElement } from 'react';
 import classnames from 'classnames';
 import Link from '../Link';
 
@@ -7,7 +7,7 @@ const newChild = (children, href) => {
   if (typeof children === 'string' && !(href === undefined)) {
     return <Link href={href}>{children}</Link>;
   } else {
-    return React.cloneElement(React.Children.only(children), {
+    return cloneElement(Children.only(children), {
       className: 'bx--link',
     });
   }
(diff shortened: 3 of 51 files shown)


lovemecomputer:
I'm good with using `import React, { Component } from 'react'`, that's what I am accustomed to doing in previous work, always felt right, but I'm not currently knowledgeable enough to have an opinion on the merits outside of what you already outlined. If it's beneficial for `Component`, then I'd imagine it's beneficial for most everything. 

I don't know the team's opinion on this regarding workflow on consistent imports, or readability, so I'm just here to listen to that discussion if there is any 🙂

---

Regarding `<React.Fragment>`/`<Fragment>`: is there any plan to move toward the new [short syntax](https://reactjs.org/docs/fragments.html#short-syntax) ?\

asudoh:
@lovemecomputer @alisonjoseph Thank you for your thoughts/reply! Wrt `<></>`, I was thinking of having another discussion in the team - Personally on the fence. While it’s more concise (cool!), I know that several toolstack (e.g. syntax engine for some editors) hasn’t supported that yet.

joshblack:
I think this is a great start for helping to improve stuff, especially for users downstream, although I think the benefits listed seem like they might be mitigated by our transpiled output.

The forms:

```js
import React from 'react';
import React, { Component } from 'react';
```

Both end up becoming:

```js
var _react = require("react");

var _react2 = _interopRequireDefault(_react);

function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : { default: obj };
}
```

Where `Component` just becomes bound to: `_react2.default.Component`. This also means that the minification benefits follow this trend, as well since the value is bound to `_react2.default.Component`.

---

With respect to the fragment syntax, definitely fine with doing `<></>` 👍

aledavila:
+1 on the <></> and updating.

alisonjoseph:
I'm fine w/ `<></>`

lovemecomputer:
also +1 on `<></>` if our toolchain is cool with it! just wanna confirm, do we need to import `React.Fragment` to use them in a file?

asudoh:
@lovemecomputer Good point! I see two options:

1. Standardize on `<Fragment></Fragment>`
2. `<></>` for regular cases, `<Fragment key={theKey}></Fragment>` there `key` is required

Which one do you guys like? Thanks!

asudoh:
@joshblack Good point from you, too!

joshblack:
> <></> for regular cases, <Fragment key={theKey}></Fragment> there key is required

This should work out, I think 👍

asudoh:
@joshblack While the typo concern still remains, it’s all up to you guys on the specific choices in this PR - The whole purpose of this PR again is improving the coherency. Thanks!

asudoh:
Thank you for putting more thoughts on this @lovemecomputer! I think “searchablity” thing is for getting a shortcut of searching for `React.Component` vs. looking up `import { Component } from ‘react'` statement (again I’m fine either way).

lovemecomputer:
I resolved 2 merge issues that was just swapping `React.Component` for `Component`. It looks like we've already settles on Fragment `<></>` syntax being chill, as I've seen it cropping up throughout the codebase. I like it! to me it makes it very obvious at a glance that it really is an empty element.

are we still waiting for word on `Component` vs `React.Component`? if that's a holdup getting this PR merged, maybe we should split the Fragment syntax discussion and the importing Component discussion syntax into 2 separate PRs.

asudoh:
@lovemecomputer Yeah, fair enough - Did it! #1317

joshblack:
Are we okay to close this PR since we had the fragment stuff pulled out into #1317?

Issue #1111, opened 2018-07-20
DataTable sorted column arrow direction wrong


rikkertbf:
In exploring use of DataTable component in our app, noticed that the sort order indication in the table header appears to be _counter-intuitive_ (and inconsistent with most other grid impls - including our existing table component). 
I.E. up arrow represents DESC (as well as unordered state), and down arrow for ASC...
Confirmed per #carbon-react channel discussion.

Found in sample provided here: 
https://pages.github.ibm.com/cloud-integration-design/ap-carbon-theme/?selectedKind=DataTable&selectedStory=with%20sorting&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

![image](https://user-images.githubusercontent.com/18743489/42980593-03fce60c-8ba6-11e8-81ac-373ab139e8fe.png)

lovemecomputer:
addressed in https://github.com/IBM/carbon-components-react/pull/1234

@rikkertbf  thanks so much for noticing this! it's a small subtle thing, i think it was just temping to type "caret down" for descending order, but your right, it should be "caret up", at least to match with the mac programs i'm looking at!

Bluedrop:
I think this should be configurable instead of fixed one way or another. I've encountered libraries and tools that use it differently, as well as client demanding the arrows to be pointing a certain way. Granted you can program each column sort as needed, but if you need all tables in the entire app to just show it "backwards" it would be so much easier to just configure what icon to use....

There is logic in both sorting orders: iconCaretUp has a **small** top and **big** bottom (suggesting the ascending order) and iconCareDown has a **big** top and a **small** bottom. Granted not so significant on strings, but if we're talking about **big** and **small** numbers the suggested sorting is different.

Just a thought :) I may fork this and add a configurable icon if that's permitted :)

PR #1234, opened 2018-08-24
fix(sort icon): reverse direction of svg sort direction icon

lovemecomputer:
Closes IBM/carbon-components-react#1111

In the headers for sortable tables, the sort direction icon is opposite that of Finder etc.

I think it was tempting to use `iconCaretDown` for *descending* order, since the words match. but graphically, most programs us an upward-facing arrow for descending order. 

resolved by swapping to use `iconCaretUp`

#### Changelog

**Changed**

* sortable headers now use `iconCaretUp` instead of `iconCaretDown`


alisonjoseph:
One comment, you have this labeled as "chore". I don't believe that will trigger an automatic release? Only fix and feature will.

lovemecomputer:
@joshblack @alisonjoseph @asudoh thanks y'all! okay, ig that was my first time to make a change that touched something a snapshot would be compared against, so i messily resolved it :p 

looks like all is go for now! i was having some issues with git on my local machine so i manually edited the snapshot inside of github. lemme know if all is cool!

alisonjoseph:
@lovemecomputer can you check if we need to switch this icon in the `carbon-components` repo? You can see the vanilla example on the website http://carbondesignsystem.com/components/data-table/code and there is also a codepen https://codepen.io/team/carbon/full/MOEagV/

lovemecomputer:
@alisonjoseph sorting behavior in those examples doesn't work on my end, it just rotates the arrow without actually sorting the table. however, on first click, the arrow seems to point in the correct direction for ascending order. 

maybe @asudoh can comment? [here's the file defining orientation of the sort arrow in carbon-components](https://github.com/IBM/carbon-components/blob/master/src/components/data-table-v2/_data-table-v2-sort.scss).

i'm seeing some inconsistent errors in the react sort table example behavior too—going to test locally to see if it's just a problem on storybook.

i also checked the sketch file and the icon is oriented correctly there :)

asudoh:
@lovemecomputer Vanilla code does not take care of actual sorting. Data manipulation on table tends to involve application-specific logics, as well as JavaScript framework-specific logics, and thus we relegate that, including sorting, to user-land.

PR #1239, opened 2018-08-27
feat(DropdownV2): Update dropdownV2 to allow rendering of nodes as dr…

Anisa9514:
…opdown options #1228

Closes IBM/carbon-components-react#1228

This is an enhancement to the DropdownV2 component to allow for the rendering of nodes within the dropdown options instead of a string.

#### Changelog

**New**
* itemToNode prop in DropdownV2 that converts option item to node
* unit test to check if nodes are rendered


joshblack:
Hi @Anisa9514! We try to keep enhancements/changes to the project scoped inside of issues before creating Pull Requests. This just helps to make sure that you're time spent working on this is time well spent 👍 

Would you mind moving this over to the issue you linked for a discussion around this topic so we can make sure that introducing this feature is something that we can and should do? Going to close this in the interim, feel free to re-open it or comment on it if the work applies to what we end up deciding over there. Thanks!

Issue #1236, opened 2018-08-26
HTML title and description should be updated to improve search results


vpicone:
## Detailed description

When searching for the library, the lack of page title and description makes it difficult to suss out the correct link:

<img width="584" alt="Search results" src="https://user-images.githubusercontent.com/4078018/44630237-fa1bb400-a91f-11e8-8c8e-4bfc941ca014.png">

## Additional information

* It looks like Storybook doesn't offer a clean escape hatch to do this via the options addon, however, it could be added with a `manager-head.html` file. 
* Unfortunately, you can't really overwrite the original title tag but a simple <script> could do the trick.
* This would also be a great opportunity to add meta tags for sharing, a custom favicon etc.

PR #1249, opened 2018-08-28
docs: update meta tags on root html

vpicone:
Closes IBM/carbon-components-react#1236

This adds some meta tags to improve search results/sharing of Carbon Components React. At this time, Storybook doesn't offer an option to change the page title so it needs to be done with a script tag. Tacking on a second title tag is a no-op.

#### Changelog

**New**

* `manager-head.html` is the escape hatch provided by Storybook for this sort of thing.

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -1,0 +1,19 @@
+<meta name="keywords" content="IBM, design, system, Carbon, design system, Bluemix, styleguide, style, guide, components, library, pattern, kit, component, cloud, React, React.js">
+<meta name="description" content="This is the React implementation of the Carbon Design System. Carbon is a series of individual styles and components, that when combined make beautiful, intuitive designs.">
+
+<!-- Open Graph -->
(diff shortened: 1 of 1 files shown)


.storybook/manager-head.html
```diff
@@ -0,0 +1,19 @@
...
+<meta property="og:url" content="http://react.carbondesignsystem.com">
+
+<!-- Social -->
+<meta name="twitter:card" content="summary_large_image">
+<meta name="twitter:image:alt" content="Carbon Design System">
+<meta name="twitter:site" content="@_carbondesign">
+
+<!-- Storybook override -->
+<script>
+    document.title = 'Carbon Components React'
```
asudoh: Can we do two spaces here?

.storybook/manager-head.html
```diff
@@ -0,0 +1,19 @@
...
+</script>
```
asudoh: Can we add a LF here?

vpicone:
@asudoh you got it!

Issue #857, opened 2018-05-03
childrenOf and childrenOfType fail with react-hot-loader


liviudobrea:
## Detailed description

`react-hot-loader (v4.x.x)` (will further refer to this as **RHL**) wraps every component in a **ProxyComponent**, to internally handle the change / reload mechanism. That is all good, however, by using `RHL` with `carbon-components-react`, this statement https://github.com/carbon-design-system/carbon-components-react/blob/706584394a321c30c56367f1fee930a82fb26236/src/prop-types/childrenOf.js#L24-L28 and this statement https://github.com/carbon-design-system/carbon-components-react/blob/706584394a321c30c56367f1fee930a82fb26236/src/prop-types/childrenOfType.js#L20-L24  will always throw.

```javascript
warning.js:33 Warning: Failed prop type: Invalid prop `children` of type `ListBoxField` supplied to `ListBox`, expected each child to be one of: `[ListBoxField, ListBoxMenu]`.
```

> Is this a feature request (new component, new icon), a bug, or a general issue?

This is not necessary an issue with the current code, but it becomes a bug when using this with the **RHL** 3rd party library (which I'm guessing many are).

This is an acknowledged issue with within **RHL**. More details here https://github.com/gaearon/react-hot-loader/issues/304.

Given that the `RHL` API might change at any time, one temporary solution could be making `RHL` a dependency of this project, and do something like this.

### Solutions
> Depend on `react-hot-loader`

I'm not sure if this is a good enough reason to add **RHL** as a project dependency, but it provides the cleanest solution.
```javascript
import { areComponentsEqual } from 'react-hot-loader';

...

if (
  !expectedChildTypes.includes(child.type) && 
  (
    typeof child.type === 'function' && 
    !areComponentsEqual(child.type, expectedChildTypes.find(c => c.name === childDisplayName)))
  )
)
```
> Don't depend on `react-hot-loader`

This one's a tad uglier, but it also works. However, since **RHL** API could change at any given time, this is more likely to fail at some point.

```javascript
(
  typeof child.type === 'function' &&
  Object.prototype.hasOwnProperty.call(
    child.type,
    '__reactstandin__getCurrent'
  ) && 
  child.type.__reactstandin__getCurrent() !== expectedChildType
)
```

More than happy to PR if you're ok with either of those.

joshblack:
[...]

PR #1253, opened 2018-08-29
Fix(#857) ProxiedComponents Type Checking

liviudobrea:
Closes IBM/carbon-components-react#857

Add internal helper for proxied component type checking.

#### Changelog

**New**

* React-proxy dev dependency

**Changed**

* childrenOf and childrenOf-Type


src/prop-types/tools/areComponentsEqual.js
```diff
@@ -0,0 +1,12 @@
+const RHL_PREFIX = '__reactstandin__getCurrent';
+
+const getProxyOrType = c => {
+  const isProxy = c && typeof c !== 'string' && Reflect.has(c, RHL_PREFIX);
```
asudoh: `Reflect` is not supported by IE11 - Any alternatives?

src/prop-types/tools/areComponentsEqual.js
```diff
@@ -0,0 +1,12 @@
...
+  return isProxy ? c[RHL_PREFIX]() : c;
+};
+
+const areComponentsEqual = (a, b) => {
```
asudoh: Probably we'd take an array for `b`, or make this function varargs?

src/prop-types/tools/areComponentsEqual.js
liviudobrea: Sure, will add fallback.

src/prop-types/tools/areComponentsEqual.js
liviudobrea: Yep, we can, but I wouldn't do that in the helper method though. I still prefer this https://github.com/IBM/carbon-components-react/blob/47b7447f18792f8d1655863f75dcf749cec19ce4/src/prop-types/childrenOf.js#L20-L33

joshblack:
Also, make sure to run `yarn prettier` so that ci checks pass on formatting 😄

liviudobrea:
There we go, replaced `Reflect` in favour of globally supported js API. Ran `prettier` and tested with our own project successfully.

liviudobrea:
@joshblack Can't merge master inside branch from git UI until
```
Merging is blocked
Merging can be performed automatically once the requested changes are addressed.
```
Doesn't look like it's something that I can do from my side.

alisonjoseph:
Closing this due to inactivity. Feel free to reopen if this is something you still need and have time to make the requested updates.

Issue #1259, opened 2018-08-30
@storybook/addon-options should be a dev dependency


vpicone:
It was added as a regular dependency

PR #1261, opened 2018-08-30
docs(Storybook): switch @storybook/addon-options to dev dependency

vpicone:
Closes IBM/carbon-components-react#1259

@storybookaddon-options was added as a regular dependency when it should have been a dev dependency.

#### Changelog

**Changed**

* @storybook/addon-options from `dependencies` to `devDependencies`


alisonjoseph:
Agreed @vpicone, although luckily this isn't something that gets updated often. Just have to remember we need to run prettier before merging in the changes to issue templates.

Issue #1264, opened 2018-08-30
onClick function of Switch component does not work


dejunqi2008:
## Detailed description

As the title, `onClick` attribute of the `Switch` component is not working. Reproduced with `codesandbox.io`: https://codesandbox.io/s/046rox770w

```
import React from "react";
import { render } from "react-dom";
import { Button, ContentSwitcher, Switch } from "carbon-components-react";

const onChange = () => {
  // console.log("change");
};

const App = () => (
  <div>
    <ContentSwitcher onChange={onChange}>
      <Switch
        name="one"
        text="First section"
        onClick={() => console.log("is clicked")}
      />
      <Switch
        name="two"
        text="Second section"
        onClick={() => console.log("is clicked")}
      />
      <Switch
        name="three"
        text="Third section"
        onClick={() => console.log("is clicked")}
      />
    </ContentSwitcher>
  </div>
);

render(<App />, document.getElementById("root"));

```

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to see changed?

> What browser are you working in?

> What version of the Carbon Design System are you using?

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in code sandbox
> https://codesandbox.io/s/r4zxq9pvq

## Additional information

* Screenshots or code
* Notes

PR #1265, opened 2018-08-30
feat(Switch): support onClick handler on component

joshblack:
Closes IBM/carbon-components-react#1264

Support `onClick` handler on the `Switch` component

#### Changelog

**New**

**Changed**

* `events` tool in DataTable moved to top level `tools` folder
* `DataTable` now uses updated path to events
* `ContentSwitcher` now composes event handlers for `onClick`

**Removed**


joshblack:
cc @lovemecomputer

PR #1281, opened 2018-09-07
feat(Modal): ensure passing along evt for onRequestClose()/onRequestSubmit()

### LINKED ISSUE #1271: ComposedModal: ESC key closes all the opened modals at once
## Detailed description

I have 2 use cases:
1. `modal B` inside `modal A`. `modal B` can be opened only from inside of `modal A` (and is rendered inside as well). When `modal B` is closed, I expect `modal A` to stay opened.
2. form element inside modal. When user edits something in a form and tries to close a modal, I want to ask user does he want to close it before saving new data, but pressing ESC closes it automatically.

No.1 is way more important in my case.

ComposedModal

I'd like to be able to control, when ESC closes it. Similar to `Modal` component. Btw, I'm not using `Modal`, because I have a requirement not to close modal, when outside click occurs, which is a case for `Modal`.

latest

1. go to https://codesandbox.io/s/qxo6xx92q4
2. click `Open modal`
3. click `Open inner modal`
4. press ESC

asudoh:
This change allows consumers to determine whether modal should be closed or not by looking at what event requested submitting/closing modal.

Resolves #1271.

#### Changelog

**New**

* Code to ensure passing along event for `onRequestClose()`/`onRequestSubmit()` in `<Modal>`


asudoh:
@joshblack If you are referring to `This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property cancelable on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://facebook.github.io/react/docs/events.html#event-pooling for more information.`, no - Those callbacks are called synchronously and thus the events haven’t been recycled when they are called. Let me know if you are thinking of a different thing - Thanks!

PR #1286, opened 2018-09-09
fix(clickable-tile): add rel prop

alisonjoseph:
Adds the ability to add a rel value to the clickable tile 

_related to building out Gatsby site, google lighthouse recommends `rel="noopener` on links opening in a new window_


asudoh:
Would we want to keep just the prop type? In this way consumers know that it’s supported, the type is checked, but don’t need any additional (our) code to copy the prop value to the inner-node. What do you think?

alisonjoseph:
@asudoh updated to just have prop type

alisonjoseph:
done @joshblack 😸

joshblack:
You rock! 🙏

alisonjoseph:
thanks for keeping us accountable 😉 @joshblack

Issue #1291, opened 2018-09-10
Multiselect increments the counter when trying to deselect an existing item


scottdickerson:
## Detailed description

> Describe in detail the issue you're having.

If you have a different array passed to the MultiSelect initialSelectedItems component (needed to support loading from Redux), then you cannot deselect an initial Item

> Is this issue related to a specific component?

MultiSelect
> What did you expect to happen? What happened instead? What would you like to see changed?

Deselection of the previously populated item didn't occur.

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

6.28.2

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. Try to deselect the previously populated selected item in this code sandbox here:
https://codesandbox.io/s/21xowk8vlj

You'll see the overall count go up, and the item is not deselected

## Additional information

* Screenshots or code
* Notes
![image](https://user-images.githubusercontent.com/6663002/45312952-1eb48600-b4f3-11e8-8d87-10efdeeb6b45.png)

lekterable:
Hey man, this is not a bug in the code.
Instead of passing a similar looking item pass an item from your items array.
Here is the working demo => https://codesandbox.io/s/pj53k7jz7q
The only difference is `initialSelectedItems={[items[0]]}` instead of `initialSelectedItems={[{ id: 1, label: "Item 1" }]}`

joshblack:
Hi @scottdickerson! 👋 Like @lekterable said (thanks for the support here btw!), `initialSelectedItems` should be an array containing the references to the items that are selected. I went ahead and made a feature request for improving this prop-type over in: https://github.com/IBM/carbon-components-react/issues/1292, feel free to weigh in on there if there is a better pattern to use to help with this use-case.

Going to close this for now, feel free to comment if the suggestion isn't working. Thanks!

scottdickerson:
ok I was able to get this to work, but this required lots of workarounds in my code where I ensured that as I passed down my Items array through my component stack that the selectedItems directly matched the memory address of the items array.  Can we fix it with that suggested PR instead?

scottdickerson:
this was fixed by https://github.com/IBM/carbon-components-react/pull/1314

PR #1293, opened 2018-09-10
fix(Selection): prevent selection from reselecting when deselecting

scottdickerson:
Prevent selection from reselecting when deselecting

Closes IBM/carbon-components-react#1291

{{short description}}

#### Changelog

**New**

* {{new thing}}

**Changed**

* Fix the selection criteria to actually check the id when determining reselection, this prevents the need to pass the exact in-memory item to the initialSelectedItems list.

**Removed**

* {{removed thing}}


scottdickerson:
If you remove the Selection change and go to the MultiSelect storybook for multiple items with initial Values selected, you can see the bug when you deselect the existing selection (see bug #1291).  Then putting back in the Selection change will fix it.

scottdickerson:
yes I can do that, thanks for the suggestions. @joshblack and @asudoh.

scottdickerson:
this has been superceded by this track: #1314

Issue #1298, opened 2018-09-11
ModalWarapper - after closing modal, the trigger button should return the focus to the trigger button


zerostatic:
### Summary
For the ModalWarapper - after closing the modal, we should set focus back on the trigger button that opened the modal.

### Justification
I believe this is best practice for accessibility

### Desired UX and success metrics

After closing modal, the trigger button should return the focus to the trigger button and display the focus outline.

PR #1299, opened 2018-09-11
feat(ModalWrapper): return focus to modal trigger button on modal close

emyarod:
Closes IBM/carbon-components-react#1298

This PR will return focus to the element that launches the modal when the modal is closed, which should improve keyboard a11y. This is accomplished by attaching a ref to the trigger button, which is then used to focus on the button in the callback of the `setState` operation when closing the modal

#### Changelog

**New**

* attach a ref to the `<Button>` in `<ModalWrapper>` which will launch the `<Modal>` element
* add logic to call `HTMLElement.focus()` on the button after modal is closed

**Changed**

* access `<Modal>` refs through the refs' `current` property

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -53,6 +53,8 @@ export default class Modal extends Component {
   };
 
   button = React.createRef();
+  outerModal = React.createRef();
+  innerModal = React.createRef();
 
   elementOrParentIsFloatingMenu = target => {
     if (target && typeof target.closest === 'function') {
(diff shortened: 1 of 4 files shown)


emyarod:
hi @asudoh do you mean accessing the ref through the `current` attribute? If I'm not mistaken, I don't think I can use it in this case because `<Button>` is a stateless functional component. Am I incorrectly understanding `ref` usage with functional components?

joshblack:
@emyarod I think what @asudoh is referring to is that if you use the form `React.createRef` for a component instance property, like `this.triggerButton`, then you would do `inputRef={this.triggerButton}` instead of the old callback style where you assign the button to `this.triggerButton`.

Ref: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

emyarod:
gotcha @joshblack, I think that is what allows me to use the `current` attribute like I mentioned earlier. Is [this](https://github.com/IBM/carbon-components-react/pull/1299/commits/f9e8fe35d2691ae2c1acc940d583afbffbbe0a69) more in line with what we want?

emyarod:
@joshblack is [this test](https://github.com/IBM/carbon-components-react/pull/1299/commits/cf1851de40ee98e3d21ba4a7da1cfa303d8b813d) sufficient for testing this new behavior?

Issue #1297, opened 2018-09-11
Console warning when ClickableTile is used with React 16.5


scottdickerson:
## Detailed description

> Describe in detail the issue you're having.

Simply using ClickableTile from carbon-components-react results in this warning written to the console.log
```
Warning: `ClickableTile` uses `getDerivedStateFromProps` but its initial state is undefined. This is not recommended. Instead, define the initial state by assigning an object to `this.state` in the constructor of `ClickableTile`. This ensures that `getDerivedStateFromProps` arguments have a consistent shape.
```

> Is this issue related to a specific component?

ClickableTile
> What did you expect to happen? What happened instead? What would you like to see changed?

No console errors
> What browser are you working in?

Chrome
> What version of the Carbon Design System are you using?

6.28.2

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in code sandbox
https://codesandbox.io/s/7w3pkyv116

## Additional information

* Screenshots or code
* Notes

PR #1300, opened 2018-09-11
fix(ClickableTile): add initial state to prevent warnings

lekterable:
Closes IBM/carbon-components-react#1297

Added an initial state for ClickableTile component to prevent warning.

`Warning: 'ClickableTile' uses 'getDerivedStateFromProps' but its initial state is undefined. This is not recommended. Instead, define the initial state by assigning an object to 'his.state' in the constructor of 'ClickableTile'. This ensures that 'getDerivedStateFromProps' arguments have a consistent shape.
`
#### Changelog

**New**

* initial state for ClickableTile component

### DIFF
--- a/src/components/Tile/Tile.js
+++ b/src/components/Tile/Tile.js
@@ -81,8 +81,8 @@ export class ClickableTile extends Component {
   };
 
   static getDerivedStateFromProps({ clicked }, state) {
-    const { prevClicked } = state || {};
-    return state && prevClicked === clicked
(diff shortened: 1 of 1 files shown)


lekterable:
@asudoh sure, I'm on it!

I've implemented this library in our project in IBM in Poland and working with it is a pure pleasure.
I had to give back!

lekterable:
@asudoh done, I did the same thing for SelectableTile as it also looked unnecessary.

Issue #1148, opened 2018-08-01
MultiSelect: setState() causes rendering issue with selected 


alimabean:
https://codesandbox.io/s/x94xn85p4o

![multiselect_wonky](https://user-images.githubusercontent.com/14437246/43539907-4139d99c-9594-11e8-9b78-2caf41caddc1.gif)

- When MultiSelect's onChange issues a setState, the selected items are reset yet the counter is kept.

EDIT:
- prop changes via redux also cause this issue

joshblack:
Hi @alimabean! 👋 Thanks so much for taking the time to make this issue. Definitely seems like a bug on our end, so will try and get it added to a sprint. Will also accept any PRs that help to fix this!

polgas16:
Had an issue with this awhile back. 
https://codesandbox.io/s/rrj511j4mq

trb727:
@polgas16 I don't see a bug in your example or recreate it locally. Perhaps it has been fixed in an release since you posted?

PR #1314, opened 2018-09-13
fix(MultiSelect): fix setState() causes rendering issue

trb727:
Closes IBM/carbon-components-react#1148

Update Multiselect, FilterableMultiSelect and Selection components to fix setState() rendering issues

#### Changelog

**Changed**
***Selection***
- Use `lodash.isEqual` instead of `indexOf` when determining to add or remove an item from the `selectedItems` state array. Currently if a parent component re-renders and passes the `items` prop inline or receives new props from a redux store , the objects in `items` will have new references. `indexOf` uses a strict equality check and always returns a -1 when checking if an item is selected in this scenario. Items will never be removed from the `selectedItems` array, only added.`lodash.isEqual` solves this issue.

***MultiSelect***
- Use `lodash.isEqual` to determine if item is checked. CheckBox `checked` prop currently relies on `item.id` property and this change allows more generic `items` objects. 
- Add snapshot test and test that passes inline `items` prop and calls `setState()`

***FilterableMultiSelect***
- Use `lodash.isEqual` to determine CheckBox `checked` prop instead of `indexOf` and match MultiSelect
- Destructure and order Downshift render props to match MultiSelect

**Rationale**
* The comment for the `items` prop says y'all want to keep it generic and allow users to pass in an array with a flexible object structure. These changes will support that with the following constraints:
   * `items` objects include a `label` property so it works with Downshift by default if a `itemToString` prop is not passed
   * `itemToString` prop is passed that returns a string

* `lodash.isEqual` is used in other places in the library

**Questions**
- Implementation relies on `findIndex` array function which ie11 does not support. Can this be covered by adding to the list of polyfills? It could be useful to use this function in other parts of the library as well. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)
- Any concerns performance-wise or other about always doing object equality checks with lodash.isEqual?


src/components/MultiSelect/MultiSelect.js
```diff
@@ -207,11 +208,9 @@ export default class MultiSelect extends React.Component {
                       const itemProps = getItemProps({ item });
                       const itemText = itemToString(item);
                       const isChecked =
-                        selectedItem
-                          .map(selected => {
-                            return selected.id;
-                          })
-                          .indexOf(item.id) !== -1;
+                        selectedItem.findIndex(selected => {
```
asudoh: `.findIndex()` is not supported by IE11 and we refrain from usage of new polyfills given it tends to be a breaking change. Could you try an implementation without `.findIndex()`?

trb727: Okay good to know 👍 that answers a question I had. I will try one without it

src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -252,9 +252,11 @@ export default class FilterableMultiSelect extends React.Component {
                       const itemProps = getItemProps({ item });
                       const itemText = itemToString(item);
                       const isChecked =
-                        selectedItem.findIndex(selected => {
-                          return isEqual(selected, item);
-                        }) !== -1;
+                        selectedItem.indexOf(
+                          selectedItem.find(selected => {
```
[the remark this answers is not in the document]
trb727: Ah right, my apologies! I have tried once again with es5 methods which I double checked for support on MDN with and [kanganx](https://kangax.github.io/compat-table/es5/).

PR #1315, opened 2018-09-13
Issue1247

David-Yuen:

Add feature to allow DataTable to have preselected rows on initialization. Initial rows being sent in can optionally be provided a isSelect boolean that will set if the row should be selected or unselected by default.

#### Changelog

**Changed**

DataTable Normalize tools to accept a parameter isSelected (optional), which can set if the row should be selected or not


src/components/DataTable/tools/normalize.js
```diff
@@ -19,7 +19,7 @@ const normalize = (rows, headers, prevState = {}) => {
     // expansion
     rowsById[row.id] = {
       id: row.id,
-      isSelected: false,
+      isSelected: row.isSelected ? true : false,
```
asudoh: Though the code here is totally fine, wondered if you wanted is type coercion of `row.isSelected`? If so, you can do: `!!row.isSelected` or `Boolean(row.isSelected)`

David-Yuen: Actually I was thinking really simple: to ensure isSelected is either a value of `true` or `false`, we just directly set it to value `true` or `false`, instead of grabbing it from row.isSelected. 

I didnt put in a PropTypes check, but there could be one in DataTable.js > propTypes.rows > .. > isSelected. 

Also if `Boolean(row.isSelected)` will return false when row.isSelected, then I'm thinking it would be very clean to just set:
```
isSelected: Boolean(row.isSelected)
```

I haven't tested, but if the above works, it could be rewritten like this? 
You guys are the experts, please let me know what's better.

asudoh: Thank you for giving this a thorough thought @David-Yuen! Both `!!row.isSelected` and `Boolean(row.isSelected)` yields `true` if `row.isSelected` is truthy (e.g. `true`, non-zero numbers, non-null object, non-empty strings). It yields `false` otherwise.

That said, your example should work. Also agreed that prop type check there will be a good addition. Thanks!

David-Yuen: Added the prop type check for isSelected, and changed it to Boolean(row.isSelected)

joshblack:
Going to reiterate here that we should not have `isSelected` in the row definition as it can conflict with user-supplied values in the `rows` object. We have to be agnostic in that prop because of the fact that `rows` could contain any data/shape/property name.

alisonjoseph:
closing this for now. can always reopen in future

Issue #1289, opened 2018-09-10
DatePicker - calendar icon displayed too low when no labelText 


dzaqb:
DatePicker displays calendar icon below a date field when DatePickerInput labelText prop is empty. 
I cannot set labelText because it is displayed above the date field while I need it in another place. 
In such case, the calendar icon position is calculated based on some absolute position which doesn't take into account that there is no label above. 

![image](https://user-images.githubusercontent.com/43142933/45305328-54209a00-b51a-11e8-96cb-45a905a76be6.png)

tw15egan:
Seems like `bx--date-picker--nolabel` class needs to be ported over, as we support a no-label version in the vanilla repo

lekterable:
Thanks @tw15egan for the investigation, I'd love to open a PR for that.

tw15egan:
Thanks for taking the time to fix this, @lekterable!

PR #1316, opened 2018-09-13
fix(DatePickerInput): apply nolabel class if label is undefined

lekterable:
Closes IBM/carbon-components-react#1289

If label is undefined `bx--date-picker--nolabel` class will be applied to container, thanks to that icon will be displayed correctly.

#### Changelog

**New**

* `bx--date-picker--nolabel` class

### DIFF
--- a/src/components/DatePicker/DatePicker.js
+++ b/src/components/DatePicker/DatePicker.js
@@ -389,6 +389,9 @@ export default class DatePicker extends Component {
           : null;
   };
 
+  isLabelTextEmpty = children =>
+    children.every(child => !child.props.labelText);
+
   render() {
     const {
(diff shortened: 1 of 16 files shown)


lekterable:
wow ur fast @asudoh

lekterable:
@tw15egan that's a good catch, thanks! I just checked and this:
>  Seems like adding bx--date-picker--nolabel to the bx--date-picker container fixes the issue.
doesn't work for both, using this alone single calendar icon is rendered too high.
Looks like we have to have this in both components.

tw15egan:
Sounds good. Once the changes are pushed up I'll take a look

lekterable:
Now it should work for both of them, WDYT @tw15egan ?

Issue #1304, opened 2018-09-12
Selected table row background color is missing


alisonjoseph:
selected table row background displays in vanilla library but not react

PR #1322, opened 2018-09-14
fix(DataTable): add background color to selected table row

emyarod:
Closes IBM/carbon-components-react#1304

Right now the `TableRow` component has no access to whether or not it is selected, so it would be up to the user to pass in the `selected` prop, which would toggle the Carbon CSS class. I'm not sure if this is the approach we want to take, but I would like some feedback on this implementation. After the approach has been decided I can edit the remaining storybook examples to reflect it

#### Changelog

**Changed**

* pass `selected` prop into `<TableRow>` and `getRowProps` method for expansion rows
* toggle CSS class for table row background color on selection

### DIFF
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -174,6 +174,7 @@ export default class DataTable extends React.Component {
       onExpand: composeEventHandlers([this.handleOnExpandRow(row.id), onClick]),
       isExpanded: row.isExpanded,
       ariaLabel: t(translationKey),
+      selected: row.isSelected,
     };
   };
 
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -174,7 +174,7 @@ export default class DataTable extends React.Component {
       onExpand: composeEventHandlers([this.handleOnExpandRow(row.id), onClick]),
       isExpanded: row.isExpanded,
       ariaLabel: t(translationKey),
-      selected: row.isSelected,
+      isSelected: row.isSelected,
     };
   };
 
--- a/src/components/DataTable/TableExpandRow.js
+++ b/src/components/DataTable/TableExpandRow.js
@@ -12,14 +12,14 @@ const TableExpandRow = ({
   isExpanded,
   onExpand,
   expandIconDescription,
-  selected,
+  isSelected,
   ...rest
 }) => {
   const className = cx(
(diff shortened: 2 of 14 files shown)


tw15egan:
This implementation makes sense to me. Could you update the stories to show this behavior?

emyarod:
sure @tw15egan. One question about `getRowProps` though, is there a reason why it is only used for the expandable table rows and not for the default table row stories? Rather than `<TableRow key={row.id} selected={row.isSelected}>` we could just use `<TableRow {...getRowProps({ row })>` like the `<TableExpandRow>`

tw15egan:
I'm not too sure, @joshblack might know why

src/components/DataTable/DataTable.js
```diff
@@ -174,6 +174,7 @@ export default class DataTable extends React.Component {
       onExpand: composeEventHandlers([this.handleOnExpandRow(row.id), onClick]),
       isExpanded: row.isExpanded,
       ariaLabel: t(translationKey),
+      selected: row.isSelected,
```
joshblack: Could this be `isSelected` to match the other prop naming conventions?

src/components/DataTable/TableExpandRow.js
```diff
@@ -12,12 +12,14 @@ const TableExpandRow = ({
   isExpanded,
   onExpand,
   expandIconDescription,
+  selected,
```
joshblack: Update to match `isSelected` note above

src/components/DataTable/TableRow.js
```diff
@@ -1,10 +1,17 @@
 import React from 'react';
 import omit from 'lodash.omit';
+import classNames from 'classnames';
```
joshblack: Could we use the `cx` convention for the name here? e.g. `import cx from 'classnames'`

src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
```diff
@@ -117,7 +117,9 @@ exports[`DataTable selection should have select-all default to un-checked if no
           <TableHead>
             <thead>
               <TableRow>
-                <tr>
+                <tr
+                  className=""
```
joshblack: Could we update the `classNames` usage so that we don't have an empty `className` generated in the snapshot?

PR #1323, opened 2018-09-14
Issue1285

David-Yuen:

Fix issue 1285, where DataTable currently has no proper way to register/get selected row changes. 
Added an event handler prop which can be provided by user (optional). Whenever a row is selected or unselected (including the selectAll / deselectAll), it triggers onSelectChange(), which will allow developers to capture the changes/differences.

#### Changelog

Added onSelectChange to DataTable


joshblack:
Hi @David-Yuen! 👋 Thanks so much for taking the time to make this Pull Request! For API changes like this, we suggest first creating an issue talking about the proposed change before making the Pull Request. This makes sure that we align on the problem you're trying to solve for, as well as the API change to get there, in order to make sure that all the work you do is productive.

For this PR specifically, we'd want to have something more abstract than `onSelectionChange`. Unfortunately, if we added props for each state change we would have ones for selection, expansion, sorting, etc. Ideally, this would all be encapsulated under an `onChange` API for folks to subscribe to and a set of exported state change types that people could use to filter out the stream for the events that they are interested in.

joshblack:
@David-Yuen the goal isn't for us to control the implementation, the idea is that we should get a consensus around a proposed API before trying to merge work in, that's all. If you have any suggestions/recommendations to improve this process, please let us know 😄

David-Yuen:
I support the fact that the code design/architecture should be proper and formatted. 

But I'll highlight what I'm seeing as an issue.

The problem is, for someone who is not familiar (such as me) who in this case, needs the bugs to be fixed, have limited knowledge in how the code should be written.
Our project depends on carbon-component-react, and we at the same time have large number of priority items to address. So we can only allocate very little time to other projects (including the dependant project here). 

I understand that the experts in this repository project are often busy with other priority items or other priority projects and it seems there's not enough resources. 
For us who project depends on this, we dont have the time to learn in depth with this project's details, so we can only come in do a little bit here and there. 
So at this point, all I can say is hope the experts who know or are able to write these types of pieces can work on them sooner.
 
Maybe an expert can pair up with a contributor, and the expert can guide them this and that, basically give the shortcuts (shortcut in knowledge, not code shortcuts) so that contributor knows exactly what to put in. 
Maybe there can be clear details (template?) of how the code is or should be written, without needing to read many many articles.. I know there are many good articles out there that teaches good practices and etc, but code can be written differently by different projects/teams/people, and that takes A LOT of time to learn. Contributors won't be able to focus enough time to do that, so with someone like an expert and some template/details could help give the shortcut in knowledge to know exactly how to write it.

joshblack:
Thanks for the suggestions @David-Yuen! Definitely agreed.

joshblack:
Going to close this in favor of the discussion in #1285 👍

Issue #1033, opened 2018-06-27
feat(TextInput): Add Show/hide password for password inputs


AndySeymour2904:
Similar to this;
![image](https://user-images.githubusercontent.com/36265276/41973224-411a0dcc-7a0c-11e8-8424-710e13845a2d.png)

Quite a common UX feature. My suggestion would be to put an 'eye' icon in the side like is done above with carbon-icons 'icon--visibility-on' and only show password whilst this button is pressed>

Also add a new prop to TextInput to allow user customisation for this feature.

Happy to put this in a PR once agreed

asudoh:
@carbon-design-system/design Would this be something that we should maintain in core Carbon? Thanks!

tay-aitken:
I actually made designs around this a while back but since its a pattern we decided not to introduce it into the system just yet. Something we will hopefully be doing in the future!

AndySeymour2904:
@tay-aitken Is there a conversation or topic around why it was decided not to use this pattern?

As mentioned, I come to expect it of all user forms I use on the web nowadays.

tw15egan:
I think it was just a lack of dev resources. @tay-aitken, do we still have designs for this? Seems like @AndySeymour2904 said he was willing to put this in a PR if we can align on a design

tay-aitken:
It would sit in a PR for a while. We need to make a whole new patterns section on the site for this documentation. We currently only have components. It’s one we’ve been wanting to add but we haven’t had time to flesh out the new section.

tw15egan:
I think if someone is willing to create it, we can at least add it to the repo as a variation and prioritize documentation. I don't think this is a sufficient reason to _not_ build a component, especially if someone is willing to build it for us

tay-aitken:
I'm not saying we don't build, definitely think we should! I'm just saying we might not be able to surface it on the kit and website for a while. 

@AndySeymour2904 the only design requirements I have for it are:

- the `icon--visibility-off` and `icon--visibility-on` icons from the [carbon library](http://www.carbondesignsystem.com/style/iconography/library)are used
- the icons are 16px from the right side of the text input field
- that they follow our hover and active [guidelines for icons](http://www.carbondesignsystem.com/style/iconography/usage) > usage > visual usage > color

tw15egan:
Okay awesome, just want to make sure we encourage contributions 😄

[...]

PR #1344, opened 2018-09-20
feat(TextInput): add password visibility toggle for text fields

emyarod:
TODO:

* [x] ~~explore derived state vs separate component~~
* [x] ~~tests~~

---

Closes IBM/carbon-components-react#1033

This PR will add the password visibility toggle button to `<TextInput>`

Related: IBM/carbon-components#1058 and IBM/carbon-components#1060

#### Changelog

**New**

* `passwordVisibilityToggle` boolean prop on `<TextInput>`
* storybook example
* CSS selectors and data attributes needed for `carbon-components` styles
* tests for password visibility toggle `<TextInput>`

**Changed**

* bump `carbon-icons` and `carbon-components` versions
* convert `<TextInput>` into stateful class component only if  `passwordVisibilityToggle` is true, otherwise remain stateless component


src/components/TextInput/TextInput.js
```diff
@@ -1,104 +1,190 @@
...
-      aria-describedby={errorId}
-      className={textInputClasses}
-    />
-  ) : (
-    <input {...other} {...textInputProps} className={textInputClasses} />
-  );
+  togglePasswordVisibilityIconProps = () => {
+    const passwordIsVisible = this.state.type === 'text';
+    return {
+      alt: `${passwordIsVisible ? 'Hide' : 'Show'} password`,
```
asudoh: A couple of things from g11n perspective here:

1. Should be customizable via `props` (in case you haven't done this)
2. Should avoid string interpolation. Avoiding string interpolation seems easy given we can prepare different strings for different states

src/components/TextInput/TextInput.js
```diff
@@ -1,104 +1,190 @@
...
-    onClick: evt => {
-      if (!other.disabled) {
-        onClick(evt);
-      }
-    },
-    placeholder,
-    type,
+export default class TextInput extends React.Component {
+  state = {
+    type: this.props.type,
```
asudoh: Changing this component to stateful means that you need to take care of prop -> state sync (coping with change in `type` prop). Two ways to address them:

1. [`getDerivedStateFromProps()`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) (search for the method in our codebase to see how)
2. Keep this component stateless and let consumer to handle show/hide button's event to change `type` prop

src/components/TextInput/TextInput.js
```diff
@@ -1,104 +1,190 @@
...
+        {input}
+        {helper}
+        {error}
+      </div>
+    );
+  }
+}
 
 TextInput.propTypes = {
+  /**
```
asudoh: Great that you added prop comments! 👍

src/components/TextInput/TextInput.js
```diff
@@ -1,104 +1,190 @@
...
+    const invalidProps = {
+      'data-invalid': invalid,
+      'aria-invalid': invalid,
+      'aria-describedby': errorId,
+    };
+    const textInputProps = ({ invalid }) => ({
+      ...sharedTextInputProps,
+      ...(invalid ? invalidProps : {}),
+      ...{
+        'data-toggle-password-visibility':
```
asudoh: Seems that we don't need to create an object and spread it? Another question I have is what distinction you had in mind here vs. `sharedTextInputProps`, esp. given I see only one reference to that.

src/components/TextInput/TextInput.js
```diff
@@ -1,104 +1,190 @@
...
-      {label}
-      {input}
-      {helper}
-      {error}
-    </div>
-  );
-};
+    return (
+      <div
+        data-text-input={passwordVisibilityToggle || undefined}
```
asudoh: `data-text-input` is only for vanilla, not needed in React variant here.

src/components/TextInput/TextInput.js
emyarod: yeah I am still unsure about how I want to handle state with this component. Ideally I would like to keep this component stateless but I don't know if the consumer should be handling the `type` prop change. I think I would prefer to keep the `type` change local to `<TextInput>`, so I will go down the `getDerivedStateFromProps` route. What do you think @asudoh ?

src/components/TextInput/TextInput.js
emyarod: @asudoh I'm a bit confused about the string interpolation part. Would we expect users to provide strings for the two different button states?

alisonjoseph:
@emyarod is this still WIP or ready for review? Thanks! 😸

emyarod:
hi @alisonjoseph sorry I have been out sick the last few days but am still deliberating on how to proceed with this component along with @asudoh !

emyarod:
@asudoh I've opted for two variants of the password input component: one that is fully controlled and one that is fully uncontrolled. If the consumer finds the need to manipulate the input `type` themselves, they have the option to use `<TextInput.ControlledPasswordInput>`. Otherwise, consumers can use `<TextInput.PasswordInput>`

src/components/TextInput/util.js
```diff
@@ -0,0 +1,19 @@
...
+export const textInputProps = ({ invalid, sharedTextInputProps, errorId }) => ({
+  ...sharedTextInputProps,
+  ...(invalid ? invalidProps({ invalid, errorId }) : {}),
+});
+
+export const togglePasswordVisibilityIconProps = ({
+  passwordIsVisible,
+  alt,
+}) => ({
+  alt: alt || `${passwordIsVisible ? 'Hide' : 'Show'} password`,
```
asudoh: Can user configure those strings via props (for translation)?

src/components/TextInput/ControlledPasswordInput.js
```diff
@@ -0,0 +1,170 @@
...
+  const passwordIsVisible = type === 'text';
+  const input = (
+    <>
+      <input
+        {...textInputProps({ invalid, sharedTextInputProps, errorId })}
+        data-toggle-password-visibility={type === 'password'}
+      />
+      <button
+        className="bx--text-input--password__visibility bx--tooltip__trigger bx--tooltip--icon__bottom"
+        aria-label={alt || `${passwordIsVisible ? 'Hide' : 'Show'} password`}
```
asudoh: This portion needs to be from props for translation.

src/components/TextInput/PasswordInput.js
```diff
@@ -0,0 +1,180 @@
...
+          />
+        </button>
+      </>
+    );
+    const helper = helperText ? (
+      <div className="bx--form__helper-text">{helperText}</div>
+    ) : null;
+
+    return (
+      <div className="bx--form-item bx--text-input-wrapper">
```
asudoh: Can we render the stateless version here...?

src/components/TextInput/TextInput.js
```diff
@@ -1,102 +1,144 @@
...
-    onClick: evt => {
-      if (!other.disabled) {
-        onClick(evt);
-      }
-    },
-    placeholder,
-    type,
-  };
+export default class TextInput extends React.Component {
+  static PasswordInput = PasswordInput;
```
asudoh: Can we change `<TextInput>` back to functional given this is still stateless? For static properties here, I see a couple of choices:

1. `TextInput.PasswordInput`
2. Named exports (in [`index.js`](https://github.com/IBM/carbon-components-react/blob/v6.46.0/src/components/TextInput/index.js))

The latter is needed even if we take the former.

src/components/TextInput/PasswordInput.js
emyarod: did you mean render a ControlledPassword Input here?

asudoh: Yep!

joshblack:
@emyarod do you want to table this till post-X or is this something you feel strongly about bringing in before?

emyarod:
@joshblack I believe this is not an urgent feature so this can be kept on hold until post-X

joshblack:
Sounds good! Going to close this but definitely keep the branch around and then we can bring this back up post-X 😄 Want to make an issue for it just so we can assign it afterwards?

emyarod:
can we continue using the original issue https://github.com/IBM/carbon-components-react/issues/1033?

joshblack:
Oh duh, of course 👍

PR #1347, opened 2018-09-21
feat(build): introduce feature flags

### LINKED ISSUE #1339: Feature flags (React)
### Summary

This issue for introducing feature flag system, that has been introduced in vanilla repo recently, to this React repo.

### Justification

Provide the business reasons for this request.

A way to conditionally turn on/off specific set of features, without causing any additional (size) cost to the built version.

### "Must have" functionality

1. A way to build Storybook Sass code with different feature flags
2. Build-time flags, processed by [Webpack DefinePlugin](https://webpack.js.org/plugins/define-plugin/) or [`rollup-plugin-replace`](https://github.com/rollup/rollup-plugin-replace)

For Q3 new theme ffort.

asudoh:
Fixes #1339.

#### Changelog

**New**

* `CARBON_USE_EXPERIMENTAL_FEATURES` environment variable, used in Storybook environment, to enable experimental features (both CSS and JS) in Storybook (if `true` is set there)
* `featureFlags.js`, which defines feature flags
* `babel-plugin-transform-carbon-feature-flags.js`, an optional Babel plugin, to ensure that code of features disabled by feature flags, are DCE'd


joshblack:
Thanks for taking the time to do this! Could you go over your design decisions for why we should introduce the changes this way? Couldn't find them in the linked issue. Would be nice to have the simplicity of how React does feature flags: https://github.com/facebook/react/blob/master/packages/shared/ReactFeatureFlags.js

But I understand that part of the requirements here is proper DCE so I'm sure there are other factors at play 😄

asudoh:
@joshblack Thank you for reviewing, let me put a brief response to your question here. You seem to allude that this changeset is much larger than you expected, but in fact, the only change in (production) code is `featureFlags.js`, which looks exactly like what you expected. All other changes are build toolchain changes and documentation. That said, you may want to focus on reviewing `featureFlags.js` change and `.md` changes. Thanks!

shinytoyrobots:
I think there is a broad conversation about feature flag approach, and also a shorter term need for that functionality. Does this approach commit us in the long term, or can we pivot after our more immediate need to flag experimental components in the short term (after a team discussion/decision)?

asudoh:
@shinytoyrobots Good question - The only change that commit us in the long term is `featureFlags.js`, which is implemented as a set of named ESM exports of `const`s (as [brought up in issue 1339](
https://github.com/IBM/carbon-components-react/issues/1339#issuecomment-423247958)). All other (code) changes are toolstack additions/changes that can be changed any time without impacting consumers.

.storybook/webpack.config.js
```diff
@@ -1,8 +1,26 @@
 const path = require('path');
 
+const useExperimentalFeatures =
```
joshblack: Instead of modifying webpack config, could we define the feature flags in the scss file that we have?

scripts/README.md
```diff
@@ -0,0 +1,45 @@
+# `scripts` directory
+
+This directory contains some code to build stuffs.
+
+## `babel-plugin-transform-carbon-feature-flags.js`
```
joshblack: I might be misunderstanding here, but what is this transform doing? It feels like if the feature flag is false then it would be DCE'd by uglify but I could be wrong here.

src/featureFlags.js
```diff
@@ -0,0 +1,13 @@
+/**
```
joshblack: Should we name this `src/FeatureFlags` since it should be a singleton? Totally not necessary, just thinking out loud. Should this file also be included in the exports in `index.js`?

.storybook/webpack.config.js
asudoh: That can definitely be an option. The approach with WebPack config is to allow feature flag switch without modifying code.

scripts/README.md
asudoh: This Babel plugin allows you to replace feature flags with the given values. This is same as [WebPack's `DefinePlugin`](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin) in that sense, however, `DefinePlugin` works only with non-referenced variables, whereas our code using feature flags are written in a manner referring to ESM imports from `featureFlags.js`.

src/featureFlags.js
asudoh: We can do either way for former. For the latter - I'm thinking of reserving feature flags for internal usage to avoid addition of backward-compatibility concerns. We can move this file to `src/internal` in that sense.

scripts/README.md
joshblack: Totally! What's the difference in the resulting bundle when using the babel plugin versus going through a build process without it?

.storybook/webpack.config.js
joshblack: Definitely, which can totally be useful. I can go either way 👍

scripts/README.md
asudoh: Good question - Let's use the example in this document (assuming `aFeatureFlag` equals to `true` - focus on the child nodes in the `<div>`):

# In

```javascript
 import { aFeatureFlag } from '/path/to/featureFlags';
 ...
 const MyComponent = props => (<div {...props}>{aFeatureFlag ? 'foo' : 'bar'}</div>);
 ```

# Out _without_ the Babel plugin (Exactly same as "In")

```javascript
 import { aFeatureFlag } from '/path/to/featureFlags';
 ...
 const MyComponent = props => (<div {...props}>{aFeatureFlag ? 'foo' : 'bar'}</div>);
 ```

# Out _with_ the Babel plugin

```javascript
 import { aFeatureFlag } from '/path/to/featureFlags';
 ...
 const MyComponent = props => (<div {...props}>{'foo'}</div>);
 ```

Comment (asudoh) [reply scripts/README.md]: Also, you can see the difference by changing a component to use feature flags and looking at the resulting Rollup bundle (in our `umd` directory after you run `yarn build`)

Comment (asudoh) [reply scripts/README.md]: Went ahead with removing the Babel plugin for now.

File: README.md [hunk 4]
```diff
@@ -44,17 +44,19 @@ Please refer to the [Contribution Guidelines](./.github/CONTRIBUTING.md) before
 
 We recommend the use of [React Storybook](https://github.com/storybooks/react-storybook) for developing components.
 
-1. Start the server:
+1. (Optional) Set `true` to `CARBON_USE_EXPERIMENTAL_FEATURES` environment variable to test some of the experimental features.
+
```
alisonjoseph: Can we add the actual code sample here they would need to use, in case someone isn't familiar with environment variables. 

`CARBON_USE_EXPERIMENTAL_FEATURES=true yarn storybook`

asudoh: Done - Thanks @alisonjoseph for your feedback!

Issue #1001, opened 2018-06-14
NumberInput: Marked as invalid if the value supplied is ''


AndySeymour2904:
https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/NumberInput/NumberInput.js#L156

What happens in the situation where you have an optional NumberInput field?

It seems there is no way to do this, my suggestion is to remove the check and let the consumer make all the decisions about validity with the invalid flag provided

asudoh:
Thanks @AndySeymour2904 for reporting! Sounds that it makes sense - Let’s see what others in the team think.

enniel:
@asudoh Maybe better add property "strict" for this.

PR #1352, opened 2018-09-23
feat(NumberInput): strict mode property

enniel:
Fixes: #1001

#### Changelog

**New**
* `strict` boolean prop on `<NumberInput>`
* `strict` prop in `storybook` example `<NumberInput>`
* tests for checking showing invalidText message if strict mode disabled `<NumberInput>`


enniel:
@asudoh I think the best solution would be to remove the error warning when the value is empty, and allow the consumer to make all decisions about validity with an `invalid` flag, but think it might lead to errors in further work for some users. This solution will not break backward compatibility. And in the next version, you can completely remove the error warning.

tw15egan:
I'm good with this 👍

Issue #1369, opened 2018-09-28
Slider: Need prop -> state sync for value prop


asudoh:
## Detailed description

> Describe in detail the issue you're having.

Slider doesn't update its knob position even if I change `value` prop

> Is this issue related to a specific component?

Yes, `<Slider>`

> What did you expect to happen? What happened instead? What would you like to see changed?

Slider to react to change in `value` prop to update its internal state (and change knob position)

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

`v6.33.3`

## Steps to reproduce the issue

1. Add a Storybook knob for `value` in a similar manner as [`min`](https://github.com/IBM/carbon-components-react/blob/v6.33.3/src/components/Slider/Slider-story.js#L16)
2. Change the knob value

Result: No change in knob position.

asudoh:
Fixed by `v10` release.

PR #1370, opened 2018-09-28
fix(Slider): prop -> state sync for values

asudoh:
Given this is a breaking change, the fix is roped off under componentsX feature flag.

Refs #1369.

#### Changelog

**Changed**

* A new prop -> state sync code reacting to change in `value`, `min` and `max` props.


alisonjoseph:
Adding feature flags for breaking changes like this makes me a little nervous. It feels like it could easily get really complicated/messy. Thoughts @IBM/carbon-developers ?

joshblack:
@asudoh seems like we're changing this component from an un-controlled to controlled component with this change? I might be reading things wrong, so let me know if not 👍

asudoh:
@alisonjoseph Great point, let me share what we just discussed - The added code complexity is very valid concern. The plan to mitigate that is removing the old code a couple of weeks after new code being stable (new release), at latest. OTOH feature flag is a great way to replace the jet engine while still flying - Allowing us to make breaking changes without having to branch off the codebase.

@joshblack Good question, [slider has been a controlled component](https://github.com/IBM/carbon-components-react/blob/v6.38.0/src/components/Slider/Slider.js#L133-L135), and this PR doesn't change that.

joshblack:
Anything else you need to merge this in @asudoh? I say go for it if not! 🚀

Issue #1351, opened 2018-09-23
ExpandableTile doesn't render correctly when expanded by default.


lekterable:
## Detailed description

> Describe in detail the issue you're having.

ExpendableTile doesn't render correctly when expanded prop equals to true by default.
Text in BelowTheFoldContent is displayed but is overlaying the tile because the tile doesn't change its height. Expand icon is rotated correctly though.

> Is this issue related to a specific component?

ExpandableTile

> What did you expect to happen? What happened instead? What would you like to see changed?

I expected ExpandableTile with expanded prop equal to true to look the same as when it's expanded by clicking on it.

> What browser are you working in?

Chrome v.69 (giggity)

> What version of the Carbon Design System are you using?

Carbon Design System - 9.20
Carbon Components React - 6.32.4

> Please create a reduced test case in code sandbox
> https://codesandbox.io/s/jjjqnnq815

PR #1372, opened 2018-09-28
fix(ExpandableTile): set correct maxHeight for expanded by default tile

lekterable:
Closes IBM/carbon-components-react#1351

componentDidMount is setting maxHeight equal to the height of the non-expanded tile on the initial render but it doesn't get updated after the component receives props and because of that component height was incorrect if it was expanded by default.

#### Changelog

**New**

* componentDidUpdate event live cycle that sets the correct maxHeight value when the component receives initial props.

* new `expanded` prop in storybook


lekterable:
@asudoh 
1.) I initially tried to do that without using a new lifecycle method but was unable to. cDM() doesn't receive initial props and I don't think it needs an update unless there is a better way to do that. 
2.) Reusing that would be a good idea thanks!

lekterable:
I've reused the function but also refactored it a bit, I find this short version better. Let me know what you think.

PR #1376, opened 2018-10-01
feat(Search): add a searchLabel prop

lekterable:
Closes IBM/carbon-components-react#1361

I've added a searchLabel prop that you can use to set an aria-label for the Search component.

#### Changelog

**New**

* `searchLabel` prop


lekterable:
@asudoh do I understand correctly that you would like to use the existing prop labelText as a aria-label for the role="search" div?

asudoh:
@lekterable What I'm wondering if you could try is an approach with `aria-labeledby`. Does it make sense? Feel free to ping me otherwise - Thanks!

lekterable:
@asudoh I was trying to achieve the author's expected behavior:
> I would like to be able to pass an aria-label to the Search component.

But if this is a better approach I'm completely up for it, let me know what you think.

PR #1379, opened 2018-10-01
feat(TableToolbarSearch): placeholdertext and labeltext props

lekterable:
Closes IBM/carbon-components-react#999

Just a quick and intuitive way of changing placeholder and label in Search component rendered by TableToolbarSearch. If one wants to change it using `translateWithId` prop that still works.

#### Changelog

**New**

* `placeHolderText` and `labelText` props that are passed down to Search component.


asudoh:
Thanks @lekterable for jumping in! @joshblack Would you agree on the added flexibility for those props?

PR #1381, opened 2018-10-01
fix(Components): add initial state and remove empty state validation

lekterable:
There were still a couple components that didn't have its initial state set and using gDSFP() which would produce an error, and also a few had empty state validation that was unnecessary. Didn't include here components I took care of here https://github.com/IBM/carbon-components-react/pull/1377 and here https://github.com/IBM/carbon-components-react/pull/1373 but after those 3 are merged all components should be alright.

#### Changelog

**New**

* initial state

**Removed**

* Unnecessary empty state validation.


lekterable:
@asudoh you're right, good catch. Tooltip.js is my oversight. I will take a look at this error now.

lekterable:
I've investigated this error for a while and I think that it was caused by not saving defaultSelected value in the state. Earlier if the state was undefined on initial render this was handled by gDSFP(), good thing that tests caught that. What's interesting storybook example worked completely fine all the time because it had a `valueSelected=""` prop. I got worried about not breaking anything else so I've checked the rest of the updated components, but they all seem to work fine. Apparently, you can never be sure of anything in JavaScript no matter how small and trivial your change may seem.

PR #1397, opened 2018-10-04
fix(multi-select): add ability to add title tooltip on hover

tw15egan:
Closes https://github.com/IBM/carbon-components-react/issues/1393

Adds in the ability to add `title` to a `MultiSelectItem`

#### Changelog

**New**

- `title` prop to toggle whether or not a title is added

### DIFF
--- a/src/components/MultiSelect/MultiSelect.js
+++ b/src/components/MultiSelect/MultiSelect.js
@@ -63,7 +63,7 @@ export default class MultiSelect extends React.Component {
     /**
      *  Specify title to show title on hover
      */
-    title: PropTypes.bool,
+    useTitleInItem: PropTypes.bool,
 
     /**
      * `true` to use the light version.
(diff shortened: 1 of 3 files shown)


src/components/MultiSelect/MultiSelect.js
```diff
@@ -60,6 +60,11 @@ export default class MultiSelect extends React.Component {
      */
     type: PropTypes.oneOf(['default', 'inline']),
 
+    /**
+     *  Specify title to show title on hover
+     */
+    title: PropTypes.bool,
```
asudoh: I naïvely imagined prop named `title` is simply propagated as `title` attribute, which does not seem to be the case. Probably `useTitleInItem` or a like?

tw15egan: I see the confusion! I will change the prop name to `useTitleInItem`

Issue #1416, opened 2018-10-09
Composed modal does not close by clicking outside the modal


Greesy:
## Detailed description

Normal modals close by clicking outside the modal area, but Composed modals do not, see https://codesandbox.io/s/1o251voxl4 for example; I think ideally the ModalFooter's `OnRequestClose` would handle clicking outside the modal area, or the ComposedModal itself could have an `OnRequestClose` handler.

Also, semi-related, closing the modal with escape does not trigger the footer's `OnRequestClose` handler, but still closes the modal (though I can't seem to replicate this in codesandbox, the composed modal does not close at all by using escape).

> What browser are you working in?

Tested in Firefox and Chrome

> What version of the Carbon Design System are you using?

carbon-components: 9.26.1
carbon-components-react: 6.38.1

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

I'm working on an IMB internal-use portal, no pressing release dates.

## Steps to reproduce the issue

Check code sandbox:
> https://codesandbox.io/s/1o251voxl4

PR #1417, opened 2018-10-09
fix(#1416): Composed modal update

vpicone:
Closes IBM/carbon-components-react#1416

This change makes it so clicking outside of a composed modal closes the modal.


vpicone:
@asudoh ah yeah that’s right. I’ll update to more closely mimic the Modal component tomorrow.

tw15egan:
Thanks for fixing this! 💪

Issue #1408, opened 2018-10-09
Data table not supporting "short" style?


tw15egan:
@mlinnem commented on [Mon Oct 08 2018](https://github.com/IBM/carbon-components/issues/1223)

Data table component doesn't seem to support the "short" style. We use React so it might be specific to that. This is reported by a dev on our team (Justin Pomeroy) so I would follow up with him for any further clarification.

---

@tw15egan commented on [Mon Oct 08 2018](https://github.com/IBM/carbon-components/issues/1223#issuecomment-427952137)

Do you have an example of the `short` style? Not sure what that means

---

@mlinnem commented on [Tue Oct 09 2018](https://github.com/IBM/carbon-components/issues/1223#issuecomment-428272424)

Here's a reference to the style on the website:
<img width="800" alt="screen shot 2018-10-09 at 12 05 14 pm" src="https://user-images.githubusercontent.com/1547834/46685653-9c83b400-cbbb-11e8-9811-3f4a412eb0e7.png">

Here's the style in the kit:
<img width="1426" alt="screen shot 2018-10-09 at 12 04 57 pm" src="https://user-images.githubusercontent.com/1547834/46685672-a3aac200-cbbb-11e8-8dd0-b0dfd298747f.png">

tw15egan:
Need to add `short` version of data-table to React component

mohinderps:
From where do I take the reference for the short table?

mohinderps:
Will 'short' be a props to the DataTable ?

mohinderps:
If we accept the 'short' prop in DataTable, then to pass the prop down to Table component, We would need to write a Prop Getter function, I think. Let me know of a better approach to this.

mohinderps:
Hey @tw15egan , I have added the PR #1434 , so that you can review it and let me know if it needs to be done in some other way.

PR #1434, opened 2018-10-11
Short data table

mohinderps:
Closes IBM/carbon-components-react#1408

Added a support for a short data table. A new boolean prop 'short' is added to the Data Table component. That prop if passed to the component, created a short Data Table instead of default Data Table. The prop is not mandatory and the default value of the prop is false. 

#### Changelog

**Changed**

* Data-Table component accepts an optional prop 'short', which if set to true created a short data table.

### DIFF
--- a/src/components/DataTable/DataTable-story.js
+++ b/src/components/DataTable/DataTable-story.js
@@ -25,6 +25,22 @@ storiesOf('DataTable', module)
     )
   )
   .add(
+    'short',
+    withReadme(
+      readme,
+      withInfo({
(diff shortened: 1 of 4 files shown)


mohinderps:
Okay @asudoh, will do this.

asudoh:
Thanks @mohinderps! - Would you be able to update the snapshot, too? That may make the Travis test pass.

mohinderps:
Hey @asudoh, I have added knob and also update snapshot.

PR #1452, opened 2018-10-16
feat(ui-shell): add platform header

joshblack:
Closes IBM/carbon-components-react#1412

#### Changelog

**New**

* Add in initial selection of `Header*` components!

**Changed**

**Removed**


src/components/UIShell/PlatformHeaderNavMenu.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ChevronDownGlyph } from '@carbon/icons-react';
+
+const PlatformHeaderNavMenuItem = ({ title }) => {
+  return (
+    <li className="bx--platform-header__dropdown-item">
+      {title}
+      <ChevronDownGlyph className="bx--platform-header__arrow" />
+      <ul className="bx--dropdown-item__menu" role="menubar" aria-label={title}>
+        <li class="bx--dropdown-item__menu-item">
```
asudoh: Is this component for demonstration purpose? (Or will the content here be replaced with something else...?)

.storybook/_container.scss
```diff
@@ -1,3 +1,6 @@
+@import '~carbon-components/scss/globals/scss/feature-flags';
+
+$feature-flags: map-merge($default-feature-flags,  (ui-shell: true));
```
asudoh: I believe this line will be removed before the code is merged? (BTW [`export CARBON_USE_EXPERIMENTAL_FEATURES=true` will let you see the UI shell without changing code](https://github.com/IBM/carbon-components-react/blob/v6.40.0/.storybook/webpack.config.js#L69))

src/components/UIShell/PlatformHeaderNavMenu.js
joshblack: @asudoh honestly is just WIP code as I was copying over from `carbon-components`. I noticed that the snippet wasn't accessible though so I stopped working on it and made https://github.com/IBM/carbon-components/issues/1270 to tackle before finishing up work here 👍

joshblack:
@asudoh saw your comment, I left a TODO block when I made this WIP PR that I think addresses most of your notes 👍

.storybook/_container.scss
joshblack: Could we leave it so that the built storybook will have the UI Shell rendering correctly? Otherwise I can update the build process, I think. Just wanted to practice using the feature flags the way other teams would to be honest 👍

joshblack:
@alisonjoseph yeah! Can see the behavior described here: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html#kbd1_label

I believe the reason is that if all nav items were tabbable then there would be a lot of key strokes to go through to get through that navigation. This is one of the focus algorithms that they suggested as a result. Another tiny detail is that whatever the last focus item was in the navigation will have the active tab index, as well, so when revisiting the menu the tab should be on the last focused item.

alisonjoseph:
Thanks for the link! Totally makes sense especially for longer nav lists. and I noticed that it knows where you were before when you tab back! nice!

Issue #1458, opened 2018-10-17
Support custom column counts in the DataTableSkeleton


keithkade:
### Summary

We would like to support custom numbers of columns for the data table skeleton (like custom numbers of rows are currently supported). I'll open a PR soon. 

### Justification

When loading data for a table, we often know how many columns we will need before the data is loaded. By setting that number in the skeleton, the view will change less when the data is populated. 

### Desired UX and success metrics

Currently it always shows 5 columns. It would be nice to support arbitrary numbers (with some upper limit): 

ex: 3
![image](https://user-images.githubusercontent.com/2426829/47114768-629d5800-d222-11e8-8dd7-b433e60acaa4.png)

### "Must have" functionality

- DataTableSkeleton supports arbitrary column counts

### Specific timeline issues / requests

None

### Available extra resources

Me!

PR #1459, opened 2018-10-17
 feat(DataTableSkeleton): Support setting column count

keithkade:
Closes IBM/carbon-components-react#1458

Support for dynamic column counts in the DataTableSkeleton

#### Changelog

**New**

* `columnCount` property for the `DataTableSkeleton`


keithkade:
Is my commit log ok? Do I need to squash those merges?

PR #1475, opened 2018-10-24
feat(tools): add keyCode values and helper

joshblack:
Adds a `matches` helper and `keyCodes` constant for making it easier to match keyboard events. Right now it uses `keyCode`, in the future we should support `key` and then `keyCode` as a fallback 😄 

#### Changelog

**New**

* `keyCode` module and tests
  * `keyCodes` constant to be used for asserting keyboard event values
  * `matches` for checking if an event would match at least one provided key code

**Changed**

**Removed**

### DIFF
--- a/src/tools/__tests__/keyCode-test.js
+++ b/src/tools/__tests__/keyCode-test.js
@@ -1,0 +1,59 @@
+/**
+ * @jest-environment jsdom
+ */
+import { keyCodes, matches } from '../keyCode';
+
+// Used for the case of not matching
+const EQUAL_SIGN_KEY_CODE = 187;
+
+describe('keyCodes', () => {
+  describe('matches', () => {
+    let inputNode;
+    let eventListener;
+
+    beforeEach(() => {
+      inputNode = document.createElement('input');
+      inputNode.setAttribute('type', 'text');
+
+      eventListener = jest.fn();
+      inputNode.addEventListener('keydown', eventListener);
+
+      document.body.appendChild(inputNode);
+    });
+
+    afterEach(() => {
+      inputNode.removeEventListener('keydown', eventListener);
+      inputNode.parentNode.removeChild(inputNode);
+    });
+
+    it.each(Object.keys(keyCodes))('should match when using %s', key => {
+      const eventToMatch = new KeyboardEvent('keydown', {
+        keyCode: keyCodes[key],
+      });
+      inputNode.dispatchEvent(eventToMatch);
+      expect(matches(eventListener.mock.calls[0][0], [keyCodes[key]])).toBe(
+        true
+      );
+    });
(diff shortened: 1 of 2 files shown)


joshblack:
@asudoh from our dev channel, usage would be:

```js
How would you all feel about a helper for matching keycodes in events? Could be something like:

```handleOnKeyDown = event => {
  if (matches(event, [keyCodes.SPACE, keyCodes.ENTER])) {
    // ...
  }
}
```

Down to add in support for `key`, `which`, and `keyCode` 👍

joshblack:
Going to close this in favor of including it in the platform work with the suggested changes 👍

asudoh:
Great - Will be good if we can reduce the added layer (instead of added support for everything). Thanks!

joshblack:
What do you mean @asudoh? Just was going to update to:

```js
export const keys = {
  TAB: ['Tab', 9],
  ENTER: ['Enter', 13],
  ESC: ['Escape', 27],
  SPACE: [' ', 32],
  PAGEUP: ['PageUp', 33],
  PAGEDOWN: ['PageDown', 34],
  END: ['End', 35],
  HOME: ['Home', 36],
  LEFT: ['ArrowLeft', 37],
  UP: ['ArrowUp', 38],
  RIGHT: ['ArrowRight', 39],
  DOWN: ['ArrowDown', 40],
};

/**
 * Check to see if at least one key code matches the key code of the
 * given event.
 *
 * @example
 * import { keys, matches } from '../keys';
 * function handleOnKeyDown(event) {
 *   if (matches(event, [keys.ENTER, keys.SPACE]) {
 *     // ...
 *   }
 * }
 * @param {Event} event
 * @param {Array<number>} keyCodesToMatch
 * @returns {boolean}
 */
export function matches(event, keysToMatch) {
  for (let i = 0; i < keysToMatch.length; i++) {
    const [key, code] = keysToMatch[i];
    if (key === event.key || code === event.which || code === event.keyCode) {
      return true;
    }
  }
  return false;
}

```

asudoh:
AFAIK IE11 implements an older version of `KeyboardEvent.key` spec where several symbols are different from one in modern browsers. Just wanted team to be careful of creating a DOM abstraction API having to those kind of browser quirks in consideration tends to cause more amount of code than what people can imagine, as well as to affect the API cleanness, which reminded me of jQuery days (e.g. issue# 88 in [jQuery website repo](https://github.com/jquery/jquery.com)).

joshblack:
@asudoh so is the suggestion to use either `which` and fallback to `keyCode`, or just use `which`?

asudoh:
My suggestion is using `which`.

joshblack:
Cool! I'll do that 😄

PR #1492, opened 2018-10-29
Support a Warning type in modal wrapper

krithika-janakiraman:

Added a property in modal wrapper to support the warning type 

#### Changelog

**New**

Added a 'danger' property in modal wrapper. This is a boolean value that can be passed. If this is true, then the modal border and button appears in a red color.

**Changed**

Added a 'danger' property in modal wrapper

**Removed**

Nothing was removed from the existing modal wrapper code.


alisonjoseph:
Is this PR needed? Seems that the danger prop is already handled on the Modal component?

joshblack:
Going to close this due to inactivity. Feel free to comment and we can re-open!

Issue #916, opened 2018-05-24
Footer Create Button not required


ravinderpadigela:
Hi,
Currently I am using Footer, but I observed one 'Create' button is coming by default, I am unable to restrict it using Footer component of react-carbon. We dont require this Create button for our project functionalities. I alredy contacted in slack reagarding the same. So please help us to remove/restrict this create button option. Thanks

joshblack:
Hi @ravinderpadigela! 👋 Thanks so much for taking the time to make this issue 🙂 Just a heads up, we have an issue template that we recommend folks to follow in order to make sure that we can address the issue faster. Filling that out would help us out a ton here! You would have seen it as the text when first creating this issue 👍

emyarod:
[wouldn't using `props.children` remove the call to action button?](https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Footer/Footer.js#L24-L28)

alisonjoseph:
Closing this due to inactivity.

PR #1493, opened 2018-10-29
Footer's create button that comes by default has been disabled

krithika-janakiraman:

The create button that comes by default in the footer has been disabled. It can be enabled by passing the 'buttonText' prop if the user wishes to show the button with any label.

#### Changelog

**New**

Nothing new has been added from the footer functionality point of view.

**Changed**

Disabled the 'create' button as a default prop. 

**Removed**

The 'buttonText: 'create'' has been removed as a default prop and is optional


krithika-janakiraman:
Hi @asudoh! :)  
This change is related to the issue https://github.com/IBM/carbon-components-react/issues/916. In the footer, the "buttonText" prop is "create" by default. So in case someone doesn't want the create button appearing in the footer, they are unable to disable it .
The change i made in this PR is to remove the buttonText from default props. Incase the user passes a buttonText prop, the button would appear, else there wont be any button by default.

asudoh:
Hi @krithika-janakiraman thank you articulating what you tried to do! Have you tried what @emyarod suggested at https://github.com/IBM/carbon-components-react/issues/916#issuecomment-395877272?

joshblack:
Going to close this due to inactivity. Feel free to comment and we can re-open!

Issue #1100, opened 2018-07-18
ComboBox bugs and missing light prop


nicoxxie:
## Detailed description

> Describe in detail the issue you're having.
1) Cancel button only shown when dropdown is expanded
![screen shot 2018-07-18 at 11 03 01 am](https://user-images.githubusercontent.com/17915105/42893536-344f3d8c-8a7a-11e8-810a-d8d26e342855.png)

2) If an option is already selected, the dropdown list only contains that one item
![screen shot 2018-07-18 at 11 03 06 am](https://user-images.githubusercontent.com/17915105/42893544-3c446bac-8a7a-11e8-9ed2-be7be87f9851.png)

![screen shot 2018-07-18 at 11 03 15 am](https://user-images.githubusercontent.com/17915105/42893553-4549ba7c-8a7a-11e8-9bb3-836f1f587e83.png)

3) missing 'light' option

> Is this a feature request (new component, new icon), a bug, or a general issue?
bug

> Is this issue related to a specific component?
ComboBox

> What did you expect to happen? What happened instead? What would you like to see changed?
1) Cancel button shown even if dropdown is not expanded
2) dropdown shows list of all options, with the current selected value highlighted.
3) light background

> What browser are you working in?
all

> What version of the Carbon Design System are you using?
  "carbon-components": "9.1.20",
    "carbon-components-react": "6.12.2",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
ibm cloud

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

* Screenshots or code
* Notes

lovemecomputer:
@IBM/carbon-designers the "bug" described here is the intentional UX, though perhaps it looks a little confusing when an item is selected? it appears the UX of this component is functioning according to spec.

nicoxxie:
@lovemecomputer If the user wants to select a value, let's say the default value is 'No access'
The user has to: 
1. Click the dropdown (so the x button is shown)
2. Click the x (then dropdown shrunks)
3. Click the dropdown again
3. Select the new value

If the user makes a mistake, he/she has to repeat the step again.  
Not a ux pro here but this doesn't seem very user friendly.  You are right it's not a 'bug' *bug*, but should be *improvement* bug

lovemecomputer:
@nicoxxie thanks for the detail! 

[...]

PR #1502, opened 2018-10-30
fix(#1100): when there is a selected value, show cancel button and all the elements on expand

kfabiank:

fix(#1100): when there is a selected value, show cancel button and all the elements on expand

Issue URL: https://github.com/IBM/carbon-components-react/issues/1100

#### Changelog

src/components/Combobox.js
src/components/Combobox-test.js

**Changed**

src/components/Combobox.js
src/components/Combobox-test.js

**Removed**

Nothing removed.


src/components/ComboBox/ComboBox.js
```diff
@@ -234,13 +245,12 @@ export default class ComboBox extends React.Component {
                   onKeyDown: this.handleOnInputKeyDown,
                 })}
               />
-              {inputValue &&
-                isOpen && (
```
asudoh: Is the removal of `isOpen` here intentional?

src/components/ComboBox/ComboBox.js
```diff
@@ -26,6 +22,18 @@ const getInputValue = (props, state) => {
   return state.inputValue || '';
 };
 
+const findHighlightedIndex = ({ items, itemToString }, inputValue) => {
+  if (!inputValue) {
+    return -1;
+  }
+
+  return items.findIndex(item =>
```
asudoh: Seems that `Array.prototype.findIndex()` is not in [our polyfills list](https://github.com/IBM/carbon-components-react/blob/v6.41.0/.storybook/polyfills.js). We consider requirement of new polyfill as a breaking change. Do you want to in-line a (lightweight) polyfill or seek for an alternate implementation here?

joshblack: Good catch @asudoh! 🙏

src/components/ComboBox/ComboBox.js
joshblack: Good catch! Thanks @asudoh 🙏

kfabiank: Yes, it's intentional. By removing it, we show the clear selection option not only when the ComboBox is opened.

src/components/ComboBox/ComboBox.js
kfabiank: Fixed: I just updated this here https://github.com/IBM/carbon-components-react/pull/1502/commits/3ffd1406ce371ab2f17366396a0e49ad8a479ead

alisonjoseph:
Thanks for adding this! Can you update the PR name so that it will trigger semantic release when merged in?  either `fix:` or `feat:`

joshblack:
It seems like there might be a slight UX change with this diff, namely when clicking back into the input after clearing the selection I believe the menu should stay open. Quick video:


[Screen Recording 2018-11-01 at 9.47.00 AM.zip](https://github.com/IBM/carbon-components-react/files/2538933/Screen.Recording.2018-11-01.at.9.47.00.AM.zip)

kfabiank:
Hi @joshblack,

it looks like it's also happening in production. Attaching screencast.
To fix it, I created a ref to the input, compared if the clicked element is the input and if it's already opened, tell downshift not to execute the default event.

[screencast.zip](https://github.com/IBM/carbon-components-react/files/2539884/screencast.zip)

joshblack:
Oh whoa, thanks for the catch @kfabiank!

src/components/ComboBox/ComboBox.js
```diff
@@ -175,6 +199,13 @@ export default class ComboBox extends React.Component {
     );
   };
 
+  onToggleClick = isOpen => event => {
+    if (event.target === this.textInput.current && isOpen) {
```
joshblack: Is there a way we could do this without dropping down into downshift implementation details and without persisting the event?

src/components/ComboBox/ComboBox.js
```diff
@@ -26,6 +22,28 @@ const getInputValue = (props, state) => {
   return state.inputValue || '';
 };
 
+const findHighlightedIndex = ({ items, itemToString }, inputValue) => {
+  if (!inputValue) {
+    return -1;
+  }
+
+  let i = 0;
```
joshblack: What do you think about a `for` loop variant of this:

```suggestion
  const searchValue = inputValue.toLowerCase();
  
  for (let i = 0; i < items.length; i++) {
    const item = itemToString(items[i]).toLowerCase();
    if (item.indexOf(searchValue) !== -1) {
      return i;
    }
  }

  return -1;
```

This can help drop our usage of `includes`, as well.

src/components/ComboBox/ComboBox.js
kfabiank: I tried a couple of other approaches before getting to this one because none of them worked as good as this one.
The first one and the most obvious was to open the menu in the input's onFocus event. It worked with a timeout, which doesn't look good because it makes a weird jump.
Also, tried using Downshift's stateReducer prop, but there is no way to know if the state change comes from the input or the button or the clear selection.

Please let me know if you have any suggestion I can try/implement to avoid persisting the event and getting into Downshift details.

src/components/ComboBox/ComboBox.js
kfabiank: Better!
Thanks.

nicoxxie:
Hey, any update on this PR?

Issue #1495, opened 2018-10-29
Modify DropdownItem to accept children in place of itemText


justinpurcell1:
### Summary

Instead of itemText, which can only be styled one way, update DropdownItem such that it can render multiple child components which are styled differently. 

Eg:

   Messages _3 unread_

### Justification

I would like to use this feature in an upcoming project.

### Desired UX and success metrics

The user will be able to access more information in a compact area if text is bolded, colored differently, etc.

### "Must have" functionality

DropdownItem must be able to accept and render children when itemText is not supplied.

### Specific timeline issues / requests

Within two weeks would be ideal.

### Available extra resources

I can write the feature if needed.

asudoh:
Hi :wave: thank you for reporting! Sounds a good change, and a PR is welcome as you alluded!

Issue #1364, opened 2018-09-26
Revert DropdownV2 requirements, support both signatures for Dropdowns


joshblack:
Related to #1228, we want to revert the requirements of DropdownV2 from requiring a data structure of items to render and bring back the ability to render DropdownItem's inside of DropdownV2.

justinpurcell1:
Getting my bearings here. 

I had initially opened issue #1495 to extend the functionality of Dropdown. However, after seeing the deprecation warnings, I created PR #1505 to allow DropdownV2 to accept a function to map items to components. After getting some feedback, I adjusted it to make it more generic, and it mostly works with the current implementation of DropdownItem. It also will also accept other custom components in addition to the existing behavior with strings and objects. 

@tomkdgun does this fulfill your use cases?
@joshblack is this cohesive with existing patterns within the project?

joshblack:
Thanks so much for taking this work on so far @justinpurcell1!

I think something that might not be clear with this issue title is that it'd be great for us to support these two patterns:

```jsx
<Dropdown items={...} itemToString={...} />

<Dropdown>
  <DropdownItem />
  <DropdownItem />
  <DropdownItem />
</Dropdown>
```

If it's unrealistic to have one component have both implementations, that's totally fair. There was an idea that we could use Downshift to power all of our dropdowns but I think that the flexibility of controlling dropdown content is too important for this specific component (or other dropdown variants).

I think the real question becomes how can we leverage `downshift` while providing it's render prop surface area, or should we just try and build these components to be as easy-to-use as possible and be okay with restricting customizability in dropdown items.

ZacharyStair:
@joshblack is the benefit of the second solution that people can use the literal `DropdownItem` components? or that they can mix and match what components they pass as children to the `DropdownV2`? or is it just for ease of use, and would only accept `DropdownItem`s?

Justin's solution could be modified to accept a signature that looks like the latter example, the Dropdown with DropdownItem children: basically using a getter to select either `this.props.items || React.Children.toArray(this.props.children)` as items, then letting the default itemToString get `item.props.itemText` as appropriate. It isn't the ugliest solution but it makes the code a bit more complex.

[...]

PR #1505, opened 2018-10-30
feat(DropdownV2): allow components as items

justinpurcell1:
Closes IBM/carbon-components-react#1495

Modify DropdownV2 to render items as components instead of strings if desired. Note that the original issue specified DropdownItem, but the change has been applied to DropdownV2 since the old Dropdown is deprecated.

#### Changelog

**Changed**

* DropdownV2 now accepts an `itemToElement` prop and prefers that over `itemToString`


src/components/DropdownV2/DropdownV2-story.js
```diff
@@ -88,7 +116,8 @@ storiesOf('DropdownV2', module)
       `,
     })(() => (
       <div style={{ width: 300 }}>
-        <DropdownSkeleton />&nbsp;
+        <DropdownSkeleton />
+        &nbsp;
```
justinpurcell1: This looks like a side effect of prettier. The commit hooks also changed the bullets in .md files and added various newlines to .js files.

joshblack: Hi @justinpurcell1! Most likely this is due to using something other than `yarn` to manage dependencies (e.g. prettier version is different locally than what the project uses).

joshblack:
Hi @justinpurcell1! 👋 Thanks so much for taking the time to make this PR 😄 

`DropdownV2` is a tricky one. The requirements moving from Dropdown, DropdownItem towards the new API may be reverted as we're noting more and more valid use-cases (like this one) coming up as folks are transitioning. We have this open issue to potentially handle both signatures: https://github.com/IBM/carbon-components-react/issues/1364

Might be worthwhile having this conversation over under that to see if this could help out a lot of these use-cases that people are noting 👍

justinpurcell1:
@joshblack this can serve as a jumping off point from which to modify either DropdownV2 or DropdownItem. With this change it has most of the functionality but has some issues due to the click handling in DropdownItem

joshblack:
> @joshblack this can serve as a jumping off point from which to modify either DropdownV2 or DropdownItem.

As a general policy, we try and have conversations in issues before Pull Requests (if applicable) so it would be great to start off over in that issue 👍

justinpurcell1:
@jnm2377 yeah, that style issue shouldn't be too bad to change in a real implementation. My biggest concern was the difference in click handling and selection between Dropdown and DropdownV2

justinpurcell1:
@jnm2377 is there anything else I should clean up? Are there any more steps required to get this merged in?

jnm2377:
No, everything looks good. My apology for the delays. Thanks for the contribution @justinpurcell1

src/components/DropdownV2/DropdownV2.js
```diff
@@ -141,7 +154,7 @@ export default class DropdownV2 extends React.Component {
                     isActive={selectedItem === item}
                     isHighlighted={highlightedIndex === index}
                     {...getItemProps({ item, index })}>
-                    {itemToString(item)}
+                    <ItemToElement key={itemToString(item)} {...item} />
```
asudoh: HI @justinpurcell1 👋 when `item` is a string, the spread operator here seems to make it an array of characters (ref: https://github.com/IBM/carbon-components-react/issues/1586). Do you want to use this render-element approach here only if `itemToElement` is given?

PR #1510, opened 2018-11-01
feat(NumberInput): optionally hide the label, resolves #1509

### LINKED ISSUE #1509: NumberInput: add hideLabel prop
### Summary
Add a hideLabel prop to the NumberInput component. This prop is provided on many other components and makes the component more consistent with the rest of the repo.

### Justification
Need this functionality.

### Available extra resources
I will submit a PR for this.

McBrosa:
Closes IBM/carbon-components-react#1509

Add hideLabel prop to NumberInput component.

#### Changelog

**NumberInput**
* add hideLabel prop

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -17,6 +17,7 @@ export default class NumberInput extends Component {
   static propTypes = {
     className: PropTypes.string,
     disabled: PropTypes.bool,
+    hideLabel: PropTypes.bool,
     iconDescription: PropTypes.string.isRequired,
     id: PropTypes.string.isRequired,
     label: PropTypes.node,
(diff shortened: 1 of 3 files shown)


src/components/NumberInput/NumberInput.js
```diff
@@ -216,3 +181,49 @@ export default class NumberInput extends Component {
     );
   }
 }
+
+NumberInput.defaultProps = {
```
asudoh: Any reason why you moved `defaultProps`/`propTypes`?

McBrosa: Reverted that change, for some reason I thought that was how the reset of the components were setup.

joshblack:
@asudoh @tw15egan which prop convention are we following for boolean props again? Should it be `shouldHideLabel`?

asudoh:
> which prop convention are we following for boolean props again? Should it be shouldHideLabel?

Thank you for asking @joshblack! I don't think we have a particular naming rule for boolean props in general, I see other components use `hideLabel` for similar purposes.

PR #1511, opened 2018-11-01
docs: add CodeSandbox example with consistent URL

emyarod:
This PR will bring the code from the CodeSandbox starter example into the `examples` directory so that CodeSandbox can point to a consistent GitHub URL and stay up to date automatically. Rather than using URLs like https://codesandbox.io/s/r4zxq9pvq or https://codesandbox.io/s/x2mjypo6pp to point to updated sandboxes with each release, the documentation will always point to https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/helloworld

The URLs will go live when this PR is merged in, but an example of what the change will look like can be seen here: https://codesandbox.io/s/github/emyarod/carbon-components-react/tree/updated-codesandbox-example/examples/helloworld

#### Changelog

**New**

* starter code for example application in `examples` directory

**Changed**

* update URLs to CodeSandbox playground in GitHub issue templates


emyarod:
@asudoh regarding the suggestion to create a similar example for the vanilla repo: isn't this similar to the CodePen examples we have on the https://www.carbondesignsystem.com/ website?

.github/ISSUE_TEMPLATE.md
```diff
@@ -23,15 +23,15 @@
 ## Steps to reproduce the issue
 
 <!--
-Please try to re-create the issue using our Codesandbox template available here:
-https://codesandbox.io/s/x2mjypo6pp
+Please try to re-create the issue using our CodeSandbox template available here:
+https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/helloworld
```
joshblack: Could we do this in a subsequent PR just to make sure it works before switching the issue template?

emyarod: We can create a new PR for changing the issue template, but it should be identical to https://codesandbox.io/s/github/emyarod/carbon-components-react/tree/updated-codesandbox-example/examples/playground, except the new link will point to `.../IBM/carbon-components-react/tree/master/...` instead

but I've reverted the URL change in the issue template and changed the name from `helloworld` to `playground`

asudoh:
> regarding the suggestion to create a similar example for the vanilla repo: isn't this similar to the CodePen examples we have on the https://www.carbondesignsystem.com/ website?

Great question @emyarod! While our CodePen example works well for quick running for components, CodeSandbox is easier for using ESM and hosting examples with granular component lifecycle (e.g. instantiation/release). Given that, I was thinking of using something similar to your React CodeSandbox ref as a template of reduced test cases, as the React one is often used for that purpose. Does this make sense? Don't hesitate to ping me otherwise - Thanks!

.github/ISSUE_TEMPLATE.md
joshblack: Totally agreed that it should be identical, just want to manually verify before changing the issue template just as a sanity check 👍

joshblack:
Since this is going into our `examples` folder, can we add a `README` that communicates what it is for? I think naming the folder codesandbox might also reduce confusion for folks who might think it is a literal playground for them to pull down and use.

asudoh:
@emyarod Is this ready for another review...? Thank you for all the updates!

emyarod:
yes @asudoh this is ready for review

asudoh:
Thanks @emyarod - @joshblack Could you take another look...? Thanks!

emyarod:
yeah that makes sense, I've renamed the directory to `codesandbox`

Issue #1334, opened 2018-09-18
Unexpected behavior in keyboard navigation in FileUploaderButton


ColbyJohnIBM:
## Detailed description

> Describe in detail the issue you're having.

When using keyboard navigation, the FileUploaderButton gets focus twice.

> Is this issue related to a specific component?

Yes, FileUploaderButton.

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

6.28.3

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Watson Compare & Comply Tooling UI, no extremely pressing deadline

## Steps to reproduce the issue
1. Navigate to the FileUploaderButton Storybook story. (http://react.carbondesignsystem.com/?knob-The%20title%20%28title%29=Section%201%20title&knob-Button%20kind%20%28buttonKind%29=&knob-Label%20text%20%28labelText%29=Add%20files&knob-Form%20item%20name%3A%20%28name%29=&knob-Supports%20multiple%20files%20%28multiple%29=true&knob-ARIA%20role%20of%20the%20button%20%28role%29=&knob-Tab%20index%20%28tabIndex%29=-1&selectedKind=FileUploader&selectedStory=FileUploaderButton&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)
2. Press "tab" key. Note that element `div.bx--file.bob` receives focus. Note that pressing enter pulls up the file upload dialog box. (This is all expected.)
3. Press "tab" key again. Note that element `label.bx--btn.bx--btn--primary` receives focus. (Not expected, as this essentially focuses on the same button.) Note that pressing enter pulls up the file upload dialog box. Regardless of what is done in that dialog box, no action is logged. Note that a second dialog box then opens, which then behaves normally.

joshblack:
Re-opening due to feedback above ^

joshblack:
When is the `label` receiving focus? Having a hard time seeing where that happens.

We definitely want to maintain the initial focus target as the button, currently a `div` but ideally we'd have it be a native `button`.

A strategy here might be to have a node inside of the `div` for describing it and we can add `aria-labelledby` on the outermost div to update it's label when the content changes (e.g. a file is uploaded).

emyarod:
the label receiving focus was due to me modifying the storybook knobs in my browser, my mistake.

[...]

PR #1515, opened 2018-11-05
fix(FileUploader): remove label tabIndex

emyarod:
Closes IBM/carbon-components-react#1334

The FileUploaderButton was allowing users to focus both on the wrapper `div` and the `label` using the keyboard

#### Changelog

**Changed**

* moved the button styles off of `<FileUploaderButton>`'s `label` and placed it on the wrapper `div` instead
* set the value of the `label`'s `tabIndex` to `null` if the prop value is falsy so that it won't be focusable

### DIFF
--- a/src/components/FileUploader/FileUploader.js
+++ b/src/components/FileUploader/FileUploader.js
@@ -130,7 +130,7 @@ export class FileUploaderButton extends Component {
       ...other
     } = this.props;
     const classes = classNames({
-      'bx--file': true,
+      'bx--btn': true,
       [className]: className,
     });
 
@@ -140,15 +140,14 @@ export class FileUploaderButton extends Component {
       <div
         role="button"
         tabIndex="0"
-        className={classes}
+        className={`bx--btn--${buttonKind} ${classes}`}
         onKeyDown={evt => {
           if (evt.which === 13 || evt.which === 32) {
             this.input.click();
           }
         }}>
         <label
-          className={`bx--btn bx--btn--${buttonKind}`}
-          tabIndex={tabIndex}
+          tabIndex={tabIndex || null}
           htmlFor={this.uid}
           role={role}
           {...other}>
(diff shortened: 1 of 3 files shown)


emyarod:
hi @asudoh after discussing with @joshblack it seems that part of the reason for assigning the `role="button"` to the wrapper `div` was for a11y. I am not against having the `<label>` act as a button, but I'd like some more input on whether or not this violates any a11y rules

joshblack:
Copy/pasting from our slack conversation:

> Keyboard handlers should only be placed on elements with an interactive role. Things like label aren’t considered interactive because it is textual content that one would read but not really interact with. You also can’t switch the default role of label to an interactive one (like button) without then losing the label for the input (which is required)

src/components/FileUploader/FileUploader.js
```diff
@@ -130,7 +130,7 @@ export class FileUploaderButton extends Component {
       ...other
     } = this.props;
     const classes = classNames({
-      'bx--file': true,
+      'bx--btn': true,
```
asudoh: Would you want to add `bx--btn--${buttonKind}` here so you don't have to do ``className={`bx--btn--${buttonKind} ${classes}`}``?

src/components/FileUploader/FileUploader.js
```diff
@@ -140,15 +140,14 @@ export class FileUploaderButton extends Component {
...
+        className={`bx--btn--${buttonKind} ${classes}`}
         onKeyDown={evt => {
           if (evt.which === 13 || evt.which === 32) {
             this.input.click();
           }
         }}>
         <label
-          className={`bx--btn bx--btn--${buttonKind}`}
-          tabIndex={tabIndex}
+          tabIndex={tabIndex || null}
```
asudoh: Given I believe the original `tabIndex` prop is for the button, would you want to move it to the container `<div>`?

emyarod: I'm not sure what the original `tabIndex` prop was for but I've gone ahead and made that change since it makes more sense than to put it on the `label`

PR #1521, opened 2018-11-06
Upgrade storybook

vpicone:
## THE SEQUEL
Closes Issue #1496 

This upgrades storybook and deps to v4.04 which allows for Webpack & babel upgrades, theming. Storybook v4 also offers a backward-compatible parameters API which was implemented in this PR using the withInfo addon.

### Changes from PR https://github.com/IBM/carbon-components-react/pull/1498:

* It looks like webpack has some real haziness in their documentation around sideEffects: https://github.com/webpack/webpack/issues/6571. It's my impression they consider scss imports to be side effects (debatable?) and drop them in production.

* Adding the `sideEffect: true` declaration to our scss rule fixed the static build issues.

#### New
* Added direct dep on webpack to meet peer dep reqs

#### Changed
* Updated all of the @storybook deps to 4.0.4
* Changed loaders to be webpack 4 compatible.


vpicone:
@asudoh Nice catch! Looks like there are some really interesting new options for source-maps in webpack v4 using the [devtool](https://webpack.js.org/configuration/devtool/) preference. I believe the option I selected most closely mimics the current setup.

Issue #1481, opened 2018-10-26
Experimental Component - Structured List 


aledavila:
### Summary

Add support for markup change in structured list for v10

Design:
https://github.com/IBM/carbon-components/issues/1275

Dev:
https://github.com/IBM/carbon-components/pull/1344

### Specific timeline issues / requests

v10 release

PR #1528, opened 2018-11-06
feat(StructuredList): add experimental structured list

emyarod:
Closes IBM/carbon-components-react#1481

Add flag to allow experimental structured list

#### Changelog

**Changed**

* conditionally render experimental component based on feature flag

### DIFF
--- a/src/components/StructuredList/StructuredList-story.js
+++ b/src/components/StructuredList/StructuredList-story.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-
 import { iconCheckmarkSolid } from 'carbon-icons';
 import Icon from '../Icon';
 import {
@@ -12,6 +11,7 @@ import {
   StructuredListCell,
 } from '../StructuredList';
 import StructuredListSkeleton from '../StructuredList/StructuredList.Skeleton';
+import { componentsX } from '../../internal/FeatureFlags';
 
 storiesOf('StructuredList', module)
   .add(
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
     "prettier:diff": "prettier --list-different \"**/*.{scss,css,js,md}\"",
     "semantic-release": "semantic-release",
     "start": "yarn storybook",
+    "start:experimental": "cross-env CARBON_USE_EXPERIMENTAL_FEATURES=true CARBON_REACT_STORYBOOK_USE_EXTERNAL_CSS=true CARBON_REACT_STORYBOOK_USE_STYLE_SOURCEMAP=true yarn start",
```
asudoh: If this is (solely) for running Storybook with experimental mode, the only environment variable change needed is `CARBON_USE_EXPERIMENTAL_FEATURES=true`.

package.json
```diff
@@ -18,6 +18,7 @@
     "prettier:diff": "prettier --list-different \"**/*.{scss,css,js,md}\"",
     "semantic-release": "semantic-release",
     "start": "yarn storybook",
+    "start:experimental": "cross-env CARBON_USE_EXPERIMENTAL_FEATURES=true yarn start",
```
joshblack: Instead of adding as a script can we just add it to a md file under `docs`? This will be more of a temporary thing so would rather it live there instead of code.

package.json
```diff
@@ -116,7 +117,7 @@
     }
   },
   "peerDependencies": {
-    "carbon-components": "^9.21.0",
+    "carbon-components": "^9.46.0",
```
joshblack: Could we keep the peer dependency as the old one, but update the devDependency? I don't think we want to update peer quite yet for experimental work (unless others feel strongly about this).

package.json
alisonjoseph: Hmm, could we leave it here for now and remove it later? This makes it a lot easier while developing. Yes its "temporary" but still will be around for ~6 months or so most likely.

emyarod: yeah I initially added it for ease of development and would like to keep it for the time being, but I can also create a local gitignored script to start the experimental environment if that would be better

emyarod: @joshblack @alisonjoseph I've removed this change from this PR and created a separate PR for it here https://github.com/IBM/carbon-components-react/pull/1547

Issue #1516, opened 2018-11-05
DataTableSkeleton results in TypeError: Object doesn't support property or method 'keys' on IE11


austinbruch:
## Detailed description

> Describe in detail the issue you're having.

Rendering `DataTableSkeleton` on IE11 is resulting in an error: 
```
TypeError: Object doesn't support property or method 'keys'
```
Looking at the source, it appears that `Array.prototype.keys` is called, but that is not supported in IE11. 

> Is this issue related to a specific component?

Yes, `DataTableSkeleton`.

> What did you expect to happen? What happened instead? What would you like to see changed?

DataTableSkeleton should render, but an error is thrown instead.

> What browser are you working in?

IE11

> What version of the Carbon Design System are you using?

6.45.2

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM Cloud Identity

## Steps to reproduce the issue

1. Include a `DataTableSkeleton` in your react app's rendered output.

> Please create a reduced test case in code sandbox
> https://codesandbox.io/s/r4zxq9pvq

I cannot get codesandbox to work on IE11, so I can't produce a reduced test case. Appears there's some issue with codesandbox itself.

## Additional information

I suspect this was introduced in this commit: https://github.com/IBM/carbon-components-react/commit/63543c476710ec0a8f30ed265c0bf9f7ac5d66bb#diff-3055cc730be6e994e8d2d9abe2938111 with the goal of supporting the `columnCount` prop.

The [list of required polyfills](https://github.com/IBM/carbon-components-react/blob/master/.storybook/polyfills.js) for older browsers like IE11 does not seem to satisfy this issue. 

I think the list of required polyfills should be updated to account for this new requirement, or the `DataTableSkeleton` code should be updated to achieve the goal of supporting `columnCount` using some other mechanism.

austinbruch:
Following the pattern for polyfill inclusion from `core-js`...
Adding `import 'core-js/modules/es6.array.iterator';` seems to resolve the issue.

The `Array.iterator` module  includes the `keys` prototype function, among other things, per this issue: https://github.com/zloirock/core-js/issues/421

You might want to consider including the above module in the list of required polyfills mentioned in the original issue.

joshblack:
Thanks for the report, @austinbruch! Most likely we should just re-write this to not use `keys()` from `Array.prototype`.

PR #1531, opened 2018-11-07
IE11 DataTableSkeleton fix

slaytanic:
Closes IBM/carbon-components-react#1516

I replaced the use of Array.prototype.keys with .map for IE11 compatibility.

#### Changelog

**Changed**

* DataTableSkeleton: Replace use of .keys for .map
* README: Add myself as a contributor


src/components/DataTableSkeleton/DataTableSkeleton.js
```diff
@@ -18,7 +18,7 @@ const DataTableSkeleton = ({
 
   const rows = [];
   const rowRepeat = rowCount - 1;
-  const columnsArray = Array.from(Array(columnCount).keys()); // [0,1,2...columnCount-1]
+  const columnsArray = Array.from(Array(columnCount).map((i, idx) => idx)); // [0,1,2...columnCount-1]
```
joshblack: Could we update this to:

```suggestion
  const columnsArray = Array.from({ length: 5 }, (_, index) => index)
```

slaytanic: I changed it to use the suggested syntax, but with the prop's value instead of a hardcoded 5. Looks good to you now?

Issue #1532, opened 2018-11-07
Experimental Component - Notification


emyarod:
**Development**
A developer can discover and install the experimental React component by following accessible documentation.

* [ ]  Add to carbon-components-react
* [ ]  Update website (add to the experimental section and update component status page)

PR #1537, opened 2018-11-07
feat(Notification): add experimental notification

emyarod:
TODO:

* [x] fix failing test due to a11y warning icon workaround

---

Closes IBM/carbon-components-react#1532

This PR will add support for the experimental `<InlineNotification />` and `<ToastNotification />` components

#### Changelog

**New**

* add support for experimental `<Notification />` component

**Changed**

* fixed notification icons not appearing for `<ToastNotification />`
* conditionally render icons based on notification type (based on new notification guidelines)

---

An open question: is there a reason for separating the inline and toast notification components rather than consolidating them in a common `<Notification />` component? It seems aside from `timeout`, `caption`, and minor markup changes, the majority of the code is shared between the two types of notifications. Would it be worth deduplicating this code?

### DIFF
--- a/src/components/Notification/Notification.js
+++ b/src/components/Notification/Notification.js
@@ -9,6 +9,8 @@ import {
   iconWarningSolid,
 } from 'carbon-icons';
 import Icon from '../Icon';
+// temporary workaround for a11y warning icon. TODO: for @carbon/icons-react
+import a11yIconWarningSolid from './a11yIconWarningSolid';
 
 export class NotificationButton extends Component {
   static propTypes = {
@@ -185,44 +187,12 @@ export class ToastNotification extends Component {
       { [`bx--toast-notification--${this.props.kind}`]: this.props.kind },
       className
     );
-    const a11yWarningIcon = (
-      <svg
-        class={`bx--${notificationType}-notification__icon`}
-        width="20"
-        height="20"
-        xmlns="http://www.w3.org/2000/svg">
-        <defs>
-          <path
-            d="M18.05 15.95L9.925.95a.625.625 0 0 0-1.1 0L.7 15.95a.625.625 0 0 0 0 .625.625.625 0 0 0 .55.3H17.5a.625.625 0 0 0 .55-.925z"
-            id="a"
-          />
-          <path
-            d="M.55.006h1.406v5H.55v-5zm.7 8.119a.938.938 0 1 1 0-1.875.938.938 0 0 1 0 1.875z"
-            id="b"
-          />
-        </defs>
-        <g fill="none" fill-rule="evenodd">
-          <path
-            d="M18.675 16.575l-8.125-15a.625.625 0 0 0-1.1 0l-8.125 15a.625.625 0 0 0 0 .625.625.625 0 0 0 .55.3h16.25a.625.625 0 0 0 .55-.925z"
-            fill="#FDD13A"
-          />
-          <g>
-            <path
-              d="M9.3 6.881h1.406v5H9.3v-5zM10 15a.938.938 0 1 1 0-1.875A.938.938 0 0 1 10 15z"
-              fill="#171717"
-              fill-rule="nonzero"
-            />
-          </g>
-          <path d="M0 0h20v20H0z" />
-        </g>
-      </svg>
-    );
     const NotificationIcon = kind => {
       switch (kind) {
         case 'info':
           return null;
         case 'warning':
-          return a11yWarningIcon;
+          return a11yIconWarningSolid(notificationType);
         default:
           return (
             <Icon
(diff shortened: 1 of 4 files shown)


src/components/Notification/Notification.js
```diff
@@ -42,20 +42,17 @@ export class NotificationButton extends Component {
...
       {
         'bx--toast-notification__close-button': notificationType === 'toast',
         'bx--inline-notification__close-button': notificationType === 'inline',
       },
       className
     );
-
     const iconClasses = classNames({
-      'bx--toast-notification__icon': notificationType === 'toast',
+      'bx--toast-notification-icon': notificationType === 'toast',
```
asudoh: Looks that this change drifts away from https://github.com/IBM/carbon-components/blob/v9.46.0/src/components/notification/_toast-notification.scss#L139?

src/components/Notification/Notification.js
```diff
@@ -42,20 +42,17 @@ export class NotificationButton extends Component {
       notificationType,
       ...other
     } = this.props;
-
```
asudoh: Nitpick, but wondering if a tool against blank lines is in use - We tend to keep blank lines wherever it's meaningful.

src/components/Notification/Notification.js
```diff
@@ -158,16 +151,14 @@ export class ToastNotification extends Component {
   useIcon = kindProp =>
     ({
       error: iconErrorSolid,
-      info: iconInfoSolid,
       success: iconCheckmarkSolid,
-      warning: iconWarningSolid,
+      warning: iconErrorSolid,
```
asudoh: Just wanted to double-check if the icon list change here is intentional.

src/components/Notification/Notification.js
```diff
@@ -181,15 +172,64 @@ export class ToastNotification extends Component {
...
       ...other
     } = this.props;
-
     const classes = classNames(
       'bx--toast-notification',
       { [`bx--toast-notification--${this.props.kind}`]: this.props.kind },
       className
     );
+    const a11yWarningIcon = (
+      <svg
```
asudoh: Probably we'd make it a separate React component in a separate module?

src/components/StructuredList/StructuredList-story.js
```diff
@@ -12,6 +11,7 @@ import {
   StructuredListCell,
 } from '../StructuredList';
 import StructuredListSkeleton from '../StructuredList/StructuredList.Skeleton';
+import { componentsX } from '../../internal/FeatureFlags';
```
asudoh: Did you mean to put changes to this file in another PR? Same for `Tag.js`.

src/components/Notification/Notification.js
emyarod: I can add them back, I was just trying to fit more lines on my screen for easier reading. This branch is still WIP so it just contained work I wanted to temporarily save and check into git

src/components/Notification/Notification.js
emyarod: Thanks for catching that, my mistake! This branch is still WIP so it just contained work I wanted to temporarily save and check into git

src/components/StructuredList/StructuredList-story.js
emyarod: since I created this branch off of those feature branches (tag and structured list) they contain the commits from those branches as well. I imagined the branches would be merged "in order" so I didn't split off from `master` branch each time. If it would be better for me to recreate these branches from `master` I can do that

src/components/Notification/Notification.js
emyarod: I had some questions about that:

if I'm not mistaken, it looks like the component is using `bx--toast-notification-icon` here https://github.com/IBM/carbon-components/blob/v9.46.0/src/components/notification/notification.hbs#L88

do we want to sync the classes and styles on this component for the two libraries?

vanilla notification
image

React `bx--toast-notification-icon`
image

React `bx--toast-notification__icon`
image

based off of the `bx--inline-notification__close-icon` class, is it supposed to be `bx--toast-notification__close-icon` in both repos?

emyarod: going to go with `bx--toast-notification__close-icon` to follow @tw15egan's latest changes here https://github.com/IBM/carbon-components/pull/1412

src/components/Notification/Notification.js
emyarod: sure, do we want to keep it within the `src/components/Notification` directory then?

emyarod:
@asudoh just want to draw some discussion about the open question in the PR body:

> is there a reason for separating the inline and toast notification components rather than consolidating them in a common <Notification /> component? It seems aside from timeout, caption, and minor markup changes, the majority of the code is shared between the two types of notifications. Would it be worth deduplicating this code?

emyarod:
component now ready for design review [here](http://react-experimental.carbondesignsystem.com/?selectedKind=Notifications)

alisonjoseph:
I went ahead and merged this @emyarod. I believe there are some design tweaks needed for this but will be related to both vanilla/react. I think as long as our html matches then we are good to go.

Issue #1527, opened 2018-11-06
Missing onClick in the Copy component


Yohanna:
## Detailed description

The [Copy](https://github.com/IBM/carbon-components-react/blob/master/src/components/CodeSnippet/CodeSnippet.js#L174-L179) component is missing `onClick={onClick}` causing any function passed to the `onClick` prop to Inline `CodeSnippet` to not get called. 

> What version of the Carbon Design System are you using?

6.43.0

## Steps to reproduce the issue

https://codesandbox.io/s/2x685pop80

asudoh:
Hi 👋 thank you for writing this up! This is caused by missing `{...other}` that are seen in other variations.

PR #1538, opened 2018-11-07
fix(CodeSnippet): inline type was missing onClick handler

slaytanic:
Closes IBM/carbon-components-react#1527

On the type="inline" variant of the CodeSnippet the Copy button was missing the onClick handler.

#### Changelog

**Changed**

* CodeSnippet: Added onClick handler for Copy button when type="inline"

### DIFF
--- a/src/components/CodeSnippet/CodeSnippet.js
+++ b/src/components/CodeSnippet/CodeSnippet.js
@@ -172,6 +172,7 @@ export default class CodeSnippet extends Component {
     if (type === 'inline') {
       return (
         <Copy
+          onClick={onClick}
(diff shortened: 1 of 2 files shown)


src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -172,6 +172,7 @@ export default class CodeSnippet extends Component {
     if (type === 'inline') {
       return (
         <Copy
+          onClick={onClick}
```
asudoh:
```suggestion
          {...other}
          onClick={onClick}
```

PR #1540, opened 2018-11-08
Support configurable CSS class prefix

### LINKED ISSUE #1359: Support for configurable selector prefix
### Summary/Justification

The vanilla library has supported configurable selector prefix. It recently helped `carbondesignsystem.com` to use two versions of Carbon styles at the same time. It also helps branding, meaning Carbon to be used in different brands.

N/A

### "Must have" functionality

1. Support for alternate selector prefix system-wide
2. Support for alternate selector prefix by component-instance basis (e.g. via React props)

TBD

asudoh:
Fixes #1359.

#### Changelog

**Changed**

* All CSS class references now interpolates with [CSS class prefix settings in `carbon-components`](https://github.com/IBM/carbon-components/blob/v9.43.0/src/globals/js/settings.js#L16).

### DIFF
--- a/src/components/Search/Search-story.js
+++ b/src/components/Search/Search-story.js
@@ -15,7 +15,7 @@ const props = () => ({
   small: boolean('Small UI (small)', false),
   light: boolean('Light variant (light)', false),
   name: text('Form item name (name)', ''),
-  value: text('Value (value)', ''),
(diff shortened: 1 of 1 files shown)


asudoh:
Travis error was caused by the old Flatpickr dependency from vanilla. Contemplating possible approaches to fix that.

PR #1553, opened 2018-11-12
docs: Add next.js example

vpicone:
Styled-JSX actually has some really cool sass integration that I was going to show, but I think it might be too involved for a basic 'MVP' example. Maybe a blog post would be better.


vpicone:
One thing I wasn’t sure of here is how to handle dependencies. This is the first project I’ve worked with that uses yarn-offline, but it felt weird to have an example project messing with that directory. Would we want the examples to do that?

joshblack:
Hey @vpicone! 👋 Thanks for doing this! Went ahead and added a `.yarnrc` file for the offline mirror problem you were running into 👍

.gitignore
```diff
@@ -25,3 +25,6 @@ package-lock.json
 
 # Generated by storybook@4
 .cache
+
+# Example build artifacts
+.next
```
joshblack: Could we move this to a local `.gitignore` by the example? Like: `examples/next/.gitignore`?

joshblack:
Would love to see a blog post on this btw @vpicone! Let us know if you'd like to publish through our publication on Medium or want to coordinate when you release one 😄

.gitignore
vpicone: Totally!

examples/next/README.md
```diff
@@ -0,0 +1,12 @@
...
+> See how to integrate `carbon-components` and `carbon-components-react` into a project
+> using Next.js.
+
+## Usage
+
+To run this example, all you need to do is `cd` into the directory, run `yarn install`, and then `yarn start`.
+
+## Learn Next.js
+
+Next has some great tutorials at [nextjs.org](https://www.nextjs.org/learn).
```
asudoh: I got a dead link here - Just double-checking if the domain is right.

examples/next/pages/index.js
```diff
@@ -0,0 +1,52 @@
...
+        <div className="App">
+          <header className="App-header">
+            <h1 className="App-title">
+              Welcome to React, with Carbon and Next!
+            </h1>
+          </header>
+          <p className="App-intro">
+            To get started, edit <code>pages/index.js</code> and save to reload.
+            Then, check out the{' '}
+            <a href="https://www.nexjs.org/learn">Next.js</a> tutorial.
```
asudoh: Uh - I guess I meant here instead of what I pointed out earlier. You meant `nextjs.org` with (`t`)?

PR #1319, opened 2018-09-14
fix(#726): Icon component should have ability to remove title tab inside svg

xiaolixu:
Closes IBM/carbon-components-react#726

Icon component should have ability to remove title tab inside svg

#### Changelog

**Changed**
**Icon**

* add _iconTitle_ prop , and the Icon's title will still use _description_ prop if  _iconTitle_ is _undefined_ 

**Tooltiop**

* add _iconTitle_ prop , and the default value is _""_

PR #1554, opened 2018-11-12
fix(#726): Icon component should have ability to remove title tab inside svg

### LINKED ISSUE #726: Icon component should have ability to remove title tab inside svg
## Detailed description

In some case Icon component should not contain title tag or attribute
Eg: 
image

 feature request
Icon -> Tooltip
Icon should add a new switch to enable/disable title tag
Chrome
Latest
Cloud Object storage
Please resolve asap

xiaolixu:
Continue the work on [#1319](https://github.com/IBM/carbon-components-react/pull/1319)
Closes #726

Icon component should have ability to remove title tab inside svg

Changelog
Changed
Icon

add iconTitle prop , and the Icon's title will still use description prop if iconTitle is undefined
Tooltiop

add iconTitle prop , and the default value is ""


joshblack:
Something I'm unsure about is having both `aria-label` and a `title`. They should be fulfilling the same purpose here, I believe. Would it be helpful to have the `title` prop generate a `<title>` child and setup `aria-labelledby` to point to the `<title>` node?

joshblack:
@xiaolixu I believe certain browsers will announce the `<title>` contents for the SVGRoot, at least in Chrome it will, so we have to be careful about that. Definitely understand that `<title>` in an SVG container is shown as a browser tooltip in some browsers 👍

Issue #1542, opened 2018-11-08
[OverflowMenu Storybook]: Not rendering properly


tay1orjones:
## Detailed description

`OverflowMenu` is currently not rendering in the storybook. The associated info in the story is also incomplete with single letter component names, and some visual regression overflow with the props table.

![image](https://user-images.githubusercontent.com/3360588/48221431-b00b7180-e357-11e8-9ffd-6cf847057fc3.png)

![image](https://user-images.githubusercontent.com/3360588/48221452-c7e2f580-e357-11e8-8a26-40bf31633fd1.png)

vpicone:
This is happening in a lot of components. It looks like the displayName isn't getting set appropriately in static builds. Explicitly setting the display name in the component fixes the problem. `OverflowMenu.displayName = 'OverflowMenu';`

Something must be funky with the wrapComponent business.

![screen shot 2018-11-09 at 2 03 26 pm](https://user-images.githubusercontent.com/4078018/48285470-51fa8f00-e428-11e8-86dc-f1f2871d47fc.png)

asudoh:
Hi @vpicone 👋 thanks a lot for doing thorough analysis and providing us with a set of proposals! I agree with you that 1. is the least brittle (thus the best) solution. Let's see how others respond and unless anybody disagrees I'm looking forward to seeing a PR!

alisonjoseph:
Yes thank you so much for taking the time to research this. 🙌 I agree on 1. defining the display name for each component sounds like the best solution.

vpicone:
@joshblack that one worked really well, it essentially does what has always happened in that it grabs the function name. My PR actually broke some snapshot tests because a few of the components were given the "ForwardRef" display name. Explicitly setting the display name fixes that.

From the snapshot test:
{code}
    @@ -1,6 +1,6 @@
    - <ForwardRef // Our snapshot has ForwardRef as the component's displayName
    + <SideNavMenuItem // Explicitly setting it overrides that.
        className="custom-classname"
        href="#"
        isActive={true}
      >
        <li
    @@ -26,6 +26,6 @@
                </span>
              </SideNavLinkText>
            </a>
          </Link>
        </li>
    - </ForwardRef>
    + </SideNavMenuItem>
{code}

PR #1557, opened 2018-11-12
Add explicit display names to components

vpicone:
Closes IBM/carbon-components-react#1542

Storybook v4 automatically uglifies js so function names can no longer be reliably used to infer display names. The maintainers recommend adding display names explicitly.

#### Changelog

**New**

* Add explicit displayNames when they haven't already been set
* Add test to ensure that displayNames are set on new components.


vpicone:
The snapshot tests fail since when they were generated, the function name was set as the 'displayName' so some of the components like `SideNavMenuItem` which are assigned to `React.forwardRef(...)` got a computed displayName of `ForwardRef`. I beliiiieve Any sort of automatic inference based on function name is gonna have this same issue.

I added a __tests__ in the component directory. Since this is more like a component requirement check and not so much a behavior check, if you guys have a better spot for the test I'm down to move it.

vpicone:
@asudoh, yes this explicitly adds the displayName whenever it wasn't already set (usually in the dataTable subcomponents with the wrapComponent() method. 

I'll update the snapshots, I think the more descriptive names are an improvement. To me, the snapshots breaking in this way introduced me to the shortcoming of computed displayNames. For this sort of thing, using named (instead of anonymous functions) kind of fudges around manually setting the display name, but the snapshots would still need to be updated.

```javscript
// appears as <ForwardRef> (bad)
export default React.forwardRef((props, ref) => {
  return <HeaderMenu {...props} focusRef={ref} />;
});
```

vs

```javscript
// appears as <ForwardRef(headerMenu)> (still not great)
export default React.forwardRef(
function headerMenu(props, ref) => {
  return <HeaderMenu {...props} focusRef={ref} />;
});
```

joshblack:
I'm unsure about making this change for the sake of supporting the "story source" in the addon panel. We already provide the story source directly in the default tab, and making changes only for what is already a controversial source of documentation feels off. Could we instead:

- Use the babel plugin?
  - Drawback here is that it does not work for all cases
- Update the uglify settings on webpack to not mangle names?

vpicone:
I think both of those options are good. We could swap out the UglifyJS plugin for the one CRA uses: https://github.com/webpack-contrib/terser-webpack-plugin

After playing with the forwardRef stuff in the UI shell components, it can get a bit tricky with HOCs/exporting wrapped components. So a simpler approach could definitely be better here.

vpicone:
@joshblack right on, I'm gonna open a new PR. I think switching to full control was prob necessary with how extensive the setup is. I think the only change I have is the useExternalCss boolean at the end needs to be true maybe.

PR #1559, opened 2018-11-13
docs: Update commit message instructions and links

vpicone:
Fixes the internal and external links. Add the most significant requirements to our own docs.


joshblack:
@vpicone I think if you change `*` to `-` in the markdown list you should be good to go!

PR #1560, opened 2018-11-13
chore: Update webpack and swap out minifier

vpicone:
Closes IBM/carbon-components-react#1542

This swaps the out uglifyjs minifier for the one used by create react app. In addition, we stop the mangling of function names.

**New**

* terser-webpack-plugin

**Changed**

* switched to storybooks 'full-control' mode
* swaps the out uglifyjs minifier for the one used by create react app
* turns off mangling 

**Issues**

* Some components will still have unreadable display names (for instance those that are exported wrapped with `React.forwardRef`). They show up as `<Unknown>` in the info addon.
* A tool could be made to automatically add the appropriate displayName for these cases.


vpicone:
Here's an example of the relatively few displayNames generated from components wrapped w/ React.forwardRef. There's a couple different solutions that could come from a helper tool in a separate issue.

Unknown

PR #1566, opened 2018-11-14
fix(DatePicker): update both calendars when value is changed

lekterable:
Closes IBM/carbon-components-react#1469

When manually changing the value prop of range DatePicker, second DatePickerInput is not getting updated. Seems like this is a known bug in the flatpickr library (https://github.com/flatpickr/flatpickr/issues/1345), but during investigation I've discovered that in our case this can be fixed by forcing the calendar to update using optional second argument in setDate. I've also decided to use lodash isEqual function in the condition because the previous condition was always true when values were arrays. The last thing is that I've used componentDidUpdate instead of deprecated componentWillUpdate.

#### Changelog

**New**

* lodash.isequal import

* optional argument in setDate

**Changed**

* componentDidUpdate instead of UNSAFE_componentWillUpdate

* adjusted condition


lekterable:
@jnm2377 
Did you use sandbox to reproduce it?
Because just like the author of the issue mentioned:
> Using Sandbox I cannot really reproduce, it's not even managed at all and no input is reset.
For some reason this cannot be reproduced using sandbox.

src/components/DatePicker/DatePicker.js
```diff
@@ -212,14 +213,14 @@ export default class DatePicker extends Component {
     locale: 'en',
   };
 
-  UNSAFE_componentWillUpdate(nextProps) {
-    if (nextProps.value !== this.props.value) {
+  componentDidUpdate(prevProps) {
+    if (!isEqual(prevProps.value, this.props.value)) {
       if (this.cal) {
-        this.cal.setDate(nextProps.value);
+        this.cal.setDate(this.props.value, true);
```
asudoh: Wondering why the 2nd argument (`triggerChange`) is added here - The addition caused `onChange` event firing even for non-user-initiated change.

lekterable: @asudoh 
> Seems like this is a known bug in the flatpickr library (flatpickr/flatpickr#1345), but during investigation I've discovered that in our case this can be fixed by forcing the calendar to update using optional second argument in setDate. 

I couldn't fix it without adding it, I think that its a bug in rangeplugin in flatpickr and the second calendar doesnt update when using a setDate without the second argument.

asudoh:
Given this change causes redundant `onChange` event, we need more investigation to see if there is a better solution.

joshblack:
Hi @lekterable! Going to close this for now due to inactivity, feel free to comment and we would love to re-open and have you contribute this! 😄

Issue #1564, opened 2018-11-14
FileUploaderButton div container doesn't listen to click


kylembrake:
## Detailed description

> Describe in detail the issue you're having.
Only the label (red background) listens for a mouse click.
<img width="164" alt="screen shot 2018-11-14 at 12 13 42 pm" src="https://user-images.githubusercontent.com/39499740/48499773-2fd98600-e807-11e8-97ea-53e22a99b76d.png">

> Is this issue related to a specific component?
FileUploaderButton

> What did you expect to happen? What happened instead? What would you like to see changed?
I would expect the entire button to be clickable

> What browser are you working in?
Chrome Version 70.0.3538.102 (Official Build) (64-bit)

> What version of the Carbon Design System are you using?
"carbon-components-react": "^6.46.7"

PR #1576, opened 2018-11-16
fix(FileUploaderButton): handle container div onClick

emyarod:
Closes IBM/carbon-components-react#1564

Currently the only way to activate the `<input type="file">` on the `<FileUploaderButton />` is by clicking the button label. This PR will allow the user to click any part of the `<FileUploaderButton />` to bring up the file upload prompt.

#### Changelog

**New**

* add `onClick` handler for container `<div>`
* add `onClick` handler for file input `<label>`

**Changed**

* stop propagation of event if file input label is clicked

### DIFF
--- a/package.json
+++ b/package.json
@@ -352,5 +352,5 @@
       "enzyme-to-json/serializer"
     ]
   },
-  "bundleSizeThreshold": 80000
+  "bundleSizeThreshold": 81000
 }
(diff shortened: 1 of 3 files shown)


emyarod:
hi @asudoh my intention was to preserve the markup for compatibility purposes but I wonder if it would be better to wrap the `input` within the `label` or use a React fragment to contain the two elements instead?

emyarod:
@alisonjoseph I am planning to modify the markup later today to be better suited for a11y, so can we hold off on the merge until then? I will mark this as WIP for the time being

asudoh:
Thank you for the update @emyarod!

Issue #1500, opened 2018-10-30
Tooltip for disabled InlineCheckbox component (Missing title attribute)


cubicrace:
### Summary

Support tooltip for inline checkbox
Currently the inline checkbox doesn't accept the title attribute.

### Justification
Say we have a datatable with multselect enabled.
The datatable has users with different state like Active, Suspended, Deleted.
Now inorder for a user to perform action on multiple rows, only rows with similar status should be selected. 
So what we want is to prevent a user from selecting multiple rows in a datatable having different states in order to perform a bulk action.
If the user has selected a row with status Active then all rows with status other than Active will be disabled. Inorder to inform this to the user, we need to show a tooltip whenever the user hovers or tries to click a disabled checkbox in a datatable.

### Specific timeline issues / requests
By end of Nov 2018 for an upcoming release.

PR #1578, opened 2018-11-19
feat(InlineCheckbox): allow optional tooltip

emyarod:
Closes IBM/carbon-components-react#1500

This PR allows the user to provide an optional `title` to serve as a tooltip for the `<InlineCheckbox>` component

#### Changelog

**New**

* `title` prop for `<InlineCheckbox>`


src/components/InlineCheckbox/InlineCheckbox.js
```diff
@@ -68,6 +72,7 @@ export default class InlineCheckbox extends React.Component {
       name,
       onClick,
       onKeyDown,
+      title = null,
```
asudoh: Wondering if we can keep the default `undefined` - Though I think `null` should work, too, just thought we'd avoid snapshot change :thinking:.

PR #1580, opened 2018-11-20
fix(Modal): Allow components as modal heading

tombrowndev:
Add support to supply a component as the Model Label.

Particularly useful for passing in SVG Components.


src/components/Modal/Modal.js
```diff
@@ -19,7 +19,7 @@ export default class Modal extends Component {
     onRequestClose: PropTypes.func,
     id: PropTypes.string,
     modalHeading: PropTypes.string,
-    modalLabel: PropTypes.string,
+    modalLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
asudoh: Did you mean `PropTypes.node` instead...?

tombrowndev: Thats an even better solution, thanks!

marijohannessen:
Thanks for fixing this @tombrowndev ! There's a tiny merge conflict that needs to be resolved before we can merge it in 👍

joshblack:
Tried to merge in with what I thought would be the resolution, feel free to revert if that's not the case!

PR #1584, opened 2018-11-27
Pass function reference instead of direct call

pketki:
As discussed on slack, we should pass a reference rather than making a call otherwise the function gets called at render time even without calling the batch action.


asudoh:
Thank you for jumping in @pketki! https://github.com/IBM/carbon-components-react/blob/v6.39.0/src/components/DataTable/stories/shared.js#L60-L61 tells us that `batchActionClick` should be a "thunk" - The function body won't be executed in render time this way. But don't hesitate to let us know if you think otherwise!

joshblack:
Hi @pketki! 👋 Thanks again for taking the time to contribute this content fix! Am going to go ahead and close this for now, but feel free to comment and make updates to content from @asudoh's feedback and we'll definitely get it merged in!

Issue #1592, opened 2018-11-29
<MultiSelect> values clear even when component is disabled


Stephen-Bohner:
The <MultiSelect> component will clear its selected values even when the component is disabled

We are using the disabled MultiSelect component as a view only field. This particular bug allows the user to click the x in the <MultiSelect> component (as shown in the screenshot below) which clears the component's values. This is a problem because the user cannot re-select any values since the field is disabled.

What did you expect to happen?
**I would expect the user to not have the ability to clear the disabled field.**

What browser are you working in?
**Chrome**

What version of the Carbon Design System are you using?
**"carbon-components-react": "^6.18.3",**

![screen shot 2018-11-29 at 10 25 50 am](https://user-images.githubusercontent.com/43971523/49232167-6d224400-f3c1-11e8-8cca-ae6e8ffdd7a8.png)
![screen shot 2018-11-29 at 10 26 09 am](https://user-images.githubusercontent.com/43971523/49232168-6d224400-f3c1-11e8-8de5-0568a846109a.png)

PR #1595, opened 2018-12-02
fix(MultiSelect): disabled multiSelect being able to clear its select…

jibbey:
Closes IBM/carbon-components-react#1592

Fixed the ability to clear selected items on a disabled Multiselect


emyarod:
hey @jibbey thanks for jumping in! I had a fix ready yesterday after self-assigning the ticket but didn't cut the PR yet because I found some other issues with the disabled `<MultiSelect>` state (#1596). There is some overlap between this PR and #1597, but I'll handle any conflicts after your PR gets merged in

src/components/MultiSelect/MultiSelect.js
```diff
@@ -203,7 +203,7 @@ export default class MultiSelect extends React.Component {
                 <ListBox.Field {...getButtonProps({ disabled })}>
                   {selectedItem.length > 0 && (
                     <ListBox.Selection
-                      clearSelection={clearSelection}
+                      clearSelection={!disabled ? clearSelection : () => null}
```
emyarod: do we want to avoid creating a new instance of the anonymous function `() => null` each time a render is triggered?

jibbey: I'm not sure of a better solution, though I am sort of new to this. 👍

asudoh:
```suggestion
                      clearSelection={!disabled ? clearSelection : noop}
```

Where `const noop = () => undefined` at the top level - Meaning, seconding @emyarod.

jibbey:
@asudoh Looks like changing it to noop did not pass integration test.

asudoh:
@jibbey You need to define `noop` yourself - `const noop = () => undefined` at the top level.

jibbey:
> @jibbey Wrt latest CircleCI error - Do you want to re-run `yarn prettier` and see if Prettier re-formats one or more of files that were changed by you? Thanks!

Yes, I forgot the semicolon, I ran prettier and that added it. Hopefully my most recent commit passes!

PR #1597, opened 2018-12-02
fix(MultiSelect): prevent modification of disabled multiselect actions

### LINKED ISSUE #1592: <MultiSelect> values clear even when component is disabled

### LINKED ISSUE #1596: <MultiSelect> checkboxes modifiable even when component is disabled
## Detailed description

When `<MultiSelect>` is `disabled`, the checkboxes are still selectable/deselectable.

The easiest way to reproduce this behavior is in the [`<MultiSelect>` component storybook example](http://react.carbondesignsystem.com/?selectedKind=MultiSelect&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel)

1. Expand the selection dropdown menu
2. Disable the component via storybook knobs
3. Toggle checkbox states

Related: #1592

emyarod:
Closes #1592
Closes #1596 

This PR will prevent users from modifying `<MultiSelect>` actions while the component is disabled.

#### Changelog

**New**

- disabled styles for `<ListBox.Selection>`, dependent on https://github.com/IBM/carbon-components/pull/1508

**Changed**

- break out of `<Selection>` event handlers when component is disabled

### DIFF
--- a/src/components/ListBox/ListBoxSelection.js
+++ b/src/components/ListBox/ListBoxSelection.js
@@ -16,12 +16,16 @@ const ListBoxSelection = ({
   clearSelection,
   selectionCount,
   translateWithId: t,
+  disabled,
 }) => {
   const className = cx({
     [`${prefix}--list-box__selection`]: true,
     [`${prefix}--list-box__selection--multi`]: selectionCount,
   });
   const handleOnClick = event => {
+    if (disabled) {
+      return;
+    }
     // If we have a mult-select badge, clicking it shouldn't open the menu back
     // up. However, if we have a clear badge then we want the click to have this
     // behavior.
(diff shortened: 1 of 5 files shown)


emyarod:
@asudoh yeah it is probably worth testing the disabled state. I'll add some test cases

emyarod:
@asudoh I've added a test for disabled state behavior

asudoh:
Thank you for updating tests @emyarod!

Issue #1602, opened 2018-12-03
Number Input component bug


vidhi11:
## Detailed description

> Describe in detail the issue you're having.
- Facing UI issue with NumberInput component. We upgraded the package to `  "version": "9.57.0",` to meet PUP. And now we see: up & down arrows are not properly aligned with the field box. 

> Is this a feature request (new component, new icon), a bug, or a general issue?
- General issue/Bug.

> Is this issue related to a specific component?
- Yes, NumberInput component. 

> What did you expect to happen? What happened instead? What would you like to see changed?
<img width="860" alt="image" src="https://user-images.githubusercontent.com/32270454/49399351-ab439e80-f706-11e8-9cfe-f174565ebe30.png">
- We like to see arrows properly aligned with field. 

> What browser are you working in?
- Mozilla and Chrome.

> What version of the Carbon Design System are you using?
- 9.57.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
- We are releasing our F5 micro-service by 12/5/2018.

## Steps to reproduce the issue
- No steps. 

## Additional information

- Screenshots: 
1) On Mozilla: 
<img width="860" alt="image" src="https://user-images.githubusercontent.com/32270454/49399507-11c8bc80-f707-11e8-86eb-80a2e43eb1c3.png">

2) On Chrome: 
<img width="1348" alt="image" src="https://user-images.githubusercontent.com/32270454/49399531-1f7e4200-f707-11e8-874f-04e4a1a715ad.png">

lekterable:
Hey @vidhi11 you can quickly fix this by using hideLabel prop in NumberInput component, I still think this is a bug though and I will work on a PR for it.

vidhi11:
@lekterable yes, hideLabel prop did fix this. Thanks for quick fix.

PR #1604, opened 2018-12-03
fix(NumberInput): add no-label class if no labels provided

lekterable:
Closes IBM/carbon-components-react#1602

It seems like a good idea to automatically add a no-label class when there are no labels passed as a props.

#### Changelog

**Changed**

- no-label class is added automatically when there are no labels specified.


lekterable:
sorry for the force push here ^ but I've mixed branches

asudoh:
Hi @lekterable thank you very much for jumping in and my apologies for being late on review. It appears that `bx--number--helpertext` not being present when `helperText` is falsy seems to be causing the issue. Wondering if you can try adding that class always?

lekterable:
@asudoh if it's better this way, sure!

lekterable:
Hey, @asudoh and @joshblack I believe this one is ready for review since my last comment. Maybe you've missed my commit?

PR #1606, opened 2018-12-04
chore(components): remove deprecated components

asudoh:
(Please use "Diff settings" - "Hide whitespace changes" to review the code)

Refs IBM/carbon-components#1501.

#### Changelog

**New**

- `breakingChangesX` feature flag. That can be enabled in Storybook with `CARBON_USE_BREAKING_CHANGES=true` environment variable.

**Changed**

- Exports of deprecated features will be `null` if above feature flag is `true`.
- Stories of deprecated features are no longer added if above feature flag is `true`.
- Tests of of deprecated features no longer run if above feature flag is `true`.


emyarod (approved):

README.md
```diff
@@ -52,6 +52,12 @@ We recommend the use of [React Storybook](https://github.com/storybooks/react-st
      $ export CARBON_USE_EXPERIMENTAL_FEATURES=true
      ```
 
+   - `true` to `CARBON_USE_BREAKING_CHANGES` environment variable to test some of the breaking changes for the next release:
```
joshblack: Should we add these notes to a `docs/experimental.md` - `## Feature flags` section instead of placing in our `README.md`?

src/components/Dropdown/Dropdown.js
```diff
@@ -260,3 +261,5 @@ export default class Dropdown extends PureComponent {
     return dropdown;
   }
 }
+
+export default (!breakingChangesX ? Dropdown : null);
```
joshblack: When our feature flags are replaced with their default value, how would this code be compiled when it's set to `false`? Would it properly be stripped through uglify/terser? Just curious 👀

src/components/Pagination/Pagination-test.js
```diff
@@ -4,11 +4,12 @@ import Icon from '../Icon';
 import Pagination from '../Pagination';
 import Select from '../Select';
 import SelectItem from '../SelectItem';
+import describeBreakingChangesXFeatures from '../../tools/describeBreakingChangesXFeatures';
 import { shallow, mount } from 'enzyme';
 
 jest.useFakeTimers();
 
-describe('Pagination', () => {
+describeBreakingChangesXFeatures('Pagination', () => {
```
joshblack: Should we put this in one of our jest setup files?

src/tools/describeBreakingChangesXFeatures.js
```diff
@@ -0,0 +1,11 @@
+import { breakingChangesX } from '../internal/FeatureFlags';
+
+export default (name, callback) => {
+  if (!breakingChangesX) {
+    describe(name, callback);
+  } else {
+    describe(name, () => {
```
joshblack: Should we just noop here?

src/components/Dropdown/Dropdown.js
asudoh: Good question, it'll be `null`-export only as long as the code is tree-shakable. Added some code to demonstrate that in our Rollup build.

src/tools/describeBreakingChangesXFeatures.js
asudoh: Tried that, but got `Your test suite must contain at least one test.` error when `breakingChangeX` is `true`.

src/components/Pagination/Pagination-test.js
asudoh: Make sense - Done.

asudoh:
@joshblack Did you have any more comments...? Thanks!

joshblack:
_PR Reviews 1/11_

Notes

- We should try and review this during our team meeting so we can get the full idea of what the changes are and the motivation behind them 👍 
- Larger discussion here should be had around how we're doing removals/breaking changes

src/tools/describeBreakingChangesXFeatures.js
joshblack: @asudoh could we do `describe.skip` or something similar?

asudoh: @joshblack Something I considered, tried that today, but getting something like:

```
> yarn test dropdownitem
yarn run v1.11.1
$ jest dropdownitem
 FAIL  src/components/DropdownItem/DropdownItem-test.js
  ● Test suite failed to run

    Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: null.

    Check the render method of `WrapperComponent`.

       5 | describe.skip('DropdownItem', () => {
       6 |   describe('Renders as expected', () => {
    >  7 |     const wrapper = mount(
         |                     ^
       8 |       <DropdownItem className="extra-class" itemText="test" value="test" />
       9 |     );
      10 |
```

Seems that `describe.skip()` runs its callback, even though [the document seems to say otherwise](https://jestjs.io/docs/en/api#describeskipname-fn).

README.md
asudoh: Makes sense for `CARBON_USE_EXPERIMENTAL_FEATURES`. Removed that reference here and created: #1794

asudoh:
Removed the code for tree-shakability of deleted components. @joshblack Let me know if you have any further comments - Thanks!

asudoh:
Merging as all two reviewers' comments have been addressed or answered. Anybody please don't hesitate to speak up if you have any concerns, can revert if it's the case. Thanks!

Issue #1616, opened 2018-12-06
Default Modal component loses focus when created dynamically with prop open="true" 


mkunat:
[Title]: Brief description
Default Modal component loses focus on primary actionable button when created dynamically with prop open="true" 

## Detailed description
Modal component loses focus on primary actionable button when created dynamically with prop open="true". The issue may be recreated on Storybook deployment. When you open the default Modal example the focus on the primary button is missing. 
http://react.carbondesignsystem.com/?selectedKind=Modal&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel
If you try going to KNOBS tab and toggle `open` prop you’ll see primary button focused.
Focusing doesn't happen when Modal is mounted with `open` prop already with true value. Our modal handling, actually creates modal dynamically.

 Is this issue related to a specific component?
Modal

 What browser are you working in?
IE 11, Firefox, Chrome, Edge

 What version of the Carbon Design System are you using?
6.55.5

Please create a reduced test case in CodeSandbox
import React from 'react';
import { render } from 'react-dom';
import { Modal } from 'carbon-components-react'

const App = () => (
    Modal
      open
      modalHeading="Modal Heading"
      modalAriaLabel=""
      primaryButtonText="Primary"
      secondaryButtonText="Cancel"
      iconDescription="Close"
      
      Hello world
      /Modal
);

render(<App />, document.getElementById('root'));

PR #1621, opened 2018-12-07
feat(Modal): focus primary focus element after component mount

emyarod:
Closes IBM/carbon-components-react#1616

`<Modal />` currently does not focus the primary focus element after initial render. This PR adds a check to the `componentDidMount` life cycle method to call focus on the primary focus element.

#### Changelog

**New**

- `componentDidMount` method for calling focus on primary focus element after render

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -221,12 +221,12 @@ export default class Modal extends Component {
     }
   };
 
-  componentDidMount = () => {
+  componentDidMount() {
     if (!this.props.open) {
       return;
     }
     this.focusButton(this.innerModal.current);
-  };
+  }
 
   handleTransitionEnd = evt => {
     if (
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -208,8 +208,8 @@ export default class Modal extends Component {
     }
   };
 
-  focusButton = evt => {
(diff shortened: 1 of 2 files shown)


src/components/Modal/Modal.js
```diff
@@ -202,6 +202,22 @@ export default class Modal extends Component {
     }
   }
 
+  componentDidMount = () => {
```
joshblack:
```suggestion
  componentDidMount() {
```

src/components/Modal/Modal.js
```diff
@@ -202,6 +202,22 @@ export default class Modal extends Component {
...
+    if (!this.props.open) {
+      return;
+    }
+    const primaryFocusElement = this.innerModal.current.querySelector(
```
joshblack: Are we guaranteed that the refs will have `current` on them during this pass or do we need to assert they exist before calling/using them?

emyarod: it seems we are not checking if the `current` attribute exists in other parts of the component, and I think it's because we can expect it to be present when we are calling refs?

joshblack: Makes sense @emyarod! Wasn't sure where we were using inline callback refs versus named callback refs (or `React.createRef`) 😅

PR #1625, opened 2018-12-09
fix(build): fix broken UMD build

### LINKED ISSUE #1620: Regression when consuming UMD: ReferenceError: module is not defined
We have some old AnguarJS code that embeds Carbon components using react2angular.
The Karma tests fail with:
```
 "message": "An error was thrown in afterAll\nReferenceError: module is not defined\n    at <Jasmine>\n    at Suite.<anonymous> 
```

This code doesn’t have a webpack build it consumes the following UMDs:
```
      './node_modules/react/umd/react.production.min.js', 
      './node_modules/react-dom/umd/react-dom.production.min.js', 
      './node_modules/classnames/index.js',
      './node_modules/prop-types/prop-types.min.js',
      './node_modules/carbon-icons/dist/carbon-icons.js',
      './node_modules/carbon-components-react/umd/carbon-components-react.js',
```

The problem started in carbon-components-react v6.53.2 after this PR:
https://github.com/IBM/carbon-components-react/pull/1540

asudoh:
NOTE: Depends on IBM/carbon-components#1545 to pass the build (wrt bundle size).

This change inlines `carbon-components` contents, notably CSS prefix constant, into `carbon-components-react` UMD bundle.

Fixes #1620.

#### Changelog

**Changed**

- `carbon-components` is no longer externalized from the UMD bundle.
- `flatpickr` dependency to align with `carbon-components`.


emyarod:
It seems we need to increase our bundle size threshold again for CI check to pass

joshblack:
Seems like `carbon-components` is being included and the size is being reported as the full-size in the UMD build which is causing it. Do we want it embedded? Do we have a choice otherwise?

asudoh:
@joshblack IBM/carbon-components#1545 will fix the bundle size problem. We only need [`settings.js`](https://github.com/IBM/carbon-components/blob/v9.0.0/src/globals/js/settings.js) from vanilla library.

asudoh:
Fixed bundle size problem by upgrading `carbon-components`.

Issue #1619, opened 2018-12-06
[OverflowMenuItem] Ability to render items as links when given href


ZacharyStair:
### Summary

The Dropdown (v1) has a useful feature where passing an `href` prop to the dropdown items would cause them to be rendered as `<a>` tags with the given href, which is handy from a usability standpoint (can middle-click, bookmark, see where a link is going, etc.) vs a button that just navigates via javascript. I'd like that functionality to be available in the OverflowMenu as well, via OverflowMenuItem.

Given that there is a dropdown v2 I imagine the overflow menu might also move towards Downshift (overflow menu v2??), so I understand it may not be worth it to make more changes to this in the short term.

### Specific timeline issues / requests

The sooner the better

### Available extra resources

Conditionally replacing the button with an anchor is something that I would be happy to do, I could even be convinced to make it more general using `children` or something. I don't have the cycles right now to implement a full OverflowMenuWithDownshift though.

asudoh:
AFAIK there is no immediate plan to move `OverflowMenu` to Downshift. That said, choosing button tag in `<OverflowMenuItem>` depending on `href` existence` as you said sounds good to me.

PR #1630, opened 2018-12-10
feat(OverflowMenuItem): renders an anchor when passed href

ZacharyStair:
when an href is passed to an overflowmenuitem it will render an anchor tag instead of a button, so
users can bookmark/hover/etc.

The styling doesn't look quite right because as links the overflow menu items always have underlines... I can make a corresponding change to the `carbon-components` as well

Closes IBM/carbon-components-react#1619

#### Changelog

**New**

- OverflowMenuItem renders `<a>` when passed `href`


ZacharyStair:
@asudoh I was originally going to do it that way but I thought it might be more confusing. Seeing it in action its not too bad.

Issue #1632, opened 2018-12-11
`selectorPrimaryFocus` prop not passed into `Modal` via `ModalWrapper`


emyarod:
## Detailed description

> Describe in detail the issue you're having.

currently if the user wants to define a custom selector for the primary focusable element on a `<Modal>` within a `<ModalWrapper>` using the `selectorPrimaryFocus` prop, the prop is not passed to the `<Modal>` from the `<ModalWrapper>`

> Is this issue related to a specific component?

`<ModalWrapper>`

> What did you expect to happen? What happened instead? What would you like to see changed?

`<ModalWrapper>` should pass the `selectorPrimaryFocus` prop through to the `<Modal>` it contains

PR #1633, opened 2018-12-11
 fix(ModalWrapper): pass `selectorPrimaryFocus` prop through to `<Modal>`

emyarod:
Closes IBM/carbon-components-react#1632

This PR exposes the `selectorPrimaryFocus` prop on `<ModalWrapper>` in order to pass it through to `<Modal>` for configuration

#### Changelog

**New**

- add `selectorPrimaryFocus` prop on `<ModalWrapper>`

**Changed**

- updated Storybook examples

### DIFF
--- a/src/components/ModalWrapper/ModalWrapper-story.js
+++ b/src/components/ModalWrapper/ModalWrapper-story.js
@@ -20,6 +20,10 @@ const props = () => ({
   ),
   modalLabel: text('The modal label (optional) (modalLabel)', 'Label'),
   modalHeading: text('The modal heading (modalHeading)', 'Modal'),
+  selectorPrimaryFocus: text(
+    'Primary focus element selector (selectorPrimaryFocus)',
+    '[data-modal-primary-focus]'
+  ),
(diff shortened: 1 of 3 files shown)


src/components/ModalWrapper/ModalWrapper-story.js
```diff
@@ -106,8 +110,7 @@ storiesOf('ModalWrapper', module)
         <TextInput
           id="test2"
           placeholder="Hint text here"
-          label="Text Input:"
-          data-modal-primary-focus
```
asudoh: Just wanted to double-check if removal of `data-modal-primary-focus` here is intentional.

emyarod: I removed it to keep it consistent with the other storybook examples with the `selectorPrimaryFocus` prop. I believe none of the other examples contain this data attribute so it defaults to the focusing on primary button. If the user wants to test it autofocusing on a different element they can use the Storybook knob to choose a different selector

Issue #1623, opened 2018-12-07
Support primary focus element behavior in ComposedModal


emyarod:
### Summary

`<Modal>` currently allows the user to define a primary focus element on render/modal open, but `<ComposedModal>` does not share this feature.

### Justification

Improved keyboard navigation and a11y

### Desired UX and success metrics

when a `<ComposedModal>` is opened or rendered with the `open` prop set to `true` the primary actionable element should be focused by default, but the user can also define their own selector for calling focus.

PR #1634, opened 2018-12-11
feat(ComposedModal): support selectorPrimaryFocus prop

emyarod:
Closes IBM/carbon-components-react#1623

This PR adds support for the `selectorPrimaryFocus` prop on `<ComposedModal>` to match `<Modal>` and `<ModalWrapper>`

Related: #980 #1124 #1621 #1633

#### Changelog

**New**

- focus management methods (many shared with `<Modal>`)
- pass ref for primary focusable element to `<ComposedModal>`s using `<ModalFooter>`
- storybook example `<ComposedModal>` with trigger button

**Changed**

- declare `<Modal>` methods before they are called

### DIFF
--- a/src/components/ComposedModal/ComposedModal-story.js
+++ b/src/components/ComposedModal/ComposedModal-story.js
@@ -10,10 +10,14 @@ import ComposedModal, {
 import Button from '../Button';
 
 const props = {
-  composedModal: () => ({
-    open: boolean('Open (open in <ComposedModal>)', true),
+  composedModal: (includeOpen = true) => ({
+    open: includeOpen ? boolean('Open (open in <ComposedModal>)', true) : null,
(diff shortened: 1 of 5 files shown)


joshblack:
Related to focus, would we ever want to have our patterns follow what we do in UI Shell:

https://github.com/IBM/carbon-components-react/blob/fb398145c254117c915d1184b8baed0e638c133c/src/components/UIShell/HeaderMenu.js#L246-L262

e.g. basically make it so that children need to support a `ref` prop and we use that reference on the first item to focus the node.

emyarod:
sure, I have updated with some test cases for `selectorPrimaryFocus` @asudoh

asudoh:
Thanks you for adding the tests @emyarod!

Issue #1653, opened 2018-12-13
Experimental Component: NumberInput


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1203

https://github.com/IBM/carbon-components/issues/1210

https://github.com/IBM/carbon-components/pull/1281

https://github.com/IBM/carbon-components/issues/1557

https://github.com/IBM/carbon-components/pull/1558

PR #1654, opened 2018-12-13
feat(NumberInput): add experimental NumberInput

### LINKED ISSUE #1710: Form validation UI content text should not influence styling of number input up/down buttons
## When utilizing the number input, but stripping out the helper text/label/Form validation UI content, but clicking on the form validation, the number buttons are out of place int he input field

> See picture

> Is this issue related to a specific component? number input

I'd expect given the content the number input buttons not to have it's styling influenced

> What browser are you working in? Chrome Version 71.0.3578.98 (Official Build) (64-bit)

> What version of the Carbon Design System are you using?     carbon-components-react": "^6.58.0",

Yes
See comments/screen capture
image

emyarod:
Closes IBM/carbon-components-react#1653
Closes #1710 
Closes https://github.com/carbon-design-system/carbon-website/issues/827

This PR will arrange the NumberInput to match the vanilla spec. The `form-item` div wrapper is removed as we have an explicit `<FormItem>` component in the library that already accomplishes this.

This PR also ties `no-label` classes to the value of the `label` prop in experimental mode rather than the explicit `hideLabel` prop. Looking for feedback on this change which would eventually be reflected across the library as well.

#### Changelog

**Changed**

- use updated markup
- tie `no-label` classes to the value of `label` rather than explicit `hideLabel` prop

**Removed**

- unit test searching for `<FormItem>` class


asudoh:
Thanks @emyarod for this proposal! As I DM'ed, great to see `hideLabel` removal is done across the board and probably de-coupled from this PR.

emyarod:
issue opened @asudoh https://github.com/IBM/carbon-components-react/issues/1662

I will begin development in a separate branch

joshblack:
Do we need to guard these changes behind a feature flag at all? Was going through it quickly and that question came to mind given the changes, just wanted to ask!

src/components/NumberInput/NumberInput.js
```diff
@@ -252,43 +252,41 @@ export default class NumberInput extends Component {
     ) : null;
 
     return (
-      <div className={`${prefix}--form-item`}>
```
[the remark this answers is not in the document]
emyarod: yeah I think I can avoid the breaking change for now, just spotted inconsistency in using `<FormItem />` across components and storybook examples. I can open this discussion in a separate issue if that would be better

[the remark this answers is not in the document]
emyarod: created an issue to track this here https://github.com/IBM/carbon-components-react/issues/1681

src/components/NumberInput/NumberInput.js
```diff
@@ -110,7 +110,7 @@ export default class NumberInput extends Component {
     disabled: false,
     hideLabel: false,
     iconDescription: 'choose a number',
-    label: ' ',
+    label: '',
```
asudoh: This looks a breaking change, do we want this change for experimental mode only...?

asudoh: @emyarod Thoughts...? Thanks!

emyarod: updated!

emyarod:
possibly dependent on https://github.com/IBM/carbon-components/pull/1604

emyarod:
depends on https://github.com/IBM/carbon-components/issues/1621

emyarod:
PR has been updated to reflect changes from https://github.com/IBM/carbon-components/pull/1622

emyarod:
I thought about that @asudoh but wasn't sure if it was easier to read. I went ahead with the change though

asudoh:
Seems that all comments have been addressed, great to have another review so we can go forward with this. Thanks!

asudoh:
Thank you for reviewing @lovemecomputer!

PR #1656, opened 2018-12-13
fix(dependencies): upgrade @carbon/icons-react

### LINKED ISSUE #1624: http://react.carbondesignsystem.com doesn't work at all in edge browser
Visit http://react.carbondesignsystem.com/ using an edge browser

## Detailed description

Does carbon support the edge browser?  Some ibm customers might utilize products with edge?

asudoh:
The latest version contains IE11/Edge fix.

Fixes #1624.

#### Changelog

**Changed**

- Upgrade `@carbon/icons`.


joshblack:
@asudoh I think CI fails because the snapshot changes with the new `style` prop 👍

asudoh:
@joshblack Thank you for pointing it out!

Issue #1650, opened 2018-12-12
Aria-Expanded is Missing from Accordion


frastlin:
## Detailed description

Hello, I am a screen reader user and was looking at the [code examples](http://react.carbondesignsystem.com/?selectedKind=Accordion&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) and the accordions do not properly handle the aria-expanded attribute. Also, it's [good practice to have accordions be button elements](https://inclusive-components.design/collapsible-sections/). Currently they are delineated as tabs which is really confusing because tabs and accordions are different. Tabs show information below the tab list and can be navigated and selected using the arrow keys. Accordions use the expanded and collapsed attribute and the content is inserted below the expanded element.
For a very good example of accordions, see [this example from AccDC](http://whatsock.com/tsg/Coding%20Arena/ARIA%20and%20Non-ARIA%20Accordions/Accordion%20(Internal%20Content)/demo.htm)

## Steps to reproduce the issue

1. Using [NVDA](https://www.nvaccess.org/download/) on Chrome Version 71.0.3578.80 (Official Build) (64-bit) visit the [accordion example](http://react.carbondesignsystem.com/?selectedKind=Accordion&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)
2. hit enter on the first: "Expand/Collapse Section 1 title"

## What Happened

nothing. NVDA says nothing when enter is pressed on the top collapsible component.

## What Should Happen

NVDA should say "Expanded" and the next press of enter NVDA should announce "collapsed" (Although for some reason enter doesn't seem to collapse the component, it's probably because it's not a button and is not getting the key press).

emyarod:
hi @frastlin thanks for reporting this issue! I've pushed a quick fix to add `aria-expanded` to the accordion [here](https://github.com/IBM/carbon-components-react/pull/1657), but I am still in the process of setting up NVDA on my machine. Can you confirm whether or not the component is improved at this [staging link](https://deploy-preview-1657--carbon-components-react.netlify.com/)? 

You also brought up a good point about our usage of the `tab` role right now, and while the pull request I referenced doesn't address it yet, it is also something I would like to explore further. But first I want to make sure that the component is detectable by screen readers after this change

frastlin:
[...]

PR #1657, opened 2018-12-13
fix(AccordionItem): add aria-expanded and use tab role

emyarod:
Closes IBM/carbon-components-react#1650

This PR uses a more appropriate `role` for `<AccordionItem>`s with respect to our current markup and includes the `aria-expanded` on the component. This change is meant to support screen readers but exposes a potential larger issue with the elements used to construct our `<Accordion>`s

#### Changelog

**Changed**

- use `tab` role for `<AccordionItem>` to correspond with the `role` on `<Accordion>`
- tie `aria-expanded` state to component open/closed state

### DIFF
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -8,11 +8,7 @@ const { prefix } = settings;
 const Accordion = ({ children, className, ...other }) => {
   const classNames = classnames(`${prefix}--accordion`, className);
   return (
-    <ul
-      className={classNames}
-      role="tablist"
-      aria-multiselectable="true"
-      {...other}>
+    <ul className={classNames} {...other}>
       {children}
     </ul>
   );
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -8,7 +8,7 @@ const { prefix } = settings;
 const Accordion = ({ children, className, ...other }) => {
   const classNames = classnames(`${prefix}--accordion`, className);
   return (
-    <ul className={classNames} {...other}>
+    <ul {...other} className={classNames}>
       {children}
     </ul>
   );
--- a/src/components/AccordionItem/AccordionItem.js
+++ b/src/components/AccordionItem/AccordionItem.js
@@ -129,7 +129,6 @@ export default class AccordionItem extends Component {
             icon={iconChevronRight}
             description={iconDescription}
             role={null} // eslint-disable-line jsx-a11y/aria-role
-            aria-hidden={true}
           />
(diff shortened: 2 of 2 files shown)


src/components/AccordionItem/AccordionItem.js
```diff
@@ -117,7 +117,8 @@ export default class AccordionItem extends Component {
         className={classNames}
         onClick={this.handleClick}
         onKeyPress={this.handleKeyPress}
-        role="presentation"
+        role="tab"
+        aria-expanded={this.state.open}
```
asudoh:
```suggestion
        aria-expanded={String(Boolean(this.state.open))}
```

src/components/AccordionItem/AccordionItem.js
```diff
@@ -121,13 +121,15 @@ export default class AccordionItem extends Component {
...
           type="button"
+          aria-expanded={this.state.open}
           className={`${prefix}--accordion__heading`}
-          role="tab"
           onClick={this.handleHeadingClick}>
           <Icon
             className={`${prefix}--accordion__arrow`}
             icon={iconChevronRight}
             description={iconDescription}
+            role={null} // eslint-disable-line jsx-a11y/aria-role
```
joshblack: Do we need the `role` here if we have `aria-hidden`?

src/components/AccordionItem/AccordionItem.js
```diff
@@ -121,13 +121,15 @@ export default class AccordionItem extends Component {
...
+            aria-hidden={true}
```
joshblack: Could do a standalone `aria-hidden` if we want to use the shorthand form (right?)

src/components/Accordion/Accordion.js
```diff
@@ -8,11 +8,7 @@ const { prefix } = settings;
 const Accordion = ({ children, className, ...other }) => {
   const classNames = classnames(`${prefix}--accordion`, className);
   return (
-    <ul
-      className={classNames}
-      role="tablist"
-      aria-multiselectable="true"
-      {...other}>
+    <ul className={classNames} {...other}>
```
joshblack: Should we use this PR to swap the order of how the props get spread? e.g.

```suggestion
    <ul {...other} className={classNames}>
```

src/components/AccordionItem/AccordionItem.js
emyarod: yup it's because we don't need `role` that it is being set to `null` here, since `Icon` will assign it a default value

joshblack: Makes sense!

Issue #1660, opened 2018-12-14
Add Escape to Close Accordion Tabs


frastlin:
### Summary

Please have Escape close accordions. The [accordion Example](https://github.com/IBM/carbon-components-react/issues/1650) does not seem to have this functionality.

### Justification

Keyboard users with the focus in the middle of expanded content need to navigate all the way back to the expanded button before they can close the accordion. This could take a while if the accordion is full of links or other elements placed in the tabindex.

### Desired UX and success metrics

When the user presses escape while a collapsible accordion is open, collapse the accordion.
If one accordion needs to be expanded at all times, then escape does nothing.

### "Must have" functionality

Escape should close any collapsible accordion, preferably only when escape is pressed in the expanded content or expanded button.

### Specific timeline issues / requests

(Would be nice to have)

### Available extra resources

Once I get to know the code I can probably figure out how to do it. The most difficult part will be to figure out where the onKeyDown handler needs to go.

PR #1661, opened 2018-12-14
feat(Accordion): close accordion item on Esc keypress

emyarod:
Closes IBM/carbon-components-react#1660

This PR adds support for closing an open `AccordionItem` with the `Esc` key. The current keypress handler also does not trigger the opening and closing of `AccordionItem`s. Due to that behavior already being native to button elements, it does not look like we need to handle space bar and Enter key presses separately.

#### Changelog

**Changed**

- listen to `onKeyDown` event rather than `onKeyPress` to support `Esc` key

**Removed**

- listeners for space bar and enter key

### DIFF
--- a/src/components/AccordionItem/AccordionItem.js
+++ b/src/components/AccordionItem/AccordionItem.js
@@ -84,11 +84,9 @@ export default class AccordionItem extends Component {
     this.props.onHeadingClick({ isOpen: open, event: evt });
   };
 
-  handleKeyPress = evt => {
-    const isKeyPressTarget = evt.target === evt.currentTarget;
-    const isValidKeyPress = [13, 32].indexOf(evt.which) !== -1;
-
-    if (isKeyPressTarget && isValidKeyPress) {
+  handleKeyDown = evt => {
+    // Esc key
+    if (evt.which === 27 && this.state.open) {
       this.handleHeadingClick(evt);
     }
   };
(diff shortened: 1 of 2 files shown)


src/components/AccordionItem/AccordionItem.js
```diff
@@ -84,11 +84,13 @@ export default class AccordionItem extends Component {
...
 
-  handleKeyPress = evt => {
-    const isKeyPressTarget = evt.target === evt.currentTarget;
-    const isValidKeyPress = [13, 32].indexOf(evt.which) !== -1;
-
-    if (isKeyPressTarget && isValidKeyPress) {
+  handleKeyDown = evt => {
+    // Esc key
+    if (
+      evt.which === 27 &&
```
joshblack: Could use the helpers in: https://github.com/IBM/carbon-components-react/blob/master/src/tools/key.js if they'd be helpful, e.g.

```js
if (match(evt, keys.ESC)) {
    // ...
}
```

emyarod: wasn't aware we had these helpers, thanks!

emyarod:
hi @asudoh the enter/space bar behavior is not being removed. As far as I can tell, that behavior is handled natively by the `<button>` element, and the event handler logic I removed was not being called

Issue #1662, opened 2018-12-14
Remove `hideLabel` props


emyarod:
### Summary

Currently we require users to assign a boolean value to `hideLabel` on form items with `<label>`s, but we can streamline our exposed props by removing this prop altogether and tying the appearance of a `<label>` to the value of a `<label>` (i.e. empty string can remove the rendering of a `<label>`)

asudoh:
@emyarod Are you still interested in tackling this? Thanks!

emyarod:
yes @asudoh, should we introduce this change behind breaking change flag?

asudoh:
Yes, probably `FeatureFlags.autoHideEmptyLabel` or a like.

PR #1663, opened 2018-12-14
[WIP] fix: assign class names based on form item labels

emyarod:
Closes IBM/carbon-components-react#1662

Related/depends on: #1654

This PR will remove the `hideLabel` prop across the library in favor of rendering `<label>`s based on the text content of labels

#### Changelog

**Changed**

- `<NumberInput>`

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -4,6 +4,7 @@ import { iconCaretUp, iconCaretDown } from 'carbon-icons';
 import Icon from '../Icon';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
+import { componentsX } from '../../internal/FeatureFlags';
 
 const { prefix } = settings;
 
@@ -110,7 +111,7 @@ export default class NumberInput extends Component {
     disabled: false,
     hideLabel: false,
     iconDescription: 'choose a number',
-    label: ' ',
+    label: '',
     onChange: () => {},
     onClick: () => {},
     step: 1,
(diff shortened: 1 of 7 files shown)


src/components/NumberInput/NumberInput.js
```diff
@@ -4,6 +4,7 @@ import { iconCaretUp, iconCaretDown } from 'carbon-icons';
 import Icon from '../Icon';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
+import { componentsX } from '../../internal/FeatureFlags';
```
asudoh: Wondering if you can port `breakingChangesX` flag over this PR and use it, given the change doesn't seem to be tied to experimental theme: https://github.com/IBM/carbon-components-react/pull/1606/files#diff-860a64fa9638c1ac3f8ec09230fc26a9R21

src/components/NumberInput/NumberInput.js
```diff
@@ -252,43 +255,41 @@ export default class NumberInput extends Component {
...
-            type="number"
-            pattern="[0-9]*"
-            {...other}
-            {...props}
-            ref={this._handleInputRef}
-          />
-          {error}
-          {helper}
+      <div className={numberInputClasses} {...inputWrapperProps}>
+        {labelText}
```
asudoh: Is the element move needed for this change?

src/components/NumberInput/NumberInput.js
```diff
@@ -252,43 +255,41 @@ export default class NumberInput extends Component {
     ) : null;
 
     return (
-      <div className={`${prefix}--form-item`}>
```
asudoh: This wrapper `<div>` removal is a breaking change and need to be roped off. As I noted elsewhere, discussion with TJ is nice here - Would we anticipate one `<FormItem>` containing more than one form controls, or usage of this component without form wrapper...?

joshblack:
_PR Reviews 1/11_

Notes

- Going to close this for now since it isn't for X

Issue #1649, opened 2018-12-12
Experimental Component: Modal


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1283

https://github.com/IBM/carbon-components/issues/1284

https://github.com/IBM/carbon-components/issues/1302

https://github.com/IBM/carbon-components/pull/1488

PR #1667, opened 2018-12-18
feat(Modal): add experimental modal

emyarod:
Closes IBM/carbon-components-react#1649

This PR will add the new experimental modal and experimental composed modal changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -354,5 +354,5 @@
       "enzyme-to-json/serializer"
     ]
   },
-  "bundleSizeThreshold": 82000
+  "bundleSizeThreshold": 85000
 }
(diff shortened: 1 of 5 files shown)


emyarod:
it looks like the `@carbon/icons-react` import is blowing up the bundle size. is there a configuration setting I need to tweak to get proper tree shaking? it looks like the new icons are already being used elsewhere without affecting the bundle size this significantly

joshblack:
@emyarod in the interim, can use the direct import path:

```js
 import { Close20 } from '@carbon/icons-react/lib/close/20';
```

Most likely what is happening is that the modules are considered as having side effects and so they are being initialized without being tree-shook 😞

It's not impacting our bundles in other components just because they aren't being imported in the entrypoint yet. UI Shell is just in its own folder right now.

emyarod:
thanks for the clarification @joshblack ! I'll use the direct import path for now

Issue #1672, opened 2018-12-18
Experimental Component: Code snippet


emyarod:
While the markup has not changed for the v10 code snippet component, the icons need to be updated

Issue #1673, opened 2018-12-18
Experimental Component: Copy button


emyarod:
The copy button icons need updating, blocks #1672

PR #1675, opened 2018-12-18
feat(CodeSnippet): add support for experimental icons

emyarod:
Closes IBM/carbon-components-react#1672

This PR will add the new experimental code snippet changes

depends on https://github.com/IBM/carbon-components-react/issues/1673 and https://github.com/IBM/carbon-components/issues/1572


src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -139,13 +142,23 @@ export default class CodeSnippet extends Component {
...
-          aria-hidden="true"
-          alt={expandCodeBtnText}
-          name="chevron--down"
-          description={expandCodeBtnText}
-          className={`${prefix}--icon-chevron--down`}
-        />
+        {componentsX ? (
+          <ChevronDown16
+            aria-hidden={true}
+            aria-label={expandCodeBtnText}
```
joshblack: Do we need the `aria-label` if `aria-hidden` is set to `true`? I bet we could use the content from the previous span for the `button` instead of using the icon.

emyarod:
yes @asudoh it should be good to go

Issue #1682, opened 2018-12-21
Experimental Component: Slider


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1181

https://github.com/IBM/carbon-components/issues/1238

https://github.com/IBM/carbon-components/issues/1265

https://github.com/IBM/carbon-components/pull/1315

Issue #1681, opened 2018-12-21
Favor explicit usage of <FormItem> component


emyarod:
Currently many form item components contain a wrapper div within them with the `.bx--form-item` class. When composing forms, these wrapper divs are often duplicated or erroneously added, causing components to diverge from the spec. Examples of this can be seen in the current `<NumberInput>` and `<Slider>` stories

emyarod:
stalebot pls

PR #1683, opened 2018-12-21
feat(Slider): add experimental slider

emyarod:
Closes IBM/carbon-components-react#1682

This PR will add the new experimental slider changes

related https://github.com/IBM/carbon-components-react/issues/1681


src/components/Slider/Slider.js
```diff
@@ -154,12 +153,6 @@ export default class Slider extends PureComponent {
     };
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (!sliderValuePropSync && !isEqual(nextProps, this.props)) {
```
asudoh: The removal here seems a breaking change, please see: #1370

src/components/Slider/Slider.js
```diff
@@ -417,13 +407,16 @@ export default class Slider extends PureComponent {
...
+            {formatLabel(min, minLabel)}
+          </span>
           <span className={`${prefix}--slider__range-label`}>
             {formatLabel(max, maxLabel)}
           </span>
           {!hideTextInput && (
             <TextInput
               type={inputType}
-              id="input-for-slider"
+              id={`input-for-slider${id}`}
```
asudoh: We tend to use `${id}` as a prefix instead of a suffix (e.g. https://github.com/IBM/carbon-components-react/blob/v6.39.0/src/components/Select/Select.js#L31) - Wondering if it'd make sense to align the convention.

emyarod:
@asudoh probably is not blocked but the topic discussed in #1681 is still something to consider in the future I believe

emyarod:
@asudoh the visual bug we are seeing in the React implementation is caused by the problem addressed in https://github.com/IBM/carbon-components-react/issues/1681 and https://github.com/IBM/carbon-components-react/pull/1684

Our React `<TextInput>` includes a form item/`<FormItem>` wrapper which affects the styles that should be applied to the input field

so we end up rendering

```html
<div class="bx--form-item">
  <input>
</div>
```

rather than just

```html
<input>
```

asudoh:
@emyarod As we discussed the slider element order issue is caused by https://github.com/IBM/carbon-components/pull/1315/files#diff-f8f1a0577791a85d255a9cdf4e2f791fR30 - Did you see it differently, or were you talking about another issue? Any more background?

emyarod:
@asudoh from what I can tell, the issue is caused by our `<TextInput>` component's wrapper div.

https://github.com/IBM/carbon-components-react/blob/master/src/components/Slider/Slider.js#L464

https://github.com/IBM/carbon-components-react/blob/master/src/components/TextInput/TextInput.js#L77

removing this wrapper causes the input box to be positioned correctly

asudoh:
Thank you for the details @emyarod! I think the last link explains the issue the best. That said, when you remove the wrapper div, would you try adding `display:flex` to the new parent node in DOM inspector and see what happens?

asudoh:
@emyarod Thank you for taking a look - Further investigation revealed that I was confused [`div.bx--form-item` around text input](https://github.com/IBM/carbon-components-react/blob/v6.74.0/src/components/TextInput/TextInput.js#L77) vs. [`div.bx--form-item` around slider](https://github.com/IBM/carbon-components-react/blob/v6.74.0/src/components/Slider/Slider.js#L408), my apologies.

From that, I see two approaches to fix the problem:

1. Direct usage of `<input>` instead of using `<TextInput>`
2. Adding a CSS class for ordering `<TextInput>` within [`div.bx--form-item` around slider](https://github.com/IBM/carbon-components-react/blob/v6.74.0/src/components/Slider/Slider.js#L408)

Thoughts...?

emyarod:
@asudoh I am leaning towards option 2 right now, but eventually I would like to address this in a future release after reopening the discussion on https://github.com/IBM/carbon-components-react/pull/1684

emyarod:
blocked by https://github.com/IBM/carbon-components/pull/1640

emyarod:
updated to include changes from https://github.com/IBM/carbon-components/pull/1640

Issue #1686, opened 2018-12-21
Experimental Component: Text input


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1231

https://github.com/IBM/carbon-components/commit/f60206c2ef43b12ccf95a78c71e474d4c7dc3fd7#diff-29f8eb16b4ceb64578a7038fc5addc13

PR #1687, opened 2018-12-21
fix(TextInput): match vanilla markup

emyarod:
Closes IBM/carbon-components-react#1686

This PR will add the new experimental text input changes


emyarod:
yep `9.x` styles seem to be unchanged @asudoh

Issue #1697, opened 2019-01-02
The tooltip component needs an aria-label on the div tag with the roll of button


zgrossbart:
In the tooltip component (http://react.carbondesignsystem.com/?selectedKind=Tooltip&selectedStory=default%20%28bottom%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) we show an icon by default.  That icon gets an `aria-label` attribute, but the div containing the icon does not.  The div tag containing the icon needs an `aria-label` attribute because it has a role of button and that makes it an interactive element.

All interactive elements must have an accessible name:  https://aat.w3ibm.mybluemix.net/token/334ed6c7-0a0d-480b-a007-47d1242d0ef8/1c0df9a2-5788-4caa-8052-67d0bcadfdac/archives/2018OctDeploy//doc/w3/help/en-US/idhi_accessibility_check_g1156.html

PR #1698, opened 2019-01-02
fix: Adds an aria-label to the div surrounding the tooltip icon, reso…

zgrossbart:
…lves #1697.

Closes IBM/carbon-components-react #1697 

This is a small change to add the `aria-label` attribute in the parent div element when it is an interactive element.

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -390,6 +390,7 @@ export default class Tooltip extends Component {
                 onFocus={evt => this.handleMouse(evt)}
                 onBlur={evt => this.handleMouse(evt)}
                 aria-haspopup="true"
+                aria-label={iconDescription}
                 aria-expanded={open}
                 {...ariaOwnsProps}>
                 <Icon


joshblack:
@zgrossbart should the content of the icon populate the contents for the node? For example, if I had the following markup:

```html
<button type="button">
  <svg width="24" height="24" viewBox="0 0 25 24" role="img" aria-label="Hi there">
    <circle cx="12" cy="12" r="12"></circle>
  </svg>
</button>
```

In Chrome, it would be read as `Hi there`:

image

Just wanted to check to see if this is really the issue, where we are relying on internals to be read, when in reality we need `aria-label` on the parent.

zgrossbart:
The screen readers need the `aria-label` on the parent because that is the interactive element.

joshblack:
@zgrossbart what I'm asking about is when we need to include `aria-label`/`aria-labelledby`, versus when it is read from content.

If we took the button example from above and made it: `<button>Submit</button>` would you agree this would not need an `aria-label` or `aria-labelledby` attribute? e.g. `<button aria-label="Submit">Submit</button>`

zgrossbart:
We do need the icon to have some kind of either `aria-label` or `aria-labelledby`.  However, if we use `aria-labelledby` then we need an id for the button element.  I don't think we really need this for screen readers, but we do need this for the DAP tools.

joshblack:
Yeah, was thinking that was the case and it might be missing `role="img"` or something similar which is causing the issue. 

I think this is a great interim fix though while we figure it out, thanks!

zgrossbart:
Thanks @joshblack.  Are we waiting on something else before merging this?

PR #1701, opened 2019-01-02
fix: Set a tab index on the anchors for the tab components so it won'…

### LINKED ISSUE #1548: A11Y: Tabs have more than one tabbable element.
## Detailed description

Error in DAP accessibility test:
screen shot 2018-11-09 at 11 30 46 am

This is being caused by each of the `<a>` tags having a tabIndex of 0. Only the tab that is currently selected is supposed to have a tabIndex of 0. All others are supposed to be -1.

`<Tabs>`

Chrome

Link to a11y spec for tabbed indexes: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex

Tested from storybook: http://react.carbondesignsystem.com/iframe.html?selectedKind=Tabs&selectedStory=Default

zgrossbart:
…t be zero, resolves #1548.

Closes IBM/carbon-components-react#1548

This just makes sure that we set a valid tab index for the anchor for each tab instead of  always using zero.


asudoh:
Hi @zgrossbart 👋 thank you for jumping in! It appears that the fix seems different from what the issue requests for. Any background on why this PR took a different route?

zgrossbart:
I'm sorry.  I should have made that clearer.  I did some more research and this looked like a better fix.  It makes the tab order work in a clearer way and it fixes the problem with the accessibility tools.

zgrossbart:
@rob-richards is on my team and would be happy with this fix.

asudoh:
Hi @zgrossbart your latest fix seems to make the tab navigation rely solely on left/right keys, which make things weird when tabs turns to drop down with smaller real screen estate. Meanwhile, would your team want to try [`renderAnchor` prop of `<Tab>`](https://github.com/IBM/carbon-components-react/blob/v6.40.0/src/components/Tab/Tab.js#L100)? Thanks!

asudoh:
@zgrossbart I see your intent. Just in case, my point was that the tab turns to a drop down when you make browser window smaller.

`renderAnchor` gives you full control over how `<a>` in `<Tab>` is rendered, like `tabIndex` you wanted to change. More details can be found at [here](https://github.com/IBM/carbon-components-react/blob/v6.40.0/src/components/Tab/Tab.js#L100).

zgrossbart:
Thanks @asudoh.  That makes sense.  However, in this case we're just following the guidance for tab panels.  Are you you OK with this change?  What else do we have to do before we can merge?

asudoh:
@zgrossbart We need to have keyboard navigation for dropdown before we go forward with this PR. I see it'll be much bigger effort than what PR attempts to do. That's why I suggested user-land solution.

zgrossbart:
Thank you for the explanation.  This change does give keyboard access with the right and left arrows when the tab panel is rendered as a select list.  That's better than it was before when you would have had to use the tab key and press enter to get the next option.  We could add more handling to make the keyboard access better in the select list case, but that feels different from this issue.  This PR doesn't make that better, but it doesn't make it worse.

asudoh:
@zgrossbart If you make the window smaller, you'll see the dropdown list. With your change, left/right arrow keys now is the only way to navigate through drop down items, which is not something user would expect, whereas user can hit tab keys without your change.

zgrossbart:
@asudoh, yes, but a user wouldn't expect to use the tab keys to navigate the options in a select list either.  They would expect to use the up and down arrows.

zgrossbart:
If you don't feel like this is the correct fix we can look into making this change only when the tabs are acting like a tab panel, or making the up and down arrow keys work.

zgrossbart:
I'm sorry @asudoh, but which option sounds great?  Do we want to make this change?  Have it only apply in the tab panel case?  Try to enable the up and down arrows?  Thanks.

asudoh:
@zgrossbart The former needs to happen at least. The latter sounds nice as an addition, but it should happen only with narrow-screen mode.

zgrossbart:
@asudoh, I've changed the way this fix works.  Now we will make the `tabIndex` 0 and -1 when the panel acts like a tab panel.  When the panel acts like a select list we will make all of the `tabIndex` values 0 when the select list appears.  

This means that when the tabs look like a tab panel a user can use the tab key to put focus in the tab panel and then use the left and right arrow keys.  When the tabs look like a select list the user can use the tab key to select options in the list and doesn't need to use the arrow keys.

asudoh:
@zgrossbart Your latest change made things a bit better by supporting tab navigation in narrow screen mode. Still seeing left/right arrow keys affecting the narrow screen (dropdown) behavior, which needs to be addressed - Thanks!

zgrossbart:
Thank you for taking a look @asudoh.  Can you tell me more about what you're seeing?  When I tested this I saw that you could use the right and left arrows in the wide screen mode and the tab and enter key in the narrow screen mode.  Are you seeing something different?

zgrossbart:
Thank you @asudoh.  The right and left arrow keys are disabled in the narrow screen mode.  I can see it here:  https://deploy-preview-1701--carbon-components-react.netlify.com/?selectedKind=Tabs&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel.  We set the tab index when the drop down is opened so you can get the tab key to change selection.

asudoh:
@zgrossbart Trying now - If you open the dropdown in narrow screen mode, hit tab once, and hit left/right arrow keys, you see the tab selection is changed. Such tab selection change shouldn't happen.

zgrossbart:
Thank you for the clarification @asudoh.  You're right that the right and left arrow keys do change the selection, but that it because you are in the tab panel at that point.  The same thing happens on the master branch here:  https://react.carbondesignsystem.com/iframe.html?selectedKind=Tabs&selectedStory=Default#  This PR doesn't change that behavior.  It just fixes the accessibility issue with the wide screen.

The real issue here is that the tab panel looks like a select list when the screen is narrow, but that was just done with CSS.  It doesn't act entirely like a select list, but it's close.  This PR doesn't change that.  It just makes the behavior more accessible to visually impaired users.

zgrossbart:
Thank you very much @asudoh.

Issue #1583, opened 2018-11-26
A11Y: ExpandableTile a11y error when content contains links


ZachStowellIBM:
## Detailed description

> Describe in detail the issue you're having.

`<ExpandableTile>` has a role of button, which causes an a11y error if there are links inside.

> Is this issue related to a specific component?

ExpandableTile component.

> What did you expect to happen? What happened instead? What would you like to see changed?

Expected to not have an a11y error on the expandable tile when there is content inside that are links.  

Instead, a11y tests showed a "Widget with role of button mast have no more than one tabbable element."

Removing the `role="button"` from the expandableTile will resolve the issue, however this may not be a good practice since the entire expandable tile is clickable, and therefore _could_ be considered a button.

> What browser are you working in?

Chrome 70.0.3538.102 on Mac

> What version of the Carbon Design System are you using?

carbon-components: ^9.18.0
carbon-components-react: 6.39.3

## Additional information

- [WAI-ARIA Button Docs](https://www.w3.org/TR/wai-aria-practices-1.1/#button)
- Tested on a modified version of the ExpandableTile [Storybook example](http://react.carbondesignsystem.com/iframe.html?selectedKind=Tile&selectedStory=Expandable)

Screenshot of modifications/a11y warning:
![screen shot 2018-11-26 at 12 01 32 pm](https://user-images.githubusercontent.com/39738800/49029865-c4cb7000-f173-11e8-9eea-55046e5c208a.png)

PR #1702, opened 2019-01-02
fix: Removed the role of button from the expandable tile, resolves #1…

zgrossbart:
…583.

Closes IBM/carbon-components-react#1583

This component was assigning the role of button to a div tag that was clickable, but it didn't need the roll of button.  The div contains a button object with the same functionality.  That is the element we want to make available for screen readers.


joshblack:
Will we run into an issue where we have `tabIndex` and `onClick` specified for a non-interactive element?

PR #1703, opened 2019-01-03
Adds an option to add `onChange`

uridabomb:
Adds an option to add `onChange` prop for the input (checkbox) element and sets it to an empty function by default.

#### Changelog

**New**

- Adds an option to add `onChange` prop for the input (checkbox) element.
- Sets it as empty function by default for avoiding/removing the following warning:
```
Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
```


src/components/InlineCheckbox/InlineCheckbox.js
```diff
@@ -87,6 +92,7 @@ export default class InlineCheckbox extends React.Component {
       ref: this.handleRef,
       checked: false,
       disabled,
+      onChange: onChange
```
emyarod: small nitpick: we can probably use the object property value shorthand to remain consistent with the rest of the codebase

emyarod: small nitpick: we can probably use the object property value shorthand to remain consistent with the rest of the codebase

joshblack:
This is coming from usage alongside `DataTable`, correct? Might be worth updating the usage in `TableSelectRow` and `TableSelectAll` which are calling `InlineCheckbox` to set `defaultChecked` or to use `onChange` as appropriate 👍

uridabomb:
@joshblack Sounds good to me. This indeed comes from usage alongside `DataTable`.

joshblack:
Looking at this again, I think the fundamental issue is that `InlineCheckbox` is using `onClick` when it should be `onChange`. What do you all think?

asudoh:
Closing as #1824 seems to have introduced a similar change. Our apologies to getting back to this late @uridabomb!

Issue #1140, opened 2018-07-30
Combined Date/Time Selector


Kyle-Cooley:
## Detailed description

> Describe in detail the issue you're having.

For the project I am working on, we need a combined dropdown selector for date and time. As far as I can tell, this does not appear to exist in carbon. This seems like something that could be useful for multiple projects. Just wondering if there was an interest in adding/amending a component for this use case.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Feature request

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

APM

## Additional information

The DatePicker component could enable this, since flatpickr has properties for this. They just need to be exposed to the users of carbon. There are a couple of css issues that would need resolved, but it is an option. I'm just not sure whether or not it would fit with the carbon design.

![image](https://imgur.com/HEByquy.png)

asudoh:
@aagonzales @designertyler Is this usage pattern something we can allow? Thanks!

aagonzales:
Yeah I think this could be a good pattern. My only concern/feedback is figuring out how to show the selected time in the closed view so its still visible to user when closed. Maybe inline or stacked with the date?

ShuangLiLi:
@aagonzales We want a range date picker with time selection, how about the schedule? We are urgent about the new version or workaround. Thanks!

asudoh:
Hi @ShuangLiLi though I don't think this feature is in our imminent roadmap due to the number of teams asking for this - You can create a custom React component with Carbon style and some custom styles (and possibly Flatpickr our component libraries use). PRs are appreciated if you come up with a component that aligns to Carbon design well and is solid and flexible enough for many teams to use. Thanks!

ShuangLiLi:
@aagonzales @asudoh  Hi, I also use flatpickr to support range date picker can select time, but how can I show the time in label with date?

Also, I have another question about how to set a default date and time for a DatePickerInput?

Thanks a lot for your help!

diego-codes:
Just chiming in, my application also requires this type of solution. It would be great if Carbon provided this out of the box. I will do the workaround that @asudoh suggested, but I would urge you to making this feature a higher priority 😅.

martynjames:
My application already has this kind of component, and we'd love to be able to switch to using carbon

aagonzales:
[...]

PR #1704, opened 2019-01-03
feat(DatePicker): add time picker

yasotnik:
Closes IBM/carbon-components-react#1140

Added the time picker functionality. 
Since the flatpickr support time picker, just the additional props were added to `DatePicker` to be able to use time picker.
This PR is directly connected to PR that solves styling issue after time picker was added to date picker. https://github.com/IBM/carbon-components/pull/1586

#### Changelog

**New**

- Added additional props to `DatePicker`:
      _enableTime
      time_24hr
      minTime
      maxTime_
- Added a story for single `DatePicker` with time picker and min/max time in 24hr format.

image


joshblack:
Hi @yasotnik! Thanks for taking the time to make this PR 😄 For Time Picker, I think we try and defer to the `TimePicker` component. For integrating alongside Date Picker, let me cc our design group to get their thoughts cc @IBM/carbon-designers 

[Story link](https://deploy-preview-1704--carbon-components-react.netlify.com/?selectedKind=DatePicker&selectedStory=single%20with%20calendar%20and%20time%20picker%20with%20min%2Fmax%20time&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) for design to view

asudoh:
Seems that this PR merely attempts to enable [Flatpickr's `enableTime` feature](https://flatpickr.js.org/examples/#datetime). IIUC @marijohannessen and the team didn't enable that because it goes out of our design scope. As @joshblack alluded, the discussion need to start from designers's perspective.

JSoet:
@asudoh @joshblack For our application we'd much prefer the integrated timepicker with the date picker. In issue #1140 @aagonzales said that this could be a good pattern and there's obviously a number of other people who feel the same who mentioned it there...

yasotnik:
@joshblack Thanks for the story link.
As @aagonzales mentioned in #1140 DatePicker with TimePicker should be able to display time along with the date, it works perfectly, only thing that you need to do is to add proper `dateFormat` you could check it in the [Story](https://deploy-preview-1704--carbon-components-react.netlify.com/?selectedKind=DatePicker&selectedStory=single%20with%20calendar%20and%20time%20picker%20with%20min%2Fmax%20time&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel). The only thing that you won't be able to see there is the time, because the `max-width` of `picker__input` is fixed. Here is a screenshot with disabled max-width.
image

designertyler:
From the visual:
It's important the style of the time selector follows the standalone[ time picker ](https://www.carbondesignsystem.com/components/date-picker/code) and the rest of the system. The mockups here look a little off since they're still using flatpickr styling.

For usability:
This seems specific to the type of use case where the user sets an alert and only needs to set a start time and no end time. That use case can be solved today with the separate date and time picker components. We try to avoid adding new components if their use case can reasonably be solved with existing components.

Adding another task in a dropdown like this starts to get cumbersome and why I'd recommend breaking out the date and time selection in two different steps and components. Ideally form fields would be used to enter one data point and the dropdown to make only one selection.  

The time field would benefit from a label to avoid making the user interpret the field's intent and help clarify it's for the time and not related to a date. We'd also want to reflect that somehow in the "Date Picker Label"

The addition of the up/down arrows to incrementally set the time is a nice addition that is missing form our component.

JSoet:
Thanks for the feedback @designertyler 

Regarding the style of the time selector do you have some specific feedback for what needs to be changed? The story link that Josh posted is missing the styling because the styling is in the separate PR for the carbon repo: https://github.com/IBM/carbon-components/pull/1586 But you can see it with the styling applied in the original picture that @yasotnik posted:
image
@joshblack I'm not sure if there's any way to create a story which includes the styling from the other PR as well?

Regarding the usability feedback, in our case we don't need to just select an open ended time but we need to select a date range including a time on each end. Unfortunately if we try to do that with the TimePicker in the DateRange component like below:
image

Then it doesn't properly display and the 2 date range fields get pushed together and the time picker gets pushed after them and only one of the time pickers gets displayed:
image
(without the time picker it looks like this:)
image

Anyway, I realize that's a separate bug (which we need to make an issue for, I don't think we've done that yet), but given that bug and the fact that flatpickr already supports an integrated time picker and we found that issue where multiple people were asking for it to be implemented we thought that would be a nicer and easier fix to just allow the selection of the time... 

I'm not sure I agree with you about only allowing the selection of 'one thing' per dropdown - in my opinion the selection of a time and day are a single datapoint/value/field, so even though it requires multiple clicks you are still only giving the value for a single datapoint (the datetime) and so it's a pretty clear user interaction in my opinion... similar to a multi-select field where you can select multiple things for a single field.

Anyway, let us know what the next steps are, if you think this has a chance of getting approved with just some styling changes or if you think it'd need a lot of work on the usability before allowing it.

designertyler:
@JSoet The best route might be to add this as to an add-on repo if this is needed ASAP. We can add this version of the component later in Carbon v10.

We're in the process of updating all our components to the latest IBM design language for Carbon v10. This change to v10 will happen soon and we wouldn't be able to update this component to the [the v10 styling](https://next.carbondesignsystem.com/components/date-picker/code) in a quick timeframe.

yasotnik:
@designertyler Well, you are right, we need it "asap please". I am ready to make changes of style to follow `carbon X` and `carbon 9` (actually any changes that the component needs). Would it be enough then to approve my PR? If so, I just need to know how it should look like at least approximately (or could I do it just by my own feel how it should look? And may be we can discuss it then) and also how I actually can test it because styles are stored in a separate github repo - `carbon-components`. Thanks!

joshblack:
@yasotnik if this is an immediate need for your product team, I would figure out a way to do this custom for the deadline. Unfortunately, the process for getting things officially into Carbon could take more time that one might have with a looming deadline. This is namely because of the breadth of consumers and use-cases we have to check for in order to fully offer some piece of functionality.

Typically, the process for including a new piece of functionality into a control is to reach out in #carbon-design-system on Slack, reach consensus on the design and behavior, create the issue on `carbon-components`, and then mirror to each implementation.

With respect to style development, what you can do is use `yarn link` to link the two projects together. You can see more details about this process [here](https://github.com/IBM/carbon-components-react/blob/master/docs/carbon-components.md).

subhak77:
Having a time picker inside data picker component is a nice Idea, why don't we consider Yaroslav.sotnyk's PR to enable/disable time, 24 hrs time format.
Am awaiting to use this component in our project where, we required date range along with time range.

PR #1711, opened 2019-01-07
chore(build): supporting experimental for deploy preview

### LINKED ISSUE #1708: Add experimental toggle to storybook
It would be awesome if we could enhance storybook to have a toggle for experimental similar to what Angular has already at angular.carbondesignsystem.com

This would be helpful for design when reviewing PRs.

asudoh:
This PR enables experimental build for deploy preview  ~~by introducing `netlify.toml`~~ by building stable/experimental into a single directory (updated, as Netlify's "context" feature does not for generating multiple deploy previews from single branch).

Closes #1708.

**Note**: Though `carbon-components-angular` has a theme switcher in deploy preview, this PR doesn't do that because `carbon-components-angular`'s theme switcher presumably works with styles only whereas `carbon-components-react`'s experimental build involves JavaScript build.

#### Changelog

**New**

- Yarn build scripts for deploy previews, generating both stable/experimental builds


emyarod:
this PR should close #1708 right?

storybook-static-preview/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Deploy preview</title>
+    <link rel="stylesheet" type="text/css" href="https://unpkg.com/carbon-components@latest/css/carbon-components.min.css">
+    <style type="text/css">
```
emyarod: tiny nitpick but should we center these buttons on the page?

asudoh: OK makes sense - Done.

PR #1717, opened 2019-01-08
fix(DatePicker): avoid rendering extra icons for experimental DatePicker

emyarod:
The vanilla experimental date picker is displaying the SVG icons as a background image. To match this approach, we can avoid using the `<Icon>` component in the React DatePicker

**Changed**

- conditionally render `<Icon>` based on feature flag state


asudoh:
Hi @emyarod date picker uses background image for range version, uses inline `<svg>` for single version. I can see it's the same between stable/experimental in `carbondesignsystem.com` site. Did you see it differently...? Thanks!

emyarod:
thanks for catching that @asudoh, it should be rendering in the single experimental version as well now

Issue #1706, opened 2019-01-07
Allow for an additional option for a label for the dropdown v2? 


kalyanixraut:
## Detailed description

DropdownV2 component does have an option to provide a label to the dropdown. It makes sense to have an optional label and helper text as all other form components have that option. 

I am using carbon-react-component of below version.
"carbon-components": "9.62.6",
    "carbon-components-react": "6.70.1",

## Steps to reproduce the issue

http://react.carbondesignsystem.com/?selectedKind=DropdownV2&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

> CodeSandbox
https://codesandbox.io/s/p5z81rov0

## Additional information

- Like the way text input can have a label and helper text so expectation is something similar for dropdownV2 also.
<img width="375" alt="image" src="https://user-images.githubusercontent.com/22093326/50776114-237d2100-12be-11e9-87f6-71b17cd0dfe3.png">

joshblack:
Hi @kalyanixraut! 👋 Thanks for taking the time to make this issue 😄 Definitely agreed that this should be a part of the component!

PR #1718, opened 2019-01-08
feat(DropdownV2): support title and helper text options

emyarod:
Closes IBM/carbon-components-react#1706

This PR adds optional props to `DropdownV2` for a title and helper text.

#### Changelog

**New**

- optional `titleText` and `helperText` props


src/components/DropdownV2/DropdownV2.js
```diff
@@ -169,5 +195,14 @@ export default class DropdownV2 extends React.Component {
         )}
       </Downshift>
     );
+    return title || helper ? (
+      <FormItem>
```
asudoh: While `<FormItem>` wrapping is something nice to consider in future, I'd be reluctant to change the behavior by `title`/`helper` existence. Wondering if we can use a fragment for now.

Issue #1720, opened 2019-01-09
Tag component can not display zero


yasotnik:
## Detailed description

> Describe in detail the issue you're having.

The `Tag` component works wrong when I am trying to display zero.

> Is this issue related to a specific component?

Yes, **Tag**

> What did you expect to happen? What happened instead? What would you like to see changed?
```
const zero = 0
<Tag type="experimental">{zero}</Tag>
```
Should display `0` but instead, it displays the type name of the `Tag` component. Probably inside component treats `{zero}` object as `false`.

> What browser are you working in?

**Chrome** Version 71.0.3578.98

> What version of the Carbon Design System are you using?

**carbon-components-react:** 6.71.0
**carbon-components:** 9.64.3
**react:** 16.7.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

A reduced case on the `codesandbox`: https://codesandbox.io/s/xjl73xmo0w

PR #1722, opened 2019-01-09
fix(Tag): remove boolean logic from rendering children

emyarod:
Closes IBM/carbon-components-react#1720

This PR removes some boolean logic which was preventing the `children` of the `<Tag>` component from rendering when it was a falsey value like `0`. It appears like initially the logic was implemented to always render a value, but since `props.children` is not a required prop, it should be fine to have nothing render if no children are passed into the `<Tag>`

alternatively, we can set a default value on `children` rather than displaying the `type` value by default, which may cause problems when users migrate from v9 to v10 too since the type values are changing

### DIFF
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -36,7 +36,7 @@ const Tag = ({ children, className, type, ...other }) => {
   const tagClasses = classNames(`${prefix}--tag`, tagClass, className);
   return (
     <span className={tagClasses} {...other}>
-      {children}
+      {children !== null && children !== undefined ? children : TYPES[type]}
     </span>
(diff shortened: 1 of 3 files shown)


src/components/Tag/Tag.js
```diff
@@ -36,7 +36,7 @@ const Tag = ({ children, className, type, ...other }) => {
   const tagClasses = classNames(`${prefix}--tag`, tagClass, className);
   return (
     <span className={tagClasses} {...other}>
-      {children || TYPES[type]}
```
asudoh: Seems that `TYPES[type]` no longer renders with this change. Any background...?

Issue #1647, opened 2018-12-12
Experimental Component: Progress Indicator


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1186

https://github.com/IBM/carbon-components/issues/1252

https://github.com/IBM/carbon-components/issues/1300

https://github.com/IBM/carbon-components/pull/1479/

depends on https://github.com/IBM/carbon-components/pull/1612

PR #1726, opened 2019-01-09
feat(ProgressIndicator): add experimental progress indicator

emyarod:
Closes IBM/carbon-components-react#1647

depends on https://github.com/IBM/carbon-components/pull/1612

This PR adds experimental progress indicator styles and features. In order to support the overflow tooltip functionality, the `<Tooltip>` component requires a style override since the existing CSS classes on the React `<Tooltip>` component conflict with the `<ProgressIndicator>` overflow tooltip's styles

specifically referring to these two lines:

https://github.com/IBM/carbon-components-react/pull/1726/files#diff-24f3b48f435eda51789be323bce053c5R98

and 

https://github.com/IBM/carbon-components-react/pull/1726/files#diff-7efca1118843036947e04d876136f7c6R421


src/components/Tooltip/Tooltip.js
```diff
@@ -418,6 +418,7 @@ export default class Tooltip extends Component {
               onBlur={evt => this.handleMouse(evt)}
               aria-haspopup="true"
               aria-expanded={open}
+              style={this.props.style}
```
asudoh: Can we do (probably additional) CSS classes rather than `style` attribute?

emyarod: the PR is open here: https://github.com/IBM/carbon-components/pull/1635

emyarod:
dependent on https://github.com/IBM/carbon-components/pull/1635

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -9,49 +9,113 @@ import PropTypes from 'prop-types';
...
+          showIcon={false}
+          triggerClassName={`${prefix}--progress-label`}
+          triggerText={label}
+          {...overflowTooltipProps}>
+          {overflowTooltipChildren}
+        </Tooltip>
+      ) : (
+        <p
+          className={`${prefix}--progress-label`}
+          aria-describedby={componentsX ? 'label-tooltip' : null}>
```
asudoh: We make sure we generate instance-speciic ID for ID refs like `aria-describedby`. Which element should this point to BTW?

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -82,6 +148,36 @@ ProgressStep.propTypes = {
...
+
+  /**
+   * Provide the content of a progress step tooltip
+   */
+  overflowTooltipChildren: PropTypes.node,
+
+  /**
+   * Provide the props that describe a progress step tooltip
+   */
+  overflowTooltipProps: PropTypes.object,
```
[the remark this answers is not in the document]
emyarod: refactored to use a `renderTooltip` prop

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -9,49 +9,98 @@ import PropTypes from 'prop-types';
...
+      </svg>
+    );
+  })();
 
   return (
     <li className={classes}>
       {currentSvg || completeSvg || incompleteSvg}
-      <p className={`${prefix}--progress-label`}>{label}</p>
+      {componentsX && renderTooltip ? (
+        renderTooltip()
```
asudoh: * Given `renderTooltip()` is a new feature, looks that we don't need to role it off for `v10`?
* Can we set default render code as the default prop?
* Can we support stateful here as well...?

Imagining something like below, wondering if it makes sense to everybody:

```diff
--- a/src/components/ProgressIndicator/ProgressIndicator.js
+++ b/src/components/ProgressIndicator/ProgressIndicator.js
@@ -17,6 +17,8 @@ import { componentsX } from '../../internal/FeatureFlags';
 
 const { prefix } = settings;
 
+const defaultRenderLabel = props => <p {...props} />;
+
 export const ProgressStep = ({ ...props }) => {
   const {
     label,
@@ -27,7 +29,7 @@ export const ProgressStep = ({ ...props }) => {
     invalid,
     secondaryLabel,
     disabled,
-    renderTooltip,
+    renderLabel: ProgressStepLabel,
   } = props;
 
   const classes = classnames({
@@ -91,11 +93,7 @@ export const ProgressStep = ({ ...props }) => {
   return (
     <li className={classes}>
       {currentSvg || completeSvg || incompleteSvg}
-      {componentsX && renderTooltip ? (
-        renderTooltip()
-      ) : (
-        <p className={`${prefix}--progress-label`}>{label}</p>
-      )}
+      <ProgressStepLabel className={`${prefix}--progress-label`}>{label}</ProgressStepLabel>
       {componentsX &&
       secondaryLabel !== null &&
       secondaryLabel !== undefined ? (
@@ -146,7 +144,7 @@ ProgressStep.propTypes = {
    * An optional parameter to allow for overflow content to be rendered in a
    * tooltip.
    */
-  renderTooltip: PropTypes.function,
+  renderLabel: PropTypes.function,
 
   /**
    * Provide the props that describe a progress step tooltip
@@ -164,6 +162,10 @@ ProgressStep.propTypes = {
   tooltipId: PropTypes.string,
 };
 
+ProgressStep.defaultProps = {
+  renderLabel: defaultRenderLabel,
+};
+
 export class ProgressIndicator extends Component {
   state = {}; 
```

emyarod: makes sense, thanks for the suggested changes @asudoh

asudoh:
Thank you for the update @emyarod! @joshblack Does @emyarod's latest code look good to you...? Thanks!

asudoh:
Merging as all review comments have been addressed. Anybody don't hesitate to bring up further concerns - Thanks!

Issue #1725, opened 2019-01-09
A11y: MultiSelect.filterable tab focus


yacineMTB:
## Detailed description

> Describe in detail the issue you're having.

When using the keyboard(tab), move focus to "All Types". It's dropdown list can't be operated by keyboard.
Pressing tab once, the drop down list can be operated by keyboard, but the focus can't be seen.

> Is this a feature request (new component, new icon), a bug, or a general issue?

general issue

> Is this issue related to a specific component?

Yes, MultiSelect.filterable

> What did you expect to happen? What happened instead? What would you like to see changed?

Dropdown list can be operated by keyboard and its focus can be seen.

> What browser are you working in?

`Google Chrome 70.0.3538.67`

> What version of the Carbon Design System are you using?

 `"carbon-components-react": "^6.60.1",`

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

ibm cloud - We need to close out AVT by early February to maintain our A11Y compliance status

## Steps to reproduce the issue

On staging, go to the `/status` microservice. Try tabbing through the filterable multi select components there.

## Additional information

![image](https://user-images.githubusercontent.com/10282244/50913313-1406f080-1402-11e9-82cf-c6d6ec367839.png)

![image](https://user-images.githubusercontent.com/10282244/50913310-11a49680-1402-11e9-88c6-647fe2da715d.png)

PR #1731, opened 2019-01-10
fix(filterablemulti-select): Fixed a11y tabbing issue on filterable multi select

yacineMTB:
Closes https://github.com/IBM/carbon-components-react/issues/1725

Fixed a11y tabbing issue on filterable multi select

#### Changelog

- Tabbing on filterable multi select works as expected
- added key down to expand options


src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -117,11 +117,14 @@ export default class FilterableMultiSelect extends React.Component {
...
     const { type } = changes;
     switch (type) {
-      case Downshift.stateChangeTypes.keyDownArrowDown:
       case Downshift.stateChangeTypes.keyDownArrowUp:
       case Downshift.stateChangeTypes.itemMouseEnter:
         this.setState({ highlightedIndex: changes.highlightedIndex });
         break;
+      case Downshift.stateChangeTypes.keyDownArrowDown:
+        this.setState({ highlightedIndex: changes.highlightedIndex });
+        this.setState({ isOpen: true });
```
joshblack: Could these be in the same `setState` call?

src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -227,6 +230,7 @@ export default class FilterableMultiSelect extends React.Component {
                 disabled={disabled}
                 invalid={invalid}
                 invalidText={invalidText}
+                tabIndex="-1"
```
joshblack: Why do we need the `-1` on the ListBox root?

yacineMTB: Good call

src/components/MultiSelect/FilterableMultiSelect.js
yacineMTB: Good call

yacineMTB:
Sorry I let this stale. Let me know if I added enough coverage.

yacineMTB:
Not too sure why the build is failing.  Locally: `PASS  src/components/ListBox/__tests__/ListBox-test.js`
FYI I'm not using yarn - don't want to jump through the hoops of getting it on RHEL

dakahn:
Do you think you could have the menu be activated on `enter` as well?

asudoh:
BTW the CI error seems to have come from:

```
$ prettier --list-different "**/*.{scss,css,js,md}"
src/components/ListBox/__tests__/ListBox-test.js
```

I guess the Prettier version @yacineMTB uses and one our `yarn.lock` points to are different.

yacineMTB:
Thanks @asudoh, I got it.

@dakahn: your comment prompted me to think a little bit more about what key presses I expect to manipulate the menu. I use my computer with my keyboard, as in I don't have a mouse, so I have some insight on how this menu should act

### 1: Initial view
On tabbing to the menu, focus is communicated to the user through changing the color of the selection box.
selected

#### Pressing enter
I expect the menu to open. It doesn't. User's that browse the web using their keyboard will use enter primarily to click the focused element. When I'm ready to finish a github comment, I will hit ctr+f to focus the comment button and hit enter. 

#### Pressing tab again: go to  2 

#### Pressing space: go to 3

### 2: Search box 
On pressing tab, we are on the search box. The focus is displayed to the user through the cursor blinking. 
search

#### key presses to fill text box
On entering some text, I expect the dialog to be forced open and auto filter. Here's what happens:
write-wrong

Here's what I expect to happen
writecorrect

### 3: Pressing space to open the menu

Here's what happens, as expected
space

#### Pressing space when menu is open

The problem is when pressing space again, the menu closes. This is not what I expect, as Item is highlighted. Pressing enter on the other hand completes the right action:

enterorspace
(^ to clarify; what should happen when enter or space are pressed when menu is open)

### Pressing escape when menu is open: go to 4

### 4: Pressing escape
When the menu is open, escape closes the menu.
escape

This works as expected.

### 5: Removing a selection
When focus is on the overall box, I expect spamming escape to remove all selections: aka clear the 1x
When focus is on the 1x, I expect space or enter to remove it which doesn't happen

escape2
escape

Let me know if you guys agree with these observations and I'll make the necessary changes for the multi-selects

dakahn:
This is so cool. Thank you so much for your thoughtful insight and hard work. I'm happy to have you continue onward and implement the proposed changes. 👍🏼

src/components/ListBox/ListBox.js
```diff
@@ -56,7 +57,7 @@ const ListBox = ({
         {...rest}
         role="listbox"
         aria-label={ariaLabel}
-        tabIndex="0"
+        tabIndex={innerTabIndex || 0}
```
joshblack: Just catching back up with this now, is there a specific reason why this is `innerTabIndex` as a prop versus supporting `tabIndex`?

yacineMTB: Didn't want to break other components that utilized listBox and expected the listBox being focusable, so provided a prop that could override that behaviour.

joshblack: @yacineMTB could you help me better understand the situation you're describing? I think I get what you're saying, but just want to confirm 😄

yacineMTB: Other components consume the listBox - and they may rely on the component being focusable for a11y purposes. 

In the case of the filterable multi select, we don't require the listbox to be focusable. I am led to believe this as the list box is explicitly set to be focusable. To avoid breaking other component's a11y, I added an override property that can change the default behaviour that other components might depend upon.

PR #1734, opened 2019-01-10
feat: add theme-provider example

vpicone:
I'm not sure if this is the right space for this; it might make a better blog post than an example.

I think there's a lot of potential in using carbon-elements and the context API to theme carbon react applications. The new context hook makes this much more appealing than the render props method so I think it could make for a great showcase of carbon's theming potential.

In this example, the ThemedButton component uses a ThemeContext to style itself. If somewhere up the tree a ThemeProvider gives a theme, it uses that. If not, it falls back to a default theme.


vpicone:
Totally! I’m hoping to make this example a little less trivial/hacky and actually use some theme values. Just wanted to get same feedback on if this was even the correct space first haha.

vpicone:
Great idea. Now that the themes are available through elements I think the example could be a little less contrived haha.

vpicone:
@asudoh I was trying to make it as library agnostic as possible when it comes to the actual theming of the components. Do you think including a css-in-js lib would be better than using the style tag to override the component styles?

asudoh:
Usage of css-in-js lib with Carbon has a caveat wrt de-duping shared styles, but using it as an example (as a choice among with non-React specific way of theming) sounds fine to me. Thank you for asking @vpicone!

joshblack:
Hey @vpicone! 👋 Were you still thinking of tackling this?

vpicone:
I kind of let the scope get out of hand haha. Working on paring it down to the essentials and readme.

vpicone:
Gonna put this one on ice and go a more simple direction

Issue #1736, opened 2019-01-11
Tile: Expandable height doesn't account for dynamic content below the fold


rob-richards:
## Detailed description

> Describe in detail the issue you're having.

The computed inline height of the Expandable Tile component does not account for dynamic content.

> Is this issue related to a specific component?

`<ExpandableTile>`

> What did you expect to happen? What happened instead? What would you like to see changed?

When the user clicks to expand the tile, maybe use maxHeight: none. When the user clicks to close the tile, revert to the computed height.

> What browser are you working in?

Chrome 70.0.3538.77

> What version of the Carbon Design System are you using?

Latest

Codesandbox
https://codesandbox.io/s/8ll6k38989

These may be helpful:
https://github.com/IBM/carbon-components-react/pull/679
https://github.com/IBM/carbon-components/pull/635

## Additional information
<img width="996" alt="screen shot 2019-01-09 at 11 10 08 am" src="https://user-images.githubusercontent.com/7707326/50912112-5844c180-13ff-11e9-8fca-8902ab47dbac.png">

emyarod:
if I'm not mistaken this has been resolved in #1737, but please feel free to reopen the ticket if this is not the case

PR #1737, opened 2019-01-11
fix(Tile): Expandable height now accounts for dynamic content below the fold

rob-richards:
Closes IBM/carbon-components-react https://github.com/IBM/carbon-components-react/issues/1736

#### Changelog

**New**
We are now setting a `max-height: none` on the expandable tile when the content below the fold is expanded.


src/components/Tile/Tile.js
```diff
@@ -442,7 +442,9 @@ export class ExpandableTile extends Component {
     );
 
     const tileStyle = {
-      maxHeight: this.state.tileMaxHeight + this.state.tilePadding,
+      maxHeight: this.state.expanded
+        ? `none`
```
emyarod: tiny nitpick: would it be better to set the value to `null` rather than `'none'` to remove the inline style rule entirely when the tile is collapsed?

rob-richards: @emyarod Thanks for the review! Sure, I just pushed an update.

PR #1760, opened 2019-01-16
chore: add copyright headers

emyarod:
Closes IBM/carbon-components-react#1757

This PR adds copyright headers to all source files


emyarod:
forgot to leave a note, but this should be updated now @joshblack

PR #1766, opened 2019-01-17
feat(inlineloading): added new prop alignTextLeft

loganmccaul:
Closes IBM/carbon-components#1659

Probably should've had the issue be in the react repo, my bad!

This adds a prop to the InlineLoading component to render the text before the loading spinner for contextual a better visual pattern when the contextual loading area is right aligned on the page.

### DIFF
--- a/src/components/InlineLoading/InlineLoading-story.js
+++ b/src/components/InlineLoading/InlineLoading-story.js
@@ -25,6 +25,7 @@ const props = () => ({
     1500
   ),
   onSuccess: action('onSuccess'),
+  alignTextLeft: boolean('Align text to the left of the spinner', false),
 });
 
 storiesOf('InlineLoading', module)
@@ -105,7 +106,7 @@ storiesOf('InlineLoading', module)
       info: {
         text: `
             This is a full example of how to levarage the <InlineLoading /> component to create a nice user experience when submitting a form.
-    
+
             For the full source code of this example, check out the 'story' panel below.
           `,
       },
--- a/src/components/InlineLoading/InlineLoading-test.js
+++ b/src/components/InlineLoading/InlineLoading-test.js
@@ -59,7 +59,12 @@ describe('Loading', () => {
     });
 
     it('should render the loading spinner first by default', () => {
-      expect(wrapper.find('.bx--inline-loading').childAt(0).hasClass('bx--inline-loading__animation')).toEqual(true)
+      expect(
+        wrapper
+          .find('.bx--inline-loading')
+          .childAt(0)
+          .hasClass('bx--inline-loading__animation')
+      ).toEqual(true);
     });
   });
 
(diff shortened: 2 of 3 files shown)


loganmccaul:
@asudoh Oh I like that! Passing in children will open the door for the whole thing to be more flexible. In that case would we need the InlineLoading wrapper? It's a bit counter intuitive to the existing component, but if we expanded the Loading component to have the check mark behavior then any additional content can be added by the product teams using other components.

Essentially I'm not even sure if this component is necessary at all 😄

asudoh:
Good point @loganmccaul! From looking at the code, the loading/complete state seems to affect nothing but the spinner/complete. That said, if we separate out `<InlineLoadingState>`, original `<InlineLoading>` will be merely a wrapper of an element with a specific CSS class. I'm open for just letting application do it.

loganmccaul:
@asudoh I like that. I think it would really expand on the small loader's use cases and the context where it can be used. My biggest concern with that is make any of those changes without needing a major version bump. What do you think the best approach would be without any BREAKING changes?

Maybe @joshblack might have some thoughts too?

asudoh:
@loganmccaul Thank you for your further thoughts! Just wanted to make sure we are in-line on wordings, etc. From reading your comments, it sounds that 1. is closer to your idea (as you are suggesting that original `<InlineLoading>` will be useless with the change we are thinking of), and you are suggesting `<Loading>` may cover `<InlineLoadingState>` features. Is it correct?

The only concern with changing `<Loading>` for that is added feature there. What I'm environing here is layered architecture - Higher-level component (`<InlineLoadingState>`) composed by lower-level component (`<Loading>`). To make what I'm environing clear, `<InlineLoadingState>` would be mostly identical to original `<InlineLoading>`, with the following changes:

```diff
--- a/src/components/InlineLoading/InlineLoading.js
+++ b/src/components/InlineLoadingState/InlineLoadingState.js
@@ -26,11 +26,6 @@ export default class InlineLoading extends React.Component {
     success: PropTypes.bool,
 
     /**
-     * Specify the description for the inline loading text
-     */
-    description: PropTypes.string,
-
-    /**
      * Provide an optional handler to be inovked when <InlineLoading> is
      * successful
      */
@@ -83,17 +78,10 @@ export default class InlineLoading extends React.Component {
       return <Loading small withOverlay={false} />;
     };
 
-    const loadingText = (
-      <p className={`${prefix}--inline-loading__text`}>{description}</p>
-    );
-
     return (
-      <div className={loadingClasses} {...other}>
       <div className={`${prefix}--inline-loading__animation`}>
         {getLoading()}
       </div>
-        {description && loadingText}
-      </div>
     );
   }
 }
```

joshblack:
Hey @loganmccaul! 👋 Are you still interested in getting this merged in?

joshblack:
Hey @loganmccaul! Going to close this for now due to inactivity, let me know if you'd like to keep working on this!

Issue #1721, opened 2019-01-09
Experimental Component: OverflowMenu


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1243

https://github.com/IBM/carbon-components/issues/1292

https://github.com/IBM/carbon-components/pull/1395/

PR #1767, opened 2019-01-18
feat(OverflowMenu): add experimental OverflowMenu

emyarod:
Closes IBM/carbon-components-react#1721

This PR adds the experimental `<OverflowMenu />` component changes.

The non-floating version of this component was incorrectly positioned, fixed by https://github.com/IBM/carbon-components/pull/1653 since only the React component depends on the positioning style rules.

The floating version of this component is now dynamically positioned according to the experimental spec, related ticket https://github.com/IBM/carbon-components/pull/1482


emyarod:
blocked by https://github.com/IBM/carbon-components/pull/1662

aledavila:
clicking on the icon is the only way to open the overflow menu. If you click anywhere in the box it won't open.

emyarod:
@aledavila looks like this is caused by the size difference between the new and old icons. just pushed an update to address that

aledavila:
Thanks @emyarod

Issue #1191, opened 2018-08-14
Stateless custom text input doesn't allow to use refs


IOIO72:
## Detailed description

The `TextInput` component is stateless and can't be used in combination with refs.

## Additional information

https://reactjs.org/docs/refs-and-the-dom.html

joshblack:
Hi @IOIO72! 👋 Thanks for taking the time to make this issue.

Could you elaborate a bit more on what you're expecting here? Is the request to convert the form elements to be class-based in order to use `ref`s?

IOIO72:
Hi Josh, yes. Sorry for being too brief.
It would be nice, if the input fields world accept refs and therefore they should be converted to classes.

asudoh:
Hi @IOIO72 wondering this is similar to what you want: https://github.com/IBM/carbon-components-react/blob/be9b7d0/src/components/Button/Button.js#L51

joshblack:
We could probably wrap any of our stateless form components with `React.forwardRef`, I bet, to satisfy this use-case. What do you think @asudoh?

asudoh:
Makes total sense for stateless ones @joshblack! For stateful we may want to do with next major release not to suddenly change what `ref` in our component means.

silesky:
PR here https://github.com/IBM/carbon-components-react/pull/1770

PR #1770, opened 2019-01-18
feat(TextInput): allow TextInput to handle refs like a native input e…

silesky:
Closes IBM/carbon-components-react#1191

added ability to forward refs

### DIFF
--- a/src/components/Checkbox/Checkbox-test.js
+++ b/src/components/Checkbox/Checkbox-test.js
@@ -106,6 +106,35 @@ describe('Checkbox', () => {
   });
 });
 
+describe('refs', () => {
+  it('should accept refs', () => {
+    class MyComponent extends React.Component {
+      constructor(props) {
+        super(props);
+        this.myRef = React.createRef();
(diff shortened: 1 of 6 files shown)


silesky:
Hey @asudoh, I would be happy to contribute more, but that's a significant
amount of code changes (mostly, the tests).

Since this issue is really about a single input (arguably, the most
important), I would rather see this merged in and take out enother issur/PR
for the rest of the form elements.

Thoughts?

asudoh:
@silesky I saw your immediate need for the issue to be fixed, OTOH I saw myself reluctant to introduce this only to `<TextInput>` for the sake of immediate need, because confusions arose from feature, etc. inconsistency among similar components has gradually become bigger in recent months. That said, it'd be great if you can address `<Checkbox>` and `<Select>` as the start. Thanks!

silesky:
@asudoh Select and Checkbox added

emyarod:
we can create a separate issue for the remaining components not covered in this PR

Issue #1114, opened 2018-07-20
RadioTile Component causing accessibility failures.


AsthaJain1:
## Detailed description

> Describe in detail the issue you're having.
I am using Selectable Tile component. 
http://react.carbondesignsystem.com/?selectedKind=Tile&selectedStory=Selectable&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug - causing accessibility failures at my end.
> Is this issue related to a specific component?

http://react.carbondesignsystem.com/?selectedKind=Tile&selectedStory=Selectable&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

> What did you expect to happen? What happened instead? What would you like to see changed?
Description of accessibility failure - 
https://aat.mybluemix.net/token/97b22e9c-25bd-48ea-be04-c77a42c72b11/a313b75d-5599-420e-9695-15c1599da92a/doc/w3/help/en-US/idhi_accessibility_check_g1062.html

The input has a css of `display: none` causing this failure.

> What browser are you working in?
Chrome
> What version of the Carbon Design System are you using?
    "carbon-components": "^8.18.18",
    "carbon-components-react": "^5.52.4",
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cloud

ZachStowellIBM:
@alisonjoseph Our project is still seeing this a11y issue in chrome, using "carbon-components": "^9.18.0" and  "carbon-components-react": "^6.62.2".

I've found a hacky workaround for this initial a11y error, however it leads to more issues that would require some major changes to the RadioTile component, as well as possibly some changes to the TileGroup component (and possibly changes in the `carbon-components` library).

Looking at https://xo33l4xqxp.codesandbox.io/ ([editable](https://codesandbox.io/s/xo33l4xqxp)), I'm seeing the same a11y failure as above.

[...]

PR #1774, opened 2019-01-21
fix(Tile): reorder component markup for a11y

emyarod:
Closes IBM/carbon-components-react#1114

This PR modifies `<SelectableTile>` and `<TileGroup>` component markup to address a11y failures

#### Changelog

**Changed**

- arrange `<SelectableTile>` input field before its corresponding `<label>`
- render `<TileGroup>` as `fieldset` element rather than a `div`


ZachStowellIBM:
In case you're not able to run DAP due to token issues, I've attached a screenshot of the DAP a11y results for the deploy preview for the selectable tile.

screen shot 2019-01-22 at 8 50 46 am

You can see there are no more errors related to the tiles (the two errors shown in the screenshot are due to storybook).

emyarod:
thank you @ZachStowellIBM !

gdelory:
Switching div to fieldset in TileGroup makes impossible to have the TileGroup as flex container with row direction, which was really convenient :(

gdelory:
And I can't find a work around to have row directed items, here what I used to have which is now broken
https://codesandbox.io/s/5347p565jn

gdelory:
Using inline blocks does the trick, sorry for bothering :)
https://codesandbox.io/s/q98lw3yjy9

emyarod:
it seems `fieldset` cannot be styled as a flex container only in Chrome

glad you figured out a workaround @gdelory!

valZho:
@emyarod While we can switch from using flex-box to just inline-block elements in the short term, changing a div into a fieldset is probably not the best idea. Instead, the existing div should have been wrapped in a fieldset. Perhaps this can still be implemented?

emyarod:
@valZho yes feel free to open a new issue about this proposal

PR #1608, opened 2018-12-04
Dropdown components as items

ZacharyStair:
Reopens https://github.com/IBM/carbon-components-react/pull/1505

Closes https://github.com/IBM/carbon-components-react/issues/1495

@asudoh I used your suggestion to check itemToElement. Also added a story/unit test to sanity check the 'items as strings' case

#### Changelog

**Changed**

- DropdownV2 now accepts an itemToElement prop and prefers that over itemToString

PR #1781, opened 2019-01-23
feat(DropdownV2): allows consumers to specify render prop for dropdown items

ZacharyStair:
this is a v5 version of https://github.com/IBM/carbon-components-react/pull/1608, the functionality doesn't depend on anything v6-specific so it would be nice to port back.

### DIFF
--- a/src/components/DropdownV2/DropdownV2.js
+++ b/src/components/DropdownV2/DropdownV2.js
@@ -42,6 +42,11 @@ export default class DropdownV2 extends React.Component {
     itemToString: PropTypes.func,
 
     /**
+     * Optional function to render items as custom components instead of strings.
+     */
+    itemToElement: PropTypes.func,
+
+    /**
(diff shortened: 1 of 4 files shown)


ZacharyStair:
@asudoh @joshblack could you take a look?

src/components/DropdownV2/DropdownV2.js
```diff
@@ -41,6 +41,11 @@ export default class DropdownV2 extends React.Component {
      */
     itemToString: PropTypes.func,
 
+    /**
+     * Optional function to render items as custom components instead of strings.
+     */
+    itemToElement: PropTypes.func,
```
joshblack: As a convention, should we have functions called for rendering be called `renderX`? For this one, could be `renderItem`, potentially?

joshblack:
@ZacharyStair I think it's because we switched CI providers from Travis to Circle which is causing some issues. Sorry about that! 😄

PR #1784, opened 2019-01-23
fix(tab-content): add prefix to content classname

### LINKED ISSUE #1790: Tab: content class name without prefix
## Detailed description

The `Tab` component has styles for a class name selector that is `.tab-content` I expect all class names in the design system to be prefixed.

Yes, `Tab`

> What did you expect to happen? What happened instead? What would you like to see changed?'

I expect the all tab related class names to include the set prefix (ie `bx--`)

9-something

Security

diego-codes:
Closes #1790 

Adds missing class name prefix to TabContent component

#### Changelog

**Changed**

- Adds missing class name prefix to TabContent component


diego-codes:
Hi Akira, that's a good question. Because there were no styles applied to that class name, I didn't notice any issues with it.

asudoh:
OK sounds that we can remove the class name - Thanks @diego-codes for your insights!

diego-codes:
@joshblack you got it!

PR #1787, opened 2019-01-24
fix(components): avoid being shown as ForwardRef

### LINKED ISSUE #1786: Checkbox: Shown as ForwardRef in React DevTools
## Detailed description

React DevTools shows our `<Checkbox>` as `<ForwardRef>`:

image

`<Checkbox>`, `<Select>` and `<TextInput>`

React DevTools shows our `<Checkbox>` as it is.

`6.79.0`

### LINKED ISSUE #1789: TextInput is shown as ForwardRef in React dev tool
image

If this is permanent please close and let me know. Much prefer TextInput though ;)

See comment from me in: https://github.com/IBM/carbon-components-react/issues/1786

asudoh:
Fixes #1786.
Fixes #1789.

#### Changelog

**New**

- Added `displayName` to components with `React.forwardRef()`.


src/components/Checkbox/Checkbox.js
```diff
@@ -12,7 +12,7 @@ import { settings } from 'carbon-components';
 
 const { prefix } = settings;
 
-const Checkbox = ({
+const CheckboxImpl = ({
```
asudoh: Though merging `Checkbox` and `CheckboxImpl` prevents `<Checkbox><CheckboxImpl ... /><Checkbox>` being shown in React DevTools, such technique cannot be applied to stateful components, and thus I kept it intact.

rmullan:
Thank you :)

src/components/Checkbox/Checkbox.js
emyarod: would this be a reason not to convert `<Checkbox>` to a stateful component? I saw that you suggested that as a possible alternative solution to set indeterminate prop when a ref is also passed into `<Checkbox>`

joshblack:
Is this so that `Checkbox` retains its displayName? Would we be okay with `ForwardRef(Checkbox)` as well?

src/components/Checkbox/Checkbox.js
asudoh: @emyarod Agreed that `<...Impl>` showing up in React DevTools (and other env) is noisy, but seems that we cannot avoid that e.g. for `<NumberInput>`, which is stateful. I think this shouldn't stop us from moving `<Checkbox>` to stateful if we think we need it.

asudoh:
@joshblack `ForwardRef(Checkbox)` is fine to me, though I think just `Checkbox` being shown is better.

src/components/Checkbox/Checkbox.js
asudoh: Talked to @emyarod offline, he seems fine with `ForwardRef(Checkbox)` though he seems to prefer the other. I went ahead with `ForwardRef(Checkbox)` based on that.

silesky:
@asudoh This works fine, but might be worth refactoring into a HoC to avoid the boilerplate, like in this stackoverflow

https://stackoverflow.com/a/53875195/6216756

src/components/Checkbox/Checkbox.js
```diff
@@ -126,6 +126,8 @@ Checkbox.defaultProps = {
   indeterminate: false,
 };
 
-export default React.forwardRef((props, ref) => (
-  <Checkbox {...props} forwardRef={ref} />
-));
+export default React.forwardRef(
+  Object.assign((props, ref) => <Checkbox {...props} forwardRef={ref} />, {
+    displayName: 'Checkbox',
```
silesky: this displayName may be problematic, because from my understanding this means component tree will end up looking like
```
<Checkbox>
  <Checkbox />
<Checkbox>
```
this could break a lot of unit tests that do something like Wrapper.find('Checkbox') because suddenly there are two checkboxes, instead of one

it's better to have something like
<withForwardedRef(Checkbox)>
 <Checkbox>
</withForwardedRef(Checkbox)>

joshblack:
Oh duh, it's the function name passed to `React.forwardRef` 🤦‍♂️ 

Would it be helpful to separate out the function passed to it and set the `displayName` manually then?

```jsx
const Checkbox = () => {
  // ...
};

function forwardRef(props, ref) {
  return <Checkbox {...props} innerRef={ref} />;
}

forwardRef.displayName = 'Checkbox';

export default React.forwardRef(forwardRef);
```

src/components/Checkbox/Checkbox.js
```diff
@@ -126,6 +126,8 @@ Checkbox.defaultProps = {
...
 };
 
-export default React.forwardRef((props, ref) => (
-  <Checkbox {...props} forwardRef={ref} />
-));
+const forwardRef = (props, ref) => <Checkbox {...props} forwardRef={ref} />;
+
+forwardRef.displayName = 'Checkbox';
+
+export default forwardRef;
```
joshblack: Might need to be:

```suggestion
export default React.forwardRef(forwardRef);
```

?

asudoh: Oops you are right - Made changes as such.

asudoh:
@joshblack Did you have any further comments/thoughts...? Thanks!

asudoh:
Merging in as all comments from the reviewers have been addressed - Anybody don't hesitate to speak up otherwise - Thanks! (Note to @silesky - I see reducing the boilerplate definitely is an interesting idea and we'll take note)

PR #1788, opened 2019-01-24
fix(Checkbox): ensure indeterminate DOM property is set

### LINKED ISSUE #1785: Checkbox: indeterminate prop on DOM node is not set if ref is given
## Detailed description

`<Checkbox>` sets `indeterminate` property to its `<input>` when the corresponding prop is set. Though this is not a critical feature AFAIK, this ensures vanilla parity. With recent change with `React.forwardRef()`, it doesn't happen if `ref` is given.

To make long story short, below test fails, though it used to pass before `React.forwardRef()` was introduced:

```javascript
class MyComponent extends React.Component {
  constructor(props) {
    super(props);
    this.myRef = React.createRef();
  }
  render() {
    return (
      <Checkbox indeterminate ref={this.myRef} />
    );
  }
}
const wrapper = mount(<MyComponent />);
expect(wrapper.find('input').getDOMNode().indeterminate).toBe(true);
```

`<Checkbox>`

`6.79.0`

asudoh:
Fixes #1785.

#### Changelog

**Changed**

- Fix to `ref` callback logic so `indeterminate` property is set to `<input>` even if `ref` prop is given.


src/components/Checkbox/Checkbox.js
```diff
@@ -45,15 +44,16 @@ const Checkbox = ({
...
-            if (input) {
-              input.indeterminate = indeterminate;
-            }
-          })
-        }
+        ref={el => {
+          if (el) {
+            el.indeterminate = indeterminate;
+          }
+          if (typeof ref === 'function') {
```
asudoh: The logic may be a bit hacky, an alternative here is turning `<Checkbox>` to stateful.

lovemecomputer: look good, i prefer a little hacky-looking code, over making a stateful component if not otherwise necessary! it might be cool to see a little comment explaining the hack in these occurrences tho 😇 (if the hack is sensible enough to explain verbally at least)

asudoh: Thanks @lovemecomputer for your comment - Agreed with you that we need some explanation of this approach - @alisonjoseph?

PR #1792, opened 2019-01-24
fix(TableExpandRow): pass headers value to fix a11y violation

bakerTX:
We were experiencing the following violation in a11y testing: 
screen shot 2019-01-24 at 4 11 13 pm
The `<td>` element containing the expand button was missing programmatic information linking it to a corresponding `<th>` value in the table head. See this page for more info:

https://aat.w3ibm.mybluemix.net/token/a80a8d09-7f5a-480a-adcc-7e22f77482b0/c19cc7e4-e2e6-4475-89d7-533c2bc34979/archives/2018OctDeploy//doc/w3/help/en-US/idhi_accessibility_check_g1059.html

#### Changelog

A prop *expandHeader* in TableExpandRow.js allowing the user to pass a matching id for the corresponding `<th>` (It seems that the DataTable requires the user to write the `<th>` elements themselves via children of TableHead.js - Thus, the end-user will need to write a matching id in their table head in order to comply with a11y standards) The expandHeader prop defaults to "expand".

Updated snapshot test for TableExpandRow.js


bakerTX:
Hey @joshblack! Sure, take a look at this screenshot.
screen shot 2019-01-29 at 10 18 27 am
(hovering over the header cell, selected the corresponding table cell)

The a11y violation is thrown because a table cell needs to be programmatically linked to its corresponding table header cell. In this case, the table cell containing the expand button isn't linked to the table header cell. (This header cell is generally empty, rendered only to keep proper alignment along the columns)

Fortunately we can add an id onto that presentational header cell since Carbon allows us to render the header cells however we like. However, the table cell containing the expand button is explicitly rendered through the Carbon component (TableExpandRow.js specifically) That table cell needs to have a `headers` attribute that matches the id of the presentational header cell. My PR adds an optional prop onto that table cell, so that the end-user can link it their header cell id appropriately.

This approach is a good first step, but it relies on the user to know about this violation and implement the structure themselves. They need to 
1) render the first header cell empty to match the column containing expand buttons
2) attach an id onto that first header cell
3) pass a prop to `<TableExpandRow>` matching the id above.

Ideally the Carbon component would add these attributes internally. But since the current usage of DataTable relies on the user to render table headers, and then the expand cell is rendered internally, we need to have the ability to link their attributes ourselves.

src/components/DataTable/TableExpandRow.js
```diff
@@ -79,6 +81,15 @@ TableExpandRow.propTypes = {
    * The description of the chevron right icon, to be put in its SVG `<title>` element.
    */
   expandIconDescription: PropTypes.string,
+
+  /**
+   * The id of the matching th node in the table head. Solves a11y DAP violation to match these appropriately.
```
dakahn: Since we don't have access to this DAP test or the referenced violation this should be changed to reference [1.3.1](https://www.ibm.com/able/guidelines/ci162/info_and_relationships.html).

bakerTX: Added references into the comment here

asudoh:
@joshblack Did you have any further comments...? Thanks!

PR #1793, opened 2019-01-24
feat(MultiSelect): add open prop

McBrosa:
**Summary**: This adds the ability to pass in the`isOpen` prop to have the menu open/closed by default. This was specifically added so that the menu could default to open.

**Additional Info**:

multiselect-open

#### Changelog

**Changed**

- MultiSelect: added isOpen prop for initializing the component with an open/closed menu


asudoh:
@McBrosa Thank you for bringing up additional thoughts - Just to make sure we are in sync, what I brought up earlier (prop -> state sync) can be found by searching for `getDerivedStateFromProps()` in `/path/to/carbon-components-react/src/components/`.

McBrosa:
I am not familiar with Downshift and trying to understand what is going on with this unit test. Any insight?

screen shot 2019-01-25 at 11 18 44 am

McBrosa:
Do we want to conform to the Downshift `isOpen` or the carbon-react `open` which is used for Modal, Tooltip, OverflowMenu, AccordionItem, and etc?

asudoh:
@joshblack Any thoughts on @McBrosa's last comment? Thanks!

joshblack:
Thanks for the bump @asudoh 😄 

So sorry for the delay, @McBrosa. Great point, we're hoping to move towards prop naming conventions that are roughly:

- If we are setting a known DOM attribute, match it in the prop name (like `focusable`)
- Otherwise, props are expressed as:
  - `is_` naming for initial state, for example `isOpen`
  - `should_` naming for specifying behavior, for example `shouldCloseAfterClick`

As you noted, this isn't necessarily the case for all components but we should try and stick to these for pieces that are being added as the old components will be updated in time to match this 👍 Let me double-check with the team though on this, will send them this PR just to get their thoughts and will respond back tomorrow.

joshblack:
Following up on this, let's stick with `open` so that it's in line with the other components. If we end up updating them at some point, then regardless we'd do them all at once 😄 Better to have consistency in this case, I think 👍

asudoh:
Thanks @joshblack for the update! It appears that the prop name was your only review comment - Is it correct?

McBrosa:
@joshblack Thank you for making the chages 😄. I was swamped with work and was on vacation this past week.

Issue #1801, opened 2019-01-28
MultiSelect menu icon tooltip not translated


lrdavis2:
## Detailed description
Hover tooltip `Open menu` for the MultiSelect menu icon is not translated.
![image](https://user-images.githubusercontent.com/19335667/51861009-22fa0800-2309-11e9-8b60-779829e3880a.png)

> Is this issue related to a specific component?
MulitSelect  

> What did you expect to happen? What happened instead? What would you like to see changed?
tooltip should be translated  

> What browser are you working in?
Chrome  

> What version of the Carbon Design System are you using?
`6.45.2`  

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cloud Availability Monitoring  

## Steps to reproduce the issue
1. change browser preferences to use non-English language
2. hover over the menu icon for the drop down 
3. notice the tooltip `Open menu` is in English

> Please create a reduced test case in CodeSandbox
> https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox

## Additional information
Looks like there is the ability to translate with a function 
```
ListBoxMenuIcon.defaultProps = {
 translateWithId: function translateWithId(id) {
   return defaultTranslations[id];
 }
};
```
but doesn't look like MultiSelect will forward a property like that...

PR #1802, opened 2019-01-28
feat(ListBoxMenuIcon): support translatable SVG icon title

emyarod:
Closes IBM/carbon-components-react#1801

This PR will add support for translatable SVG titles in `<ListBox />`

#### Changelog

**New**

- `listBoxMenuIconTranslationIds` prop, which takes an object with `close.menu` and `open.menu` properties

**Changed**

- return SVG `<title>` if `description` prop is truthy for experimental `ListBoxMenuIcon`

### DIFF
--- a/src/components/ListBox/ListBoxMenuIcon.js
+++ b/src/components/ListBox/ListBoxMenuIcon.js
@@ -40,7 +40,9 @@ const ListBoxMenuIcon = ({ isOpen, translateWithId: t }) => {
   return (
     <div className={className}>
       {componentsX ? (
-        <ChevronDown16 name="chevron--down" />
+        <ChevronDown16 name="chevron--down" aria-label={description}>
+          <title>{description}</title>
+        </ChevronDown16>
       ) : (
         <Icon
           icon={iconCaretDown}
--- a/src/components/MultiSelect/MultiSelect.js
+++ b/src/components/MultiSelect/MultiSelect.js
@@ -93,11 +93,9 @@ export default class MultiSelect extends React.Component {
     invalidText: PropTypes.string,
 
     /**
-     * Object containing localized alt text for the listbox menu icon
-     * listBoxMenuIconTranslationIds[close.menu] - alt text for closing listbox
-     * listBoxMenuIconTranslationIds[open.menu] - alt text for opening listbox
+     * Callback function for translating ListBoxMenuIcon SVG title
      */
-    listBoxMenuIconTranslationIds: PropTypes.object,
+    translateWithId: PropTypes.function,
   };
 
   static defaultProps = {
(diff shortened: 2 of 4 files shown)


src/components/MultiSelect/MultiSelect.js
```diff
@@ -103,6 +110,10 @@ export default class MultiSelect extends React.Component {
     type: 'default',
     light: false,
     title: false,
+    listBoxMenuIconTranslationIds: {
```
asudoh: Wondering if we can make it a callback style as some other components do.

emyarod: makes sense, refactored to take a callback function rather than a config object

src/components/MultiSelect/MultiSelect.js
```diff
@@ -103,6 +108,7 @@ export default class MultiSelect extends React.Component {
     type: 'default',
     light: false,
     title: false,
+    translateWithId: () => {},
```
asudoh: AFAIK default `translateWithId` in this codebase returns pseudo translation. For `<MultiSelect>`, letting `<ListBox.MenuIcon>` do that seems enough.

src/components/ListBox/ListBoxMenuIcon.js
```diff
@@ -40,7 +40,9 @@ const ListBoxMenuIcon = ({ isOpen, translateWithId: t }) => {
   return (
     <div className={className}>
       {componentsX ? (
-        <ChevronDown16 name="chevron--down" />
+        <ChevronDown16 name="chevron--down" aria-label={description}>
+          <title>{description}</title>
```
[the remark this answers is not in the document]
emyarod: I'm a bit confused, the description is given here https://github.com/IBM/carbon-components-react/blob/9204ab3bae7876a9300e401475b05f9af30c4da4/src/components/ListBox/ListBoxMenuIcon.js#L39

am I misunderstanding your question?

src/components/MultiSelect/MultiSelect.js
emyarod: don't we need to expose this on `<MultiSelect>` to pass through to `<ListBox.MenuIcon>`?

src/components/ListBox/ListBoxMenuIcon.js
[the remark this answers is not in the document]
asudoh: My apologies, my oversight and you are right @emyarod!

src/components/MultiSelect/MultiSelect.js
asudoh: If I understand your question correctly - What I mean here is we don't need `defaultProps` here - Passing along `undefined would be fine.

emyarod: got it @asudoh, removed the default function value

asudoh:
@emyarod Not sure why CircleCI is stall, but I believe the snapshot needs update. Could you do that...? Thanks!

asudoh:
Merged as the CircleCI job itself seems to have passed - @joshblack Great if you have any insights on GH integration. Thanks!

PR #1810, opened 2019-01-31
Update 'InnerClickListener'

Calvin-Cheung:
Update InnerClickListener

#### Changelog

**Changed**

- check if event target exists in DOM
- if it does not exist, should not trigger onClickOutside

### DIFF
--- a/src/internal/InnerClickListener.js
+++ b/src/internal/InnerClickListener.js
@@ -38,10 +38,12 @@ export default class InnerClickListener extends React.Component {
   }
 
   handleDocumentClick(event) {
-    if (this.element) {
-      if (this.element.contains && !this.element.contains(event.target)) {
-        this.props.onClickOutside(event);
-      }
+    // Ensure that the target exists in the DOM before checking the element
+    // documentElement necessary for IE
+    if (document.documentElement.contains(event.target) && this.element) {
+        if (this.element.contains && !this.element.contains(event.target)) {
+            this.props.onClickOutside(event);
+        }
     }
   }
 


src/internal/InnerClickListener.js
```diff
@@ -38,10 +38,12 @@ export default class InnerClickListener extends React.Component {
   }
 
   handleDocumentClick(event) {
-    if (this.element) {
-      if (this.element.contains && !this.element.contains(event.target)) {
-        this.props.onClickOutside(event);
-      }
+    // Ensure that the target exists in the DOM before checking the element
+    // documentElement necessary for IE
+    if (document.documentElement.contains(event.target) && this.element) {
```
emyarod: I believe we avoid usage of globals like `document` or `window` to prevent issues with scoping cc @asudoh

```suggestion
    if (this.element.ownerDocument.contains(event.target) && this.element) {
```

emyarod:
hi @Calvin-Cheung thanks for submitting this PR! It looks like you will also need to run `prettier` to get your changes to pass CI

src/internal/InnerClickListener.js
```diff
@@ -40,7 +40,10 @@ export default class InnerClickListener extends React.Component {
   handleDocumentClick(event) {
     // Ensure that the target exists in the DOM before checking the element
     // documentElement necessary for IE
-    if (this.element.ownerDocument.documentElement.contains(event.target) && this.element) {
+    if (
+      this.element.ownerDocument.documentElement.contains(event.target) &&
+      this.element
```
Calvin-Cheung: Going to change ordering of conditional to make more sense

asudoh:
Hi @Calvin-Cheung 👋 thank you for the contribution! My glance at the code indicates that the code can be simpler (e.g. just checking if React node's ancestor `<html>` contains the event target). To see that, could you quickly create a reduced test case? Thanks!

Calvin-Cheung:
Hi @asudoh, thanks for the suggestion. What would you recommend at this point?

asudoh:
@Calvin-Cheung Your latest code is close to my suggestion, however, as I stated before, we need a reduced case (or with some detailed steps) to see if we should move forward with this. Thanks!

PR #1813, opened 2019-02-01
fix(ui-shell): improve keyboard accessibility

dakahn:
This PR removes application menu style keyboard navigation and replaces it with a simplified tab navigation pattern. It also removes hover interaction on "menus" and replaces it with a more accessible, input agnostic, and mobile friendly expand on click. It also addresses two existing bugs on the [/design site](https://design-website.netlify.com/design/). 

1. Menu items being out of the tab order and unreachable by keyboard
2. Page crashing when a user tabs to a menu item and attempts to expand it with the arrow keys

##### HeaderNavigation
- remove handleOnKeyDown method
- clean up some commenting
- got rid of tabindex determination in _renderNavItem
- removed unused imports

##### HeaderMenu
- clean up some commeting
- remove handleMouseOver method
- remove handleMouseLeave method
- create handleOnClick method for toggling expanded menu state
- simplify handleOnKeyDown method to handle ENTER, SPACE, and ESCAPE
- explicitly declare tabindex in the render method
- remove tabindex determination in _renderMenuItem
- remove unused imports
- added some comments

##### HeaderMenuItem
- explicitly declare tabindex in the render method

#### Work left to do

Still need to get in there and remove aria role tech (menu, menuitem) found in HeaderMenu, HeaderNavigation, and HeaderMenuItem, but those attributes were closely coupled with some styling and functionality and that decoupling seemed like a job for another PR. 🏋️‍♀️ 

#### References
- [Keyboard interaction pattern by US Gov Department of Web Design](https://designsystem.digital.gov/components/headers/basic/)
- [Adrian Roselli's phenomenal blog post talking about the why you shouldn't turn website navs into application menus](http://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html)
- [Discussion on the W3C Github regarding their confusing and problematic documentation that led to navs becoming application menus and the trouble that causes](https://github.com/w3c/aria-practices/issues/353)
- [Marcy Sutton's Accessible Nav Example](https://codepen.io/marcysutton/pen/JoQqVw)
- [WCAG Fly-out Menu Guide (especially relevant is the toggle button variant)](https://www.w3.org/WAI/tutorials/menus/flyout/#flyoutnavkbbtn)
- [Heydon Pickering's Accessible Components entry for Menu's and Menu Items](https://inclusive-components.design/menus-menu-buttons/)


joshblack:
Seems like a great change 👍 Are we also updating the markup in another PR to get rid of the aria roles?

PR #1814, opened 2019-02-01
Tooltip with custom icon

devniel:
Hi,

This PR adds support for including custom react elements as icons in the `Tooltip` component using the same property (actually, I didn't found an example where the `icon` property is provided as an object):

```
<Tooltip icon={<div />} clickToOpen triggerText="Tooltip" />
```

It's useful when you want a tooltip over a custom element such as an SVG rendered with `ReactSVG` avoiding the generated `Icon`.


joshblack:
@asudoh did we want to go down the `renderIcon` route? Especially if we shift over to `@carbon/icons-react` we wouldn't necessarily need the icon descriptor. Would also allow folks to use whatever build strategy they want to include the SVG asset in the render prop.

asudoh:
@joshblack Right, the recommendation is render-function approach.

src/components/Tooltip/Tooltip.js
```diff
@@ -384,6 +394,37 @@ export default class Tooltip extends Component {
           'aria-owns': tooltipId,
         };
 
+    const finalIcon = IconCustomElement ? (
+      <div
```
asudoh: Wondering why you are using a wrapper `<div>` here... In other words, can it be:

```javascript
<IconCustomElement
  name={iconName}
  aria-labelledby={triggerId}
  aria-label={iconDescription}
  ref={node => {
    this.triggerEl = node;
  }}
/>
```

devnieL: Hi @asudoh , with that approach I got a warning about `refs` with stateless components:

`Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.` 

Also, for some reason if a wrapper is not provided the element duplicates and nests itself, breaking the position of the tooltip. I didn't understand this behavior and tried everything (even `React.Fragment` as wrapper) :

image

The given class name helps to determine if a custom icon is provided.

asudoh: OK thank you for explaining @devnieL! For the warning the user should provide a component with `React.forwardRef()`. For the positioning do you have a reduced case so I can see?

devnieL: Hi @asudoh, sorry, I got confused, I was always passing the `props` argument to the content of `renderIcon` reason because I saw some duplicated attributes, but now I fixed that. 

In respect of the positioning, it got resolved by using the `React.forwardRef()`.

So, will it be required to always use something like this ? 

```js
<Tooltip renderIcon={React.forwardRef(() => <Icon name='icon--add' />)} />
```

I think that's kind of tricky, also it's playing the role of a wrapper that could be considered inside the Tooltip component, maybe there is a way to use React.forwardRef internally when grabbing the `renderIcon` value.

devnieL: I'm looking for the proper way to valid the type of a component with `React.forwardRef`because it returns an object and not a function.

asudoh: @devnieL To better explain this - We need a DOM node of icon, so that we can [calculate the position of icon in the viewport](https://github.com/IBM/carbon-components-react/blob/v6.39.0/src/components/Tooltip/Tooltip.js#L225-L230). My reference to `React.forwardRef()` was too specific and I agree it's confusing, my apologies.

So, the answer is, you need to provide a component where `ref` prop works. Wrapping it with `<div>` makes `ref` prop works automatically, but it causes redundant DOM structure - Such technique should be at user-land therefore.

devnieL: Thanks for the explanation. If we are using refs then the render-function will always result in the warning about ` Stateless function components cannot be given refs.`. So, to confirm, React.forwardRef() will be always required, right ?

```js 
<Tooltip renderIcon={React.forwardRef((props, ref) => <Icon name='icon--add' ref={ref} />)} />
```

asudoh: Makes sense, thank you for double-checking!

src/components/Tooltip/Tooltip-story.js
```diff
@@ -106,4 +142,68 @@ storiesOf('Tooltip', module)
...
+      info: {
+        text: `
+            Tooltips are used to supply additional information to an element when hovering over it. By default,
+            the tooltip will render with an information Icon. The example below shows the option to exclude the Icon.
+          `,
+      },
+    }
+  )
+  .add(
+    'only icon',
```
asudoh:
```suggestion
    'only custom icon',
```

src/components/Tooltip/Tooltip-story.js
```diff
@@ -39,6 +39,42 @@ const props = {
...
+          width: '10px',
+          height: '10px',
+          borderRadius: '5px',
+          background: 'red',
+        }}
+        ref={ref}
+      />
+    )),
+  }),
+  onlyIcon: () => ({
```
asudoh:
```suggestion
  customIconOnly: () => ({
```

src/components/Tooltip/Tooltip-story.js
```diff
@@ -106,4 +142,68 @@ storiesOf('Tooltip', module)
...
+            the tooltip will render with an information Icon. The example below shows the option to exclude the Icon.
+          `,
+      },
+    }
+  )
+  .add(
+    'only icon',
+    () => (
+      <div style={{ marginTop: '2rem' }}>
+        <Tooltip {...props.onlyIcon()}>
```
asudoh:
```suggestion
        <Tooltip {...props.customIconOnly()}>
```

src/components/Tooltip/Tooltip.js
```diff
@@ -163,13 +163,26 @@ export default class Tooltip extends Component {
...
     triggerText: PropTypes.node,
 
+    /**
+     * The callback function to optionally render the icon element.
+     * It should be a component with React.forwardRef().
+     */
+    renderIcon: function(props, propName, componentName) {
+      if (props[propName] == undefined) return;
+
+      if (props[propName].$$typeof !== Symbol.for('react.forward_ref'))
```
asudoh: Not sure if the assumption of the React internal is safe for future React. Let's see what others think.

src/components/Tooltip/Tooltip.js
```diff
@@ -384,6 +398,37 @@ export default class Tooltip extends Component {
           'aria-owns': tooltipId,
         };
 
+    const finalIcon = IconCustomElement ? (
```
asudoh: Can we assign `Information` to `IconCustomElement` if it's not given and `componentsX` is truthy, so we can merge the block here and the next block?

src/components/Tooltip/Tooltip.js
```diff
@@ -384,6 +398,37 @@ export default class Tooltip extends Component {
...
+      <IconCustomElement
+        name={iconName}
+        className={`${prefix}--tooltip__custom-icon`}
```
asudoh: Is `className` here absolutely necessary?

src/components/Tooltip/Tooltip.js
asudoh: Also - Please see IBM/carbon-elements#231. Once you update `@carbon/icons-react` dependency in our `package.json` to the latest and re-run `yarn install`, you can `import OverflowMenuVertical16 from '@carbon/icons-react/lib/overflow-menu--vertical/16'` in `Tooltip-story`.js and put `OverflowMenuVertical16` directory to `renderIcon`. That said, I'd recommend making such update throughout `Tooltip-story.js`, or create a similar forward-ref'ed icon component and use it.

src/components/Tooltip/Tooltip.js
devnieL: I have tried `react-is` with `isForwardRef` but it didn't work well; the other solution could be to accept any React element but the developer will never know that a ref is necessary.

src/components/Tooltip/Tooltip.js
devnieL: No, I will remove it.

src/components/Tooltip/Tooltip-story.js
devnieL: ok

src/components/Tooltip/Tooltip-story.js
devnieL: ok

src/components/Tooltip/Tooltip-story.js
devnieL: ok

src/components/Tooltip/Tooltip.js
devnieL: done, the latest `OverflowMenuVertical16` is wrapped internally with `React.forwardRef` so it can be use directly, thanks!

src/components/Tooltip/Tooltip.js
devnieL: I have done something like this 

```js 
    const IconCustomElement = renderIcon || componentsX && Information;
    
    const finalIcon = IconCustomElement ? (
      <IconCustomElement
        name={iconName}
        aria-labelledby={triggerId}
        aria-label={iconDescription}
        ref={node => {
          this.triggerEl = node;
        }}
      />
    ) : (
      <Icon
        icon={!icon && !iconName ? iconInfoGlyph : icon}
        name={iconName}
        description={iconDescription}
        iconTitle={iconTitle}
        iconRef={node => {
          this.triggerEl = node;
        }}
      />
    );
```

But maybe a proper name instead of IconCustomElement, considering this change, could be `IconCustomOrInformation`.

src/components/Tooltip/Tooltip.js
devnieL: @asudoh with the latest version of `@carbon/icons-react` there is a problem with the CircleCI and netlify checks. how can it be solved ?

devnieL:
@asudoh Hi, do you know why I'm getting this error in the `src/components/UIShell/__tests__` ?, it's very strange, the package is already installed and the problem doesn't happen with other test directories.

image

Thanks

asudoh:
@devnieL Seems that it's caused by https://github.com/IBM/carbon-elements/commit/4294fec191d2775d40a44b668ed7a2b489fa2801#diff-bbfece9ac417f7dd6ec8096440b769a4. Would you try pinning `@carbon/icons-react` one version earlier? Our apologies for this - Thanks!

devnieL:
Hi @asudoh , thanks, I'll wait for #1876 to be merged.

asudoh:
Given CI passes now, one last remaining item seems running manual with IE11 - @devniel If you have a chance to do that it'd be great. Thanks!

devniel:
Hi @asudoh , what you do mean by 'running manual with IE11' ? I have checked the netlify deploy in a virtual machine with IE11 and didn't found problems.

src/components/Tooltip/Tooltip.js
emyarod: just curious, what kind of issues was `isForwardRef` causing? I'm also a bit iffy on using React internal API here so I guess we'll just have to keep an eye on this over time

src/components/Tooltip/Tooltip.js
```diff
@@ -165,13 +165,26 @@ export default class Tooltip extends Component {
      */
     triggerText: PropTypes.node,
 
+    /**
+     * The callback function to optionally render the icon element.
+     * It should be a component with React.forwardRef().
+     */
+    renderIcon: function(props, propName, componentName) {
+      if (props[propName] == undefined) return;
```
joshblack: Style nit:

```suggestion
      if (props[propName] == undefined) {
        return;
      }
```

src/components/Tooltip/Tooltip.js
devniel: Hi @emyarod, thanks for the question, `react-is` use the same approach used here, though while debugging I found that they consider first that the given data should be an `Element` and then a `forwardRef`, maybe I found a bug, reason because of doing ` ReactIs.isForwardRef(props[propName])` always returns `false`. 

I'll open an issue on `react` to confirm this behavior.

https://github.com/facebook/react/blob/master/packages/react-is/src/ReactIs.js#L30

image

devniel: @emyarod Added a new commit, the trick was to create the element with `const RefForwardingComponent = props[propName]` then `<RefForwardingComponent/>`; this as a parameter for `isForwardRef` will pass the evaluation.

emyarod: awesome thanks for the clarification and glad you figured it out @devniel! and I guess we won't have to file an issue over on the React repo either

src/components/Tooltip/Tooltip.js
```diff
@@ -7,6 +7,7 @@
 
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import * as ReactIs from 'react-is';
```
asudoh: Can we bring in only the specific import we need?

devniel: yep, I have just committed it.

asudoh:
Thanks @devniel for the update! Could you update our [Rollup config for the named exports](https://github.com/rollup/rollup-plugin-commonjs#usage) to go through the CI?

devniel:
Hi @asudoh , yes, I was curious about what's happening, let me check the rollup config.

devniel:
@asudoh do you how to know about that problem without the CI ? I ran the tests and compiled the storybook without problems.

asudoh:
Hi @devniel you can run `yarn ci-check` that's what basically runs on CI.

PR #1822, opened 2019-02-04
feat(OverflowMenu): add arrow key navigation

### LINKED ISSUE #1741: Overflow menu items need to be navigable with arrow keys
Using the keyboard, tab to an [Overflow menu](http://react.carbondesignsystem.com/?selectedKind=OverflowMenu&selectedStory=basic&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) and type space or enter to open the menu.
Now, type down arrow, which is the usual way to select menu items. Nothing happens.

Expected down and up arrow to select the items.

See also: https://github.com/IBM/carbon-components/issues/1233

### LINKED ISSUE #1742: Overflow menu needs to trap tab keys within menu
The normal way to traverse items in a menu using the keyboard is with arrow keys. (See https://github.com/IBM/carbon-components-react/issues/1741).
[Carbon Overflow menus](http://react.carbondesignsystem.com/?selectedKind=OverflowMenu&selectedStory=basic&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) are using the tab and shift+tab keys to traverse the menu.
This is ok... ([as long as arrow keys are also supported](https://github.com/IBM/carbon-components-react/issues/1741)) and as long as the user isn't allowed to tab out of the menu.

- Using the keyboard, tab to an [Overflow menu](http://react.carbondesignsystem.com/?selectedKind=OverflowMenu&selectedStory=basic&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) and type space or enter to open the menu.
- Now type the tab key through the options, and right on past the last option and into the rest of the page.
- Similarly, you can type shift+tab to go in the reverse direction and out of the top of the menu.

The user should not be able to leave the menu in this manner. They should either be trapped at the last item when moving forward, or wrap to the first item if they type tab on the last item. (Your choice whether to implement trap or wrap).

Similarly, in the reverse direction, they should not be able to exit the list of menu items with shift+tab, and should either stay on the first item or wrap to the last item.

emyarod:
Closes #1741
Closes #1742

Related discussion: https://github.com/IBM/carbon-components/issues/1629

This PR adds arrow key navigation to `<OverflowMenu>`

#### Changelog

**New**

- arrow key navigation, with keyboard trap (escapable with `tab`/`shift+tab`)
- `blur` event handler to collapse menu when focus is moved to a different element

**Changed**

- refactor `<OverflowMenu>` key handlers to use keycode matching util functions
- fix non-floating menu primary element focusing (related: #1011)
- consolidate primary element focus logic for floating and non-floating menu
- convert `<OverflowMenuItem>` to class component to allow refs


asudoh:
Thank you for jumping in @emyarod! It appears that menu closes immediately after open in `floatingMenu` mode - Does it happen your side?

emyarod:
thanks for catching that @asudoh, it should be resolved now

PR #1823, opened 2019-02-04
chore: Update wai-aria on tooltip. resolve #1494

### LINKED ISSUE #1494: A11Y: The WAI-ARIA role(s) and/or attribute(s) aria-expanded are not valid for the element div(Tooltip))
## Detailed description
Some HTML language features (i.e., element or attribute) have native semantics that map by default to an implied WAI ARIA semantics. It is possible to override implicit semantics under restricted conditions. This rule verifies adherence to those restrictions.

https://aat.w3ibm.mybluemix.net/token/54722859-73c2-4646-b1dc-fbee896e9ef9/cba2a01d-0221-451b-a29d-54d63bb7badd/archives/2018AugustDeploy//doc/w3/help/en-US/idhi_accessibility_check_g1146.html

If I remove out "aria-haspopup="true" aria-expanded="false"" from html, no error report in DAP.

Chrome

"carbon-components-react": "^6.7.4",

* Screenshots or code
image

Martskin:

Changes the Tooltip trigger element to be `<button>` instead of a `<div>` element per #1494 

#### Changelog

**Changed**

- Change `<div>` elements to `<button>` elements.
- On the Tooltip variation with Icon, the icon container element now needs the style reset that comes with the `prefix + '--tooltip__trigger'` classname.
- Moved the `role="tooltip"` attribute to the tooltip container element.

### DIFF
--- a/src/components/Tooltip/Tooltip-test.js
+++ b/src/components/Tooltip/Tooltip-test.js
@@ -29,6 +29,10 @@ describe('Tooltip', () => {
     const trigger = wrapper.find('.bx--tooltip__trigger');
 
     describe('tooltip trigger', () => {
+      it('renders a tooltip container', () => {
+        expect(trigger.length).toEqual(1);
+      });
+
       it('renders the info icon', () => {
         const icon = trigger.find(Icon);
         expect(icon.length).toBe(1);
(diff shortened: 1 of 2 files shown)


Martskin:
> Thank you for jumping in @Martskin! I think "aria-expanded on div" issue has been resolved partially by #532 with proper role. That said, did you actually see any a11y violation from the former `<div>`? Other changes look good to me.

Ahh, I see. Thanks! No I didn't see any a11y violations with button role on a `div`. I just always prefer using a button as a button when possible. I reverted and pushed.

Issue #1796, opened 2019-01-25
Allow radio button in data table


sgraden:
Data tables allow checkboxes, but should also allow some form of radio select (see images below)

## Detailed description

This would be a new feature I believe. 

Currently checkboxes are a component that can be included within the data table.
https://www.carbondesignsystem.com/components/data-table/code#skip-to-content

We are working on a feature/service and are wanting to allow the user to select a single item in a table which would then enable additional options that they can do with that selected item. Anyways, that means it is a single select at a time in a data table.
We have designed something like this:
![image](https://user-images.githubusercontent.com/6685872/51762034-423c2f80-2094-11e9-858f-7dcd4f79a32f.png)

This function is possible in structured lists, but then we aren't able to complete pagination as well. Would be nice to have something like it enabled in a standard data table.

emyarod:
cc @IBM/carbon-designers just to confirm, is this proposal one that we want to incorporate in the core library?

PR #1824, opened 2019-02-04
feat(DataTable): support radio button selection

emyarod:
Closes IBM/carbon-components-react#1796

This PR adds support for rendering data table rows with radio buttons rather than checkboxes

#### Changelog

**New**

- radio button selection support
- radio button data table selection story

**Changed**

- remove deprecated lifecycle methods in `<RadioButton>`
- add default values for required props in `<InlineCheckbox>` and `<RadioButton>`
- add default expected `onChange` prop to `<InlineCheckbox>` to remove console warning on controlled React components (setting `checked` prop without a change handler)


src/components/DataTable/TableSelectRow.js
```diff
@@ -16,19 +17,26 @@ const TableSelectRow = ({
...
+    id,
+    name,
+    onClick: onSelect,
+    checked,
+    ariaLabel,
+    disabled,
+  };
+  return (
+    <td>
+      {radio ? (
```
asudoh: How about `const CheckmarkComponent = radio ? RadioButton : InlineCheckbox` so we don't need to change the structure here...?

asudoh: Thanks @emyarod for the update - Do you think we still need [`selectionInputProps` variable](https://github.com/IBM/carbon-components-react/pull/1824/files#diff-bf0d07e8541d1f68866a05a894ee7af4R22)...?

emyarod: @asudoh I'm not entirely sure if we still need it, but we might just leave it so that we don't spread any additional props into the `<InlineInputComponent>` and for assigning `onClick` to `onSelect`

PR #1825, opened 2019-02-04
Update 'InnerClickListener' 

Calvin-Cheung:
Update InnerClickListener as the current implementation does not work if elements disappear from the DOM; if an element is clicked and it goes away such as a toggle open/close icon switching states, it should not trigger an onClickOutside event

#### Changelog

**New**

- added test case for InnerClickListener-test

**Changed**

- check if event target exists in DOM

### DIFF
--- a/src/internal/InnerClickListener.js
+++ b/src/internal/InnerClickListener.js
@@ -39,11 +39,10 @@ export default class InnerClickListener extends React.Component {
 
   handleDocumentClick(event) {
     // Ensure that the target exists in the DOM before checking the element
-    // 'body' necessary for IE
     if (
       this.element &&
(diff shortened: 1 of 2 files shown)


Calvin-Cheung:
Hi @asudoh, I have reverted back to my previous fix since searching up from the target seems more verbose and complicated than just searching within document.body. Please take a look?

asudoh:
@Calvin-Cheung thank you for asking - Come up with a counter-suggestion, hope it makes sense to you.

Calvin-Cheung:
@asudoh does that mean you'll think of a counter-suggestion? Just confirming - thank you! Hope the test helps

Calvin-Cheung:
Thank you @asudoh!

Issue #1831, opened 2019-02-05
`children` prop unused in notification components


emyarod:
In both the `<ToastNotification>` and `<InlineNotification>` components, the `children` prop is listed as a valid prop for the components in our documentation, but they are not passed through at all.

Originally reported in Slack https://ibm-cloudplatform.slack.com/archives/C2K6RFJ1G/p1549356618442300

PR #1832, opened 2019-02-05
refactor(Notification): remove unused props

emyarod:
Closes IBM/carbon-components-react#1831

This PR removes `children` from `propTypes` to prevent it from being listed in our documentation, since it is not used at all


emyarod:
related: https://github.com/IBM/carbon-components-react/issues/1834

maybe we should provide a composable notification component if design signs off on it

PR #1835, opened 2019-02-06
 feat(AccordionItem): add experimental Accordion Item

j1mie:
Hi there, I noticed that the AccordionItem is pulling in the old chevron icon when the componentsX flag is enabled, which results in the icon rendering at the wrong dimensions when used alongside the new styles.

#### Changelog

**New**

- Adds feature flag and new chevron icon to AccordionItem

**Changed**

- Updates tests


src/components/AccordionItem/AccordionItem.js
```diff
@@ -133,12 +135,22 @@ export default class AccordionItem extends Component {
...
-            className={`${prefix}--accordion__arrow`}
-            icon={iconChevronRight}
-            description={iconDescription}
-            role={null} // eslint-disable-line jsx-a11y/aria-role
-          />
+          {componentsX ? (
+            <ChevronRight16
+              className={`${prefix}--accordion__arrow`}
+              icon={iconChevronRight}
+              description={iconDescription}
```
emyarod: I believe `description` should be `aria-label` here since `description` is not a valid SVG attribute (our new icon helpers map more directly to raw SVGs compared to `<Icon>`), other than that the PR looks good to me

j1mie: I've updated that now. Thanks for reviewing!

src/components/AccordionItem/AccordionItem.js
```diff
@@ -133,12 +135,22 @@ export default class AccordionItem extends Component {
...
-            icon={iconChevronRight}
-            description={iconDescription}
-            role={null} // eslint-disable-line jsx-a11y/aria-role
-          />
+          {componentsX ? (
+            <ChevronRight16
+              className={`${prefix}--accordion__arrow`}
+              icon={iconChevronRight}
+              aria-label={iconDescription}
+              role={null} // eslint-disable-line jsx-a11y/aria-role
```
joshblack: What's the rationale again for using `null` for role? cc @emyarod

emyarod: actually that might be redundant for the new icons so we can probably remove that line

src/components/AccordionItem/AccordionItem.js
```diff
@@ -133,12 +135,20 @@ export default class AccordionItem extends Component {
...
-          <Icon
-            className={`${prefix}--accordion__arrow`}
-            icon={iconChevronRight}
-            description={iconDescription}
-            role={null} // eslint-disable-line jsx-a11y/aria-role
-          />
+          {componentsX ? (
+            <ChevronRight16
+              className={`${prefix}--accordion__arrow`}
+              icon={iconChevronRight}
```
jnm2377: I don't think the `icon prop` is needed for this. View example here: https://ibm.github.io/carbon-elements/icons-react/examples/storybook/?selectedKind=ChevronRight16&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

jnm2377: Though it still displays properly with it.

j1mie: Ah, well spotted! Have corrected this now. Cheers.

src/components/AccordionItem/AccordionItem.js
```diff
@@ -140,14 +140,12 @@ export default class AccordionItem extends Component {
...
               icon={iconChevronRight}
               aria-label={iconDescription}
-              role={null} // eslint-disable-line jsx-a11y/aria-role
             />
           ) : (
             <Icon
               className={`${prefix}--accordion__arrow`}
               icon={iconChevronRight}
               description={iconDescription}
-              role={null} // eslint-disable-line jsx-a11y/aria-role
```
asudoh: Looking at @emyarod's earlier change at https://github.com/IBM/carbon-components-react/pull/1657#discussion_r241844161, we still need `role={null}` for stable icon here. @emyarod Is it correct?

emyarod: yes we still need `role={null}` for the legacy icon 

ref: https://github.com/IBM/carbon-components-react/pull/1657/files#r242300911

j1mie: @asudoh @emyarod Have added it back in, thanks!

PR #1836, opened 2019-02-06
chore: Update react

vpicone:
Hooks are a semver minor update with no breaking changes.

https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#what-are-hooks

#### Changelog

**New**

- Hooks


joshblack:
Hi @vpicone! 👋 We might need to hold off on this till v10. Hook usage specifically can trip people up since it can be harder for them to migrate to bleeding edge given support in different tools they might use. Enzyme in particular can lag behind, unfortunately. This could easily not be the case though 😄

emyarod:
agree with @joshblack's sentiment and just as a side note I think we should emphasize https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#no-big-rewrites

vpicone:
I might have jumped the gun haha. Since it would only influence the internal version of React, could it still impact teams migrating?

@emyarod I totally agree with the rewrite overnight sentiment from the post. I was thinking there might be some opportunity to test hooks out while the experimental components are still in development.

vpicone:
@asudon I could be missing something, and usually am haha but I can’t see the breaking change aspect? Hooks were designed to work along side class components (hence the semver minor bump).

vpicone:
Ahhhh took me a bit, but I understand the motivation for waiting now. No need to force dependancy bump (even minor) if it means more friction migrating to X. Thanks for helping clarify guys.

vpicone:
Hmmm I could be wrong and I often am haha, but if we’re not exporting custom hooks I don’t see why they’d need to be a peer dependency.

Issue #1837, opened 2019-02-06
Support experimental icon strategy


emyarod:
There are still several components in our repo that exclusively use the old icon strategy. We should support our new icons behind a feature flag

Related: https://github.com/IBM/carbon-components-react/pull/1835

PR #1838, opened 2019-02-06
feat: support experimental icon strategy

emyarod:
Closes IBM/carbon-components-react#1837

This PR supports our experimental icons behind feature flags for v10 release

Depends on:

https://github.com/IBM/carbon-components-react/pull/1821
https://github.com/IBM/carbon-components-react/pull/1835

---
~~https://github.com/IBM/carbon-components/pull/1642~~
~~https://github.com/carbon-design-system/carbon-website/issues/797~~

^will handle these in a separate PR


joshblack:
Would we want to try out writing a codemod for this transformation? Could help out other teams 😄 Can go through how we could add it to carbon-upgrade, if interested

src/components/Accordion/Accordion.Skeleton.js
```diff
@@ -10,18 +10,25 @@ import React from 'react';
 import Icon from '../Icon';
 import SkeletonText from '../SkeletonText';
 import { iconChevronRight } from 'carbon-icons';
+// TODO: import { ChevronRight } from '@carbon/icons-react';
```
joshblack: I think this default entrypoint isn't going to happen because we have >2000 icon exports now 😬 

Ended up removing it from the icon package upstream because of babel having issues processing a file that large/with that many exports.

I bet we could get away with the `es` directory direct import for our components 👍

emyarod: gotcha, will prob make a new PR to remove these TODOs then

emyarod:
would a codemod be affected by inconsistent legacy icon implementations?

PR #1841, opened 2019-02-07
fix(NumberInput): add aria attributes for a11y

emyarod:
Closes IBM/carbon-components-react#1658

This PR adds `aria-live` and `aria-atomic` attributes to improve a11y

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -10,16 +10,23 @@ import React, { Component } from 'react';
 import { iconCaretUp, iconCaretDown } from 'carbon-icons';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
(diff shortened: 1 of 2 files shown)


src/components/NumberInput/NumberInput.js
```diff
@@ -293,13 +293,19 @@ export default class NumberInput extends Component {
                   <button
                     className={`${prefix}--number__control-btn up-icon`}
                     {...buttonProps}
-                    onClick={evt => this.handleArrowClick(evt, 'up')}>
+                    onClick={evt => this.handleArrowClick(evt, 'up')}
+                    aria-label="increment number"
```
asudoh: Oh should this be translatable (e.g. via prop) BTW...?

emyarod: yes good call @asudoh, added that feature

PR #1846, opened 2019-02-08
fix(Tabs): re-implement tab-content classname

lovemecomputer:
re-implements `tab-content` classname for tab content wrapper that was removed in https://github.com/IBM/carbon-components-react/pull/1784/

this came up as an issue in the [`WebsiteTabs` component](https://github.com/carbon-design-system/carbon-website/blob/master/src/components/WebsiteTabs/WebsiteTabs.js), where `.tab-content` was being used as a CSS selected to style the content tab.

do we want to keep a wrapper classname like to for project teams to use as a CSS/JS selector? in this case, the alternative is to use `.website-tabs > div`, which works but obviously could become off target if anything injects a non-tab-wrapper div as a sibling.

further note: **add prefix to this classname** as originally described in https://github.com/IBM/carbon-components-react/pull/1784/ ?

---

### 💬 

this discussion is maybe involved with a global decision about wether or not to include classnames like this that we don't specifically use for styling in a component, but may be useful in some way for developers using the components in their apps? 

(i can see the pros & cons either way, but personally lean toward including classnames or some other property that can be used as a unique selector)


asudoh:
CC @diego-codes

asudoh:
Also CC @emyarod

emyarod:
so if I understand correctly, the `.tab-content` class was hard coded in our component only for the website? In that case I agree that we should pass through a prop which allows the consumer to add classes to `<TabContent>`

lovemecomputer:
@emyarod AFAIK the `.tab-content` classname wasn't added for the website, it was just already there so i used it when implementing a website component that uses this.

in the meantime, i have changed that website component to use the selector `.website-tabs > div` to get the same effect. 

this PR is just a discussion around whether or not we want to include semantic classnames on content wrappers for anyone to use as a selector in their code, even if they aren't applied for style (In this case, `.tab-content`)

emyarod:
gotcha @lovemecomputer, yeah I think we can set a default class but still allow the user to use a custom class if they want. Would we want to assign a default class though? I don't really have a preference either way but curious to hear everyone else's thoughts

jnm2377:
> gotcha @lovemecomputer, yeah I think we can set a default class but still allow the user to use a custom class if they want. Would we want to assign a default class though? I don't really have a preference either way but curious to hear everyone else's thoughts

@emyarod We could have both. I did something similar with the carousel component in the design site. Each component needed a unique class selector for the event listeners since there were several instances of it, so I passed down a prop with a unique class for each but also had a default class I used for styling.
You can see here: https://github.com/IBM-Design/design-website/blob/f1dff81c0d876c4407241e2157b4deddae97a010/src/components/Carousel/Carousel.js#L184

emyarod:
@jnm2377 yeah the only reservation I have about hard coding `.tab-content` is that we are not using that class anywhere in the core libraries. Anything that depends on a custom class should probably be implemented in user land rather than inlined in our library, and I think https://github.com/IBM/carbon-components-react/pull/1855 would provide that flexibility to our users

jnm2377:
Ah I see you what you mean. I misunderstood the comments about the selector class earlier, I thought it was being used but I see now it's just for the website. #1855 makes sense.

lovemecomputer:
@emyarod @jnm2377 yeah, #1855 makes a lot of sense! there are probably going to be a lot of use-cases across the company's products where that is handy.

i understand your reserves about this PR @emyarod — i also don't feel super wonderful about adding classnames to an element that don't have an associated CSS rule to go with them. but, for "dev user" friendliness, there might be a lot of situations where they may need to target the `.tab-content` div in either SCSSS or JS. 

in the choice here, i personally lean toward leaving in specific selectors like `.tab-content`, at least on wrappers around semantically-meaningful content sections, or maybe some rare unique other cases. it definitely will be messy and confusing if we do this all over the place, tho, so we can pay attention on a case-by-case basis.

PR #1848, opened 2019-02-08
fix(ui-shell): add aria label to sidenav with complimentary landmark

dakahn:
Elements having a WAI-ARIA complementary role such as `<aside>` used in our sidenav must have a label specified with aria-label.

This PR gives the `<aside>` element in the sidenav an `aria-label="side navigation"` silencing the error thrown by our automated test and providing a more adequate description for our users.

### DIFF
--- a/src/components/UIShell/SideNav.js
+++ b/src/components/UIShell/SideNav.js
@@ -79,7 +79,7 @@ export default class SideNav extends React.Component {
     });
 
     return (
-      <aside className={className}>
+      <aside aria-label="side navigation" className={className}>
         <nav
           className={`${prefix}--side-nav__navigation`}
           role="navigation"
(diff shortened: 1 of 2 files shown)


src/components/UIShell/SideNav.js
```diff
@@ -79,7 +79,7 @@ export default class SideNav extends React.Component {
     });
 
     return (
-      <aside className={className}>
+      <aside aria-label="side navigation" className={className}>
```
asudoh: Should this be translatable?

dakahn: Oh, yeah probably! I'll investigate how to make that happen.

joshblack: Should this use what's in `accessibilityLabel`?

dakahn:
I I pivoted with my approach and after some research decided to nix the wrapping `<aside>` element altogether. This should simplify things. I moved the wrapping `className` attribute down to the formerly descendant `<nav>` element which seemed to work fine in the storybook.

That said, I'd give this one a close inspection just to make sure I'm not inadvertently breaking something.

Issue #1854, opened 2019-02-11
Allow custom class names for `<TabContent>`


emyarod:
Related:

https://github.com/IBM/carbon-components-react/issues/1790

https://github.com/IBM/carbon-components-react/pull/1784

https://github.com/IBM/carbon-components-react/pull/1846

The `.tab-content` class was hard coded in our component only for the website, and removing it seemed to cause a breakage. Rather than hard coding classes that are only used for our website, we should expose a prop that allows the consumer to add classes to `<TabContent>` as needed

PR #1855, opened 2019-02-11
feat(Tabs): allow custom className for `<TabContent>`

emyarod:
Closes IBM/carbon-components-react#1854

Related:

#1790

#1784

#1846

This PR exposes a `tabContentClassName` prop for the consumer to attach custom classNames to the tab content container element


asudoh:
@lovemecomputer @jnm2377 What do you think...? Thanks!

asudoh:
@jnm2377 Thank you for bringing this up! @emyarod Let's take a note, and discuss further on what we'd do with something like [`radioButtonWrapper`](https://github.com/IBM/carbon-components-react/blob/v6.60.0/src/components/RadioButton/RadioButton.js#L73) - Created: #1863

Issue #1777, opened 2019-01-22
Button: expose a 'ref' prop to allow focusing on the element


lauraalkhoury:
### Summary

In order to facilitate programmatically focusing on elements, we're asking that the Button component provides a `ref` prop that allows us to call .focus() on the html element.

> Clarify if you are asking for both design and development, or just design, or just development.

Just development.

### Justification

> Provide the business reasons for this request.

This is critical for a11y work - see this checkpoint for more info: https://www.ibm.com/able/guidelines/ci162/focus_order.html

Essentially, if the user exits from, say, a modal, we need to take the focus back to what triggered the modal. (See also https://www.ibm.com/able/guidelines/ci162/focus_order.html#PointOfRegard)

### "Must have" functionality

> Highlight any "must have" needs and functionality for the request.

Must be able to use refs to set focus on the Button.

### Specific timeline issues / requests

> Do you want this work within a specific time period? Is it related to an upcoming release?

We committed to having all a11y work complete by January 31st.

_NB: The Carbon team will try to work with your timeline, but it's not guaranteed. The earlier you make a request in advance of a desired delivery date, the better!_

lauraalkhoury:
Not sure if this is relevant, but in particular we really need this feature for the Ghost Button.

emyarod:
hi @lauraalkhoury thank you for submitting this ticket! If I'm not mistaken, [this commit](https://github.com/emyarod/carbon-components-react/commit/102f805f22b0ff8c296f9038bc84f947fb77fb85) from [this PR](https://github.com/IBM/carbon-components-react/pull/980) should have exposed refs on the `Button` component. Does your version of `carbon-components-react` include this change, or am I misunderstanding your initial request? Please let me know if the referenced PR does not address your issue

joshblack:
Closing this for now, feel free to comment or make another issue if this still occurs! 👍

Issue #1778, opened 2019-01-22
OverflowMenu: expose a 'ref' prop to allow focusing on the menu trigger


lauraalkhoury:
### Summary

In order to facilitate programmatically focusing on elements, we're asking that the OverflowMenu component provides a `ref` prop that allows us to call .focus() on the menu trigger's html element.

> Clarify if you are asking for both design and development, or just design, or just development.

Just development.

### Justification

> Provide the business reasons for this request.

This is critical for a11y work - see this checkpoint for more info: https://www.ibm.com/able/guidelines/ci162/focus_order.html

Essentially, if the user closes, say, an overflow menu, we need to take the focus back to what triggered the menu. (See also https://www.ibm.com/able/guidelines/ci162/focus_order.html#PointOfRegard)

### "Must have" functionality

> Highlight any "must have" needs and functionality for the request.

Must be able to use refs to set focus on the OverflowMenu trigger.

### Specific timeline issues / requests

> Do you want this work within a specific time period? Is it related to an upcoming release?

We committed to having all a11y work complete by January 31st.

_NB: The Carbon team will try to work with your timeline, but it's not guaranteed. The earlier you make a request in advance of a desired delivery date, the better!_

PR #1859, opened 2019-02-12
feat(components): support forwardRef for button and stateful

asudoh:
Refs #1191, #1777.

#### Changelog

**New**

- Supports `ref` for `<Button>`, `<InlineCheckbox>`, `<NumberInput>` and `<RadioButton>`.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -202,4 +202,12 @@ Button.defaultProps = {
   kind: 'primary',
 };
 
-export default Button;
+export default (!breakingChangesX
+  ? Button
+  : (() => {
+      const forwardRef = (props, ref) => {
+        return <Button {...props} inputref={ref} />;
+      };
+      forwardRef.displayName = 'Button';
+      return React.forwardRef(forwardRef);
+    })());
--- a/src/components/InlineCheckbox/InlineCheckbox.js
+++ b/src/components/InlineCheckbox/InlineCheckbox.js
@@ -8,10 +8,12 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { settings } from 'carbon-components';
+import { breakingChangesX } from '../../internal/FeatureFlags';
+import mergeRefs from '../../tools/mergeRefs';
 
 const { prefix } = settings;
 
-export default class InlineCheckbox extends React.Component {
+class InlineCheckbox extends React.Component {
   static propTypes = {
     /**
      * Specify the label for the control
@@ -97,6 +99,7 @@ export default class InlineCheckbox extends React.Component {
       onClick,
       onKeyDown,
       title = undefined,
+      forwardRef: ref,
     } = this.props;
     const inputProps = {
       id,
@@ -108,7 +111,7 @@ export default class InlineCheckbox extends React.Component {
       onKeyDown,
       className: `${prefix}--checkbox`,
       type: 'checkbox',
-      ref: this.handleRef,
+      ref: mergeRefs(ref, this.handleRef),
       checked: false,
       disabled,
     };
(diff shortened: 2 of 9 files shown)


src/tools/mergeRefs.js
```diff
@@ -0,0 +1,23 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @param {...Ref<Element>} refs List of React refs to merge.
+ * @returns {Ref<Element>} Merged React ref.
+ */
+const mergeRefs = (...refs) => el => {
+  refs.forEach(ref => {
+    // https://github.com/facebook/react/issues/13029#issuecomment-410002316
```
emyarod: thanks for providing issue reference here. wonder if there is any more discussion on why `RefObject` type definition is read only

asudoh: Yeah we need to monitor how the discussion goes. I wish React had better consideration of this use case.

asudoh:
> is this something we will have to change for each of our components that can accept refs?

Good point, I limit smaller set of changes that makes better consistency (which is, among form components). But we need to think about there are any other components that "default ref" notion can be applied - e.g. #1778

joshblack:
@asudoh what aspect of this is considered a breaking change? Is it because of the `forwardRef` refactor in the components? 

Also, is `inputref` a legacy prop name or something that's been added?

asudoh:
@joshblack Good question - For stateful components `ref` grabs the component instance, as show in the 3rd bullet at: https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs. We are changing that behavior with `forwardRef` introduction in this PR.

joshblack:
Makes sense!

RE: `inputref`, would be nice if we had it follow our camelCase patterns for prop names, e.g. `inputRef`. We should make sure these names are consistent, for example `inputRef`, `innerRef`, etc.

src/components/Button/Button.js
```diff
@@ -172,4 +172,12 @@ Button.defaultProps = {
   kind: 'primary',
 };
 
-export default Button;
+export default (!breakingChangesX
+  ? Button
+  : (() => {
```
joshblack: Are these IIFEs so that the inner block is marked as PURE and can be stripped?

src/components/InlineCheckbox/InlineCheckbox.js
```diff
@@ -97,6 +99,7 @@ export default class InlineCheckbox extends React.Component {
       onClick,
       onKeyDown,
       title = undefined,
+      forwardRef: ref,
```
joshblack: Should we use `forwardRef` since that's the method name from React? In the past we've done `innerRef`, I think.

src/components/Button/Button.js
asudoh: This IIFE is merely for better variable scoping, I'm open for other approaches.

asudoh:
Thanks for reviewing @joshblack - Wrt `inputref`, that React prop has been there before this PR and thus renaming it now will be a breaking change. Wrt `forwardRef`, the choice was from #1770, but your `innerRef` point makes sense - Made a change to align to that.

src/components/InlineCheckbox/InlineCheckbox.js
asudoh: Makes sense, made a change.

src/components/Button/Button.js
joshblack: I think it makes sense mostly for the fact that I think it'll all be stripped out in a minified build if `breakingChangesX` is false, right? Would be great to keep it around just for that.

asudoh: Cool - Let's keep the IIFE here!

asudoh:
Merging as all open comments have been addressed - Please don't hesitate to speak up if anybody has any concerns. Thanks!

PR #1868, opened 2019-02-13
fix: Change filenames to use . over dashes. Fix jest and storybook.

vpicone:
Closes IBM/carbon-components-react#1866

#### Description

Babel is failing to ignore stories and tests when we bundle es/cjs. By updating the file name convention from `-test.js` and `-story.js` to `.test.js` and `.story.js` we significantly reduce the build time and the number of files included in the es/cjs builds.

I believe this might have been caused by changes in babel's ignore pattern matching conventions&nbsp;in&nbsp;6→7. 

It took some serious finangling to get the pattern matching to match the current regex, so I think switching to this convention (as used in the babel docs) is a good path forward. Dashes could get in the way should we want to export the stories as components. 

#### Changed
- Uses a different file naming convention to easily match patterns babel recgonizes 
- `*-test.js` → `*.test.js`
- `*-story.js` → `*.story.js`
- updated build.js to use the new patterns

**Removed**

- stories and tests no longer included in es/cjs builds.
- there was a `-q` argument getting passed into babel and I couldn’t for the life of me figure out what it was supposed to be doing


vpicone:
I tried for a while this morning but couldn’t get the wildcard to match partial file names. 

The first ignore pattern with the __tests__ directory works but I didn’t have any success with getting the globs to match as is.

scripts/build.js
```diff
@@ -25,22 +25,24 @@ const exec = (command, extraEnv) =>
     env: Object.assign({}, process.env, extraEnv),
   });
 
-const ignoreGlobs = ['**/__tests__/*', '**/*-test.js', '**/*-story.js'].join(
-  ','
-);
+const ignoreGlobs = [
+  `"**/__tests__/*"`,
```
joshblack: Would these need to be:

```
**/**/*-test.js
**/**/*-story.js
```

So that it's not just top-level directories?

vpicone: I believe the double wild card ** already allows for multiple directory levels (as opposed to the single *)

vpicone:
> Any reason for not just updating the ignore globs?

With babel 7, it looks like the ignore command now uses file glob matching rather than regular string matching. https://babeljs.io/docs/en/v7-migration#path-based-only-and-ignore-patterns

They also note that the wildcards match "full slugs." In practice, I believe this means wildcards only match the **entirety** of a file name.

Accordion-test.js for example can only match *.js since * expands to include all of the filename `Accordion-test`.

joshblack:
Was chatting with @vpicone about this on Slack, might be better to use this as a chance to redo our build process. Babel transforms the glob patterns we pass into `--ignore` into regex, and subsequently it escapes any kind of `*` pattern we put in so not much that we can do on our end for this.

asudoh:
@vpicone Could you solve the merge conflicts...? Thanks!

@joshblack Did you have any other comments?

vpicone:
I think @joshblack might have been leaning on the side of refactoring the build script rather than renaming the tests/stories to be compatible with babel's pattern matching.

joshblack:
Mind if we close this in the meantime @vpicone? Might be a more involved discussion 😬

vpicone:
@joshblack sounds good to me!

PR #1869, opened 2019-02-13
feat(DropdownV2): highlight selected item in drop down

Zhao-Yi-Peng:
Selected item is highlighted with Combobox but not with DropdownV2

**Changed**

- selected item is also highlighted in the list box when open

### DIFF
--- a/src/components/DropdownV2/DropdownV2.js
+++ b/src/components/DropdownV2/DropdownV2.js
@@ -186,7 +186,11 @@ export default class DropdownV2 extends React.Component {
                   <ListBox.MenuItem
                     key={itemToString(item)}
                     isActive={selectedItem === item}
-                    isHighlighted={highlightedIndex === index}
+                    isHighlighted={
+                      highlightedIndex === index ||
+                      (selectedItem && selectedItem.id === item.id) ||
+                      false
+                    }
                     {...getItemProps({ item, index })}>
                     {itemToElement ? (
                       <ItemToElement key={itemToString(item)} {...item} />


src/components/DropdownV2/DropdownV2.js
```diff
@@ -186,7 +186,11 @@ export default class DropdownV2 extends React.Component {
                   <ListBox.MenuItem
                     key={itemToString(item)}
                     isActive={selectedItem === item}
-                    isHighlighted={highlightedIndex === index}
+                    isHighlighted={
+                      highlightedIndex === index ||
+                      (selectedItem && selectedItem.id === item.id) ||
+                      false
```
emyarod: is this `|| false` expression redundant?

Zhao-Yi-Peng: Thanks, you are completely right! Stripped the line from Combobox without closer and this was what happened. 🤕

Zhao-Yi-Peng: Hmm it seems that the expression is needed after all.

emyarod: that's strange, could you explain why it appears to be needed? I'm not sure I follow

I also checked out your feature branch and removed that part of the expression, and it appears to be functioning as expected

Zhao-Yi-Peng: Yeah I also did the same and it was working fine. However the CircleCI tests seemed to be complaining with isHighlighted={null}

emyarod: yeah after taking a look at the new test snapshots I guess we have several options here

explicit null check

```js
isHighlighted={
  highlightedIndex === index ||
  (selectedItem !== null && 
    selectedItem.id &&
    selectedItem.id === item.id)
}
```

type coercion
```js
isHighlighted={
  highlightedIndex === index ||
  (!!selectedItem && 
    selectedItem.id &&
    selectedItem.id === item.id)
}
```

or the original default to `false`

cc @asudoh

asudoh: Thank you for mulling it over @emyarod! I'm leaning toward converting the entire condition to `boolean` IIUC this topic.

PR #1873, opened 2019-02-14
chore(Tooltip): deprecate non-click-to-open in regular tooltip

### LINKED ISSUE #1864: Deprecate non-click-to-open option of default tooltip
### Summary

Our default tooltip has been moved to "click-to-open" UI, and hover-over UI has been moved to definition/icon variants. This issue issue is for tracking the effort of deprecating non-click-to-open option in default tooltip.

### "Must have" functionality

* At least make `clickToOpen` default in Storybook
* Deprecate `clickToOpen=false` in default tooltip, and then remove

Likely before we cut `v10` release.

asudoh:
Fixes #1864.

#### Changelog

**New**

- Deprecation warning for `clickToOpen=false` in `<Tooltip>`.
- Default prop for `clickToOpen`, `true` is set there for `v10`.

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -10,6 +10,7 @@ import PropTypes from 'prop-types';
 import debounce from 'lodash.debounce';
 import Icon from '../Icon';
 import classNames from 'classnames';
+import warning from 'warning';
 import { iconInfoGlyph } from 'carbon-icons';
 // TODO: import { Information } from '@carbon/icons-react';
 import Information from '@carbon/icons-react/lib/information/16';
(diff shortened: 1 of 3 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -212,6 +215,7 @@ export default class Tooltip extends Component {
     iconTitle: '',
     triggerText: 'Provide triggerText',
     menuOffset: getMenuOffset,
+    clickToOpen: breakingChangesX ? true : false,
```
emyarod: just a question but would we prefer to follow https://eslint.org/docs/rules/no-unneeded-ternary.html ?

```suggestion
    clickToOpen: !!breakingChangesX
```

but actually since it looks like we are exporting `breakingChangesX` as false we can just set the value of `clickToOpen` to `breakingChangesX` right?

```suggestion
    clickToOpen: breakingChangesX,
```

PR #1876, opened 2019-02-14
chore(dependencies): upgrade @carbon/icons-react

asudoh:
#### Changelog

**Changed**

- `@carbon/icons-react` dependency from alpha 26 to alpha 32.


devnieL:
@asudoh I think that `yarn.lock` and `.yarn-offline-mirror` changes after the proper `yarn add @carbon/icons-react@0.0.1-alpha.32` should also be committed to solve the CI problem.

joshblack:
Updated offline folder 👍 also moved to `es` directory. We should be safe including it because `es` is transpiled outside of ESM. Down to move back to `lib` though if we feel that's safer 👍

joshblack:
Ah just kidding, will move back to `lib` since Jest will complain about `es`.

asudoh:
Thanks @joshblack for the updates!!

asudoh:
Withdrawing as it appears that a similar change has got into `master` - Don't hesitate to speak up if anybody thinks otherwise - Thanks!

emyarod:
sorry @asudoh I forgot about this open PR, part of it ended up overlapping with #1838 🙇 I needed the updated dependency to fix our experimental tooltips (#1883) since the tooltip icons were not forwarding refs (https://github.com/IBM/carbon-elements/pull/231)

asudoh:
No worries @emyarod!! Though I know people tend to be frustrated by this kind of things, I 💯 understand this kind of thing may happen and it sometimes is a good thing. We can communicate, but don't be hesitant to do the right thing as you did for tooltip!

PR #1888, opened 2019-02-18
feat(DataTable): add support for object paths as keys

devniel:
Sometimes the required data to show in a DataTable are located in deep levels of an object; currently, this component requires to use direct keys to a certain value, with this PR you can use object paths to access deep properties of an object avoiding to re-format it.

```jsx
<DataTable
  rows={changes}
  headers={[
  {
    key: 'service',
    header: 'Service',
  },
   {
    key: 'baseline',
    header: 'Baseline',
   },
   {
     key: 'data.previousValue',
     header: 'Previous value',
   },
   {
     key: 'data.updatedValue',
     header: 'Updated value',
   },
]}
```


joshblack:
Hi @devniel! 👋 Thanks so much for taking the time to make this PR 😄 

With respect to added functionality, we try to suggest that one creates an issue first just to make sure it's a direction we want to take with the project, or that it aligns with future functionality of a component.

DataTable in particular is one that is likely to see changes alongside the new APIs coming out of React that could change its component API. Would you mind sharing more details about how solutions for this in user-land might not work out well? Just trying to understand more about the motivation here 👍

devniel:
Hi,

Sorry, I'll create an issue next time.

My motivation was to avoid re-formatting my object data just to fulfill the DataTable requirements, I also realized that having the data with many levels of deep makes it easy to understand when checking the headers, maybe in the future it will be useful when having sub columns.

 ```jsx
<DataTable
  rows={changes}
  headers={[
  {
    key: 'service',
    header: 'Service',
  },
   {
    key: 'baseline',
    header: 'Baseline',
   },
   {
     key: 'col1.subcol1',
     header: 'Previous value',
   },
   {
     key: 'col1.subcol2',
     header: 'Updated value',
   },
  {
     key: 'col1.subcol3',
     header: 'Another value',
   },
]}
```

**Comment by joshblack** [comment]
No problem at all 😄 

So, if I understand correctly, the main thing here is around ergonomics when passing props? In other words it'd be preferable to use `key` to describe access in an object versus having the consumer call `map` to transform for `DataTable`?

**Comment by devniel** [comment]
Hi @joshblack, yes, exactly that. Sometimes the data ready to be shown are already organized in deep levels in an object and I can't imagine '_flatting_' my object with keys like `categoryA__subcategoryA__propertyC` just for `DataTable`.

**Comment by joshblack** [comment]
@devniel do you have an example of a good nested structure for this use-case? Would love to take a look and add it to Data Table docs if appropriate. Just wanted to see what kind of options we have here for these types of data 😄 

As an aside, thanks so much for your patience going through this thread! 🙏 Really appreciate your time and involvement here!

**Comment by devniel** [comment]
The Github events API results have at least two levels depth in its JSON structure : https://api.github.com/events

It could be used like this:

```jsx
<DataTable
  ref={(e) => {this.dataTable = e}}
  rows={data}
  headers={[
    {
      header: 'Actor ID',
      key: 'actor.id',
    }, 
    {
      header: 'Actor Login',
      key: 'actor.login',
    },
    {
      header: 'Actor URL',
      key: 'actor.url'
    },
    {
      header: 'Repository Name',
      key: 'repo.name',
    },
    {
      key: 'repo.id',
      header: 'Repository ID'
    },
    {
      key: 'repo.url',
      header: 'Repository URL'
    },
    {
      key: 'type',
      header: 'Event'
    }
]             
```

But having it in one level is kind of simple, I was playing and added a way to use subheaders

```jsx
<DataTable
  ref={(e) => {this.dataTable = e}}
  rows={data}
   headers={[
    {
      header: 'Actor',
      subheaders: [
        {
          header: 'Id',
          key: 'actor.id'
        },
        {
          header: 'login',
          key: 'actor.login'
        },{
          header: 'url',
          key: 'actor.url'
        }
      ]
    },
    {
      header: 'Repository',
      subheaders: [
        {
          key: 'repo.name',
          header: 'Name'
        },
        {
          key: 'repo.id',
          header: 'Id'
        },
        {
          key: 'repo.url',
          header: 'URL'
        }
      ]
    },
    {
      key: 'type',
      header: 'event',
    }
  ]}
```
preview: https://kind-noether-f23dd7.netlify.com/
source: https://github.com/devniel/datatable-example/blob/master/src/App.js

*I have used the data directly, otherwise I would have had to create a `rows` array with one-level depth objects full of camelCase properties.

joshblack:
@devniel at one point does it become cumbersome to map a data object to the flat structure required right now by Data Table?

devniel:
Apart from traversing my data and creating another object with the flat structure, it's not a problem, `object-path` could also be used in that mapping making it easier.

joshblack:
@devniel this might be something that we'll want to keep in user-space for a bit with the changes to v10 coming. Definitely worth revisiting down the road, but for now we'll try and keep this area minimal.

Thank you so much for taking the time to make this PR, we really appreciate your time and effort here 🙏 Going to close this for now, but we should definitely take a look at this again in the future 👍

PR #1891, opened 2019-02-18
feat(Button): allows button to render arbitrary component as its root

### LINKED ISSUE #1890: [Button] allow consumer to replace root node with arbitrary component
## Summary

We are leveraging react-router https://github.com/ReactTraining/react-router for internal navigation, and react-router-dom provides its own `<Link>` component (which is basically a modified `<a>`). It would be great to be able to leverage the existing Button, which currently can render either a `<button>` or `<a>`, to instead render any arbitrary React component.

### Related

https://github.com/IBM/carbon-components-react/issues/1495
https://github.com/IBM/carbon-components-react/issues/1750

ZacharyStair:
leverages a render prop to allow consumers to render whatever they want in place of button or anchor

closes #1890


ZacharyStair:
@asudoh @joshblack what do you think?

src/components/Button/Button.js
```diff
@@ -94,6 +112,12 @@ Button.propTypes = {
    */
   children: PropTypes.node,
 
+  /**
+   * Specify how the button itself should be rendered.
+   * Make sure to apply all props to the root node and render children appropriately
+   */
+  render: PropTypes.func,
```
ZacharyStair: this proptype isn't strictly accurate; it could be either a function (i.e. a stateless functional react component), a React.component class, or even a string. it's complicated

asudoh: @Zachary String is a fair point. Given `typeof class {}` yields `function`, probably `PropTypes.oneOfType([PropTypes.func, PropTypes.string])`?

src/components/Button/Button.js
```diff
@@ -61,31 +60,23 @@ const Button = ({
...
-  const anchor = (
-    <a
+  let Component = 'button';
+  if (render) {
+    Component = render;
+  } else if (href) {
+    Component = 'a';
+  }
+  return (
+    <Component
```
asudoh: Cool - Seems that we need to cover `disabled` and `type` for `<button>`.

ZacharyStair:
the reason I picked `render` was a) it is getting to be a generally established convention across the react ecosystem (https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Route.js#L95, https://reactjs.org/docs/render-props.html#use-render-props-for-cross-cutting-concerns) and b) it isn't specific to this component, so if other components want to do something like this they can also use `render`. One potential sticking point is components that use multiple different 'render props', but it seems like you'd still want to name the most common/general render prop `render`.

src/components/Button/Button.js
ZacharyStair: fixed (sorry for the delay, i accidentally changed `inputref` to `inputRef` and was confused why all the tests were failing...)

dakahn:
Hi there 👋. Could you talk me through a specific use case for this functionality? Thanks!

ZacharyStair:
@dakahn sure thing:

I mentioned briefly in the linked issue: We're using the `Link` component from react-router-dom (https://reacttraining.com/react-router/web/api/Link) that wraps and modifies the behavior of a normal `<a>` element, and we want to use the carbon Button component (styles, functionality, etc) but instead of passing in an `href` and getting a carbon-decorated `<a>` element we want to pass a `Link` and have that be decorated with the correct classes, etc.

```
// our code (once this pr gets merged)
import { Button } from 'carbon-components-react';
import { Link } from 'react-router-dom';
...
<Button render={Link} foo="foo" to="www.ibm.com">bar</Button>
```
renders
```
<Link className="bx--btn" foo="foo" to="www.ibm.com">bar</Link>
```
which renders
```
<a className="bx--btn" foo="foo" href="www.ibm.com">bar</a>
```

dakahn:
Oh -- sorry to make you repeat yourself. Reading it now you're quite clear! 😸

joshblack:
@ZacharyStair Instead of a `render` prop, what do you think of an `as` prop pattern? (Just thinking out loud)

For example:

```jsx
// Element
<Button as="a" />
// Component
<Button as={Link} />
```

Then the internal implementation uses `React.createElement(this.props.as)` for the containing node type.

src/components/Button/Button.js
```diff
@@ -164,6 +173,7 @@ Button.propTypes = {
 };
 
 Button.defaultProps = {
+  render: null,
```
asudoh: The default prop here seems to have caused some snapshot changes - Probably we can leave it undefined?

src/components/Button/Button.js
```diff
@@ -61,31 +64,35 @@ const Button = ({
...
+    component = as;
+    otherProps = {
+      ...otherProps,
+      ...anchorProps,
+    };
+  } else if (href) {
+    component = 'a';
+    otherProps = anchorProps;
+  }
+  return React.createElement(
```
ZacharyStair: i think the jsx form might be clearer but either way works

ZacharyStair: it seems that the main difference is `createElement` automatically forwards refs, but we get an `inputRef` instead of a `ref` anyway so that part isn't strictly relevant

joshblack:
Forgot to include this before, but as a reference for that `as` pattern we have something similar in UI Shell called `Link`: https://github.com/IBM/carbon-components-react/blob/master/src/components/UIShell/Link.js

In this case, it uses `element` but I think we will change that before stable to `as` as noted above.

Issue #1909, opened 2019-02-22
Storybook button with custom root node not configurable via knobs


emyarod:
The button type for the custom root node example is hard coded in the storybook. The `<Link>` component also contains a class that does not have enough contrast with the button background colors so a different root node may be better for this demo

![image](https://user-images.githubusercontent.com/8265238/53257173-22962800-368f-11e9-8e61-f413e7224334.png)

PR #1910, opened 2019-02-22
docs(Button): add example with custom link component as root

emyarod:
Closes IBM/carbon-components-react#1909

This PR adds support for configuring button examples that use custom root nodes


lovemecomputer:
curious if there is any downside to moving away from the React `<Link>` component… otherwise looks good!

emyarod:
it's just a component example so I don't see any down side. The existing example would have color contrast problems if we add knobs to the custom example though

image

PR #1915, opened 2019-02-25
chore(components): set minimum React version

### LINKED ISSUE #1895: Set minimum React version requirement for v10
Refs #1836.

### Summary

This is is for tracking an effort of setting minimum React version requirement for v10.

### Justification

Recent React releases contain some intriguing features, like Hooks. If we can set a minimum React version requirement for v10, we can introduce new features that depend on such React new features afterwards.

* Setting a minimum React version requirement
* Provide an easy way for users to know above version requirement

### "Must have" functionality

Either (or both) of:

* Documentation
* React version detection and warning code

Before we ship `v10`.

asudoh:
For next release.

Fixes #1895.

#### Changelog

**New**

- `checkReactVersion` internal module, that checks React module version with the minimum React version required for our library (`16.8.2` for our `v10`).

**Changed**

- Component modules to bring in above module.


joshblack:
Looking quickly at this, is there a way to do this once in the entrypoint instead of per-component? If not, would this warning show up per-component?

And with the `checkReactVersion` block, I bet we could have it as:

```js
// Flag that this code is available only in dev
if (__DEV__) {
  // Make inner assertions
  if (breakingChangesX && semver.lt(React.version, '16.8.2')) {
    // side-effects can go here
  }
}
```

If `__DEV__` evaluates to `false` it should be removed in either that form or the `if (__DEV__ && breakingChangesX && ...)` form without the need for the closure.

asudoh:
@joshblack Thank you for reviewing! Fair point on closure - Made a change.

> is there a way to do this once in the entrypoint instead of per-component

Note that the module runs in module creation time and thus runs only once (as module bundlers de-dupes modules). You can see that by replacing `"react": "^16.8.2"` with `"react": "16.8.1"`, running `yarn install` and then `yarn storybook`. Considered pulling `checkReactVerson` module in just for `src/index.js` as you seems to have said, but I figured it's brittle considering imports by direct module paths.

joshblack:
Ah okay, i see what you're saying now with it running on initialize 👍 

With respect to what we want to achieve with this, what benefits does this approach provide when compared to raising the `peerDependencies` version? Is this so that it's aware to the developer in console that a version mismatch has occurred?

asudoh:
> he only reason the per-component thing throws me off is that it feels risky to introduce this constraint without some associated rule to make sure we are doing this. Makes me wonder if specifying in entrypoint is good enough as those who directly import would most likely understand this.

OK fair enough!

> With respect to what we want to achieve with this, what benefits does this approach provide when compared to raising the `peerDependencies` version? Is this so that it's aware to the developer in console that a version mismatch has occurred?

A good question - This is for setting the version requirement before we ship `v10` (canary releases we are doing with feature flags), which requires conditional check of React version (`16.8.2` for `v10` and `16.4-ish` for `v9`).

asudoh:
Withdrawing for now, we may probably go with updating `package.json` as a last-minute pre-release task.

Issue #1852, opened 2019-02-11
Tooltip hand icon should be only for Icon 


arnajdas1992:
## Detailed description
When we have a tootlip without the Tooltip icon, the trigger text should have a hand icon when hovered. But when there is an icon, the hand icon should come for only the icon and not the text.

> Is this issue related to a specific component?
Tooltip

> What did you expect to happen? What happened instead? What would you like to see changed?
The hand icon should come for only the icon and not the tooltip text. The hand icon comes for the text also. Yes, the hand icon should come only for the icon in this case.

> What browser are you working in?
Chrome, firefox

> What version of the Carbon Design System are you using?
6.14.2

emyarod:
hi @arnajdas1992 thanks for submitting this ticket! Would you be able to provide a [Code Sandbox](https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox) describing your current usage of the tooltip so I can fully understand your request?

arnajdas1992:
hi @emyarod  - https://codesandbox.io/s/ypw9qz9n61. This is the sandbox  code. When you hover over the tooltip icon, the cursor changes to a hand icon. That is correct, but the hand icon comes up even when you hover over the text, giving the impression that, it is also clickable or has a tooltip on it, which is wrong. The requirement is that, when there is a tooltip icon present, the cursor shouldn't change when hovered over the text

joshblack:
Thanks for the steps to reproduce @arnajdas1992! 🙏 

For anyone looking to contribute the fix, I believe this is caused by the outermost node having the class `bx--tooltip__trigger` versus `bx--tooltip__label` 👍

PR #1917, opened 2019-02-25
fix(Tooltip): show pointer cursor only on icon hover

emyarod:
Closes IBM/carbon-components-react#1852

This PR assigns correct `className`s to the tooltip wrapper and trigger


src/components/Tooltip/Tooltip.js
```diff
@@ -416,8 +416,9 @@ export default class Tooltip extends Component {
     );
 
     const triggerClasses = classNames(
-      { [`${prefix}--tooltip__trigger`]: !componentsX },
-      { [`${prefix}--tooltip__label`]: componentsX },
+      // { [`${prefix}--tooltip__trigger`]: !componentsX },
```
asudoh: Did you mean to remove those two lines...?

emyarod: yes I forgot to remove them while modifying the tests, should be fixed now

Issue #1916, opened 2019-02-25
RadioButton still not working inside Toolbar


asfordmatt:
This bug was recently marked as resolved, but the issue is not fixed:
https://github.com/IBM/carbon-components-react/issues/1885

See here:
https://codesandbox.io/s/148z6wn694
I've updated it to use v6.98.2, and the onChange handler is still not firing. Should this work or is there something wrong with the code?

I see this warning in the console:
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `OverflowMenu`.
    in ToolbarOption (created by OverflowMenu)
    in ul (created by OverflowMenu)
    in div (created by ClickListener)
    in ClickListener (created by OverflowMenu)
    in OverflowMenu (created by App)
    in ToolbarItem (created by App)
    in div (created by Toolbar)
    in Toolbar (created by App)
    in App
```

asfordmatt:
FYI @asudoh @emyarod

emyarod:
thanks for the update on this issue @asfordmatt. for now, it seems you can render `<OverflowMenu>` with the `floatingMenu` prop and it should work as a temp fix, and I will look into this for non-floating menus

emyarod:
in any case it would probably be best to use floating menus from here on out, as we may look into deprecating non-floating menus

asfordmatt:
Thanks @emyarod I can switch to floatingMenu. Presumably I shouldn't be getting that 'refs' warning? Can you use this issue to address that?

emyarod:
@asfordmatt #1920 addresses this ticket. I'll open a new one for the console warnings

PR #1920, opened 2019-02-26
fix(OverflowMenu): assign menu body ref for non-floating menu

emyarod:
Closes IBM/carbon-components-react#1916

This PR fixes an issue with non-floating overflow menus not properly handling click and change events before collapsing and unmounting. Non-floating overflow menus were referencing a value that never received a value when the menu was opened. Although given the limited use case of non-floating overflow menus, it may be best to deprecate this option in the future


asudoh:
@IBM/carbon-developers Can anybody do another review...? Thanks!

Issue #1924, opened 2019-02-26
OverflowMenu assigning refs to stateless functional components


emyarod:
Reported in #1916

The console is showing warnings about assigning refs to stateless functions when using `<OverflowMenu>`

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail.

Check the render method of `OverflowMenu`.
    in ToolbarDivider (created by OverflowMenu)
    in ul (created by FloatingMenu)
    in FloatingMenu (created by OverflowMenu)
    in div (created by OverflowMenu)
    in div (created by ClickListener)
    in ClickListener (created by OverflowMenu)
    in OverflowMenu (created by Container)
    in ToolbarItem (created by Container)
    in div (created by Toolbar)
    in Toolbar (created by Container)
    in div (created by Story)
    in div (created by Story)
    in Story (created by Container)
    in div (created by Container)
    in StrictMode (created by Container)
    in Container
```

PR #1925, opened 2019-02-26
fix(OverflowMenu): assign refs only to class components

emyarod:
Closes IBM/carbon-components-react#1924

Function components cannot accept refs, but `<OverflowMenu>` is attempting to assign refs regardless of child type. This PR checks if the children of `<OverflowMenu>` can accept refs before assigning them

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -485,9 +485,12 @@ class OverflowMenu extends Component {
         React.cloneElement(child, {
           closeMenu: this.closeMenu,
           handleOverflowMenuItemFocus: this.handleOverflowMenuItemFocus,
-          ref: e => {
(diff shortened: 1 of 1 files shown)


joshblack:
Let's go with what @asudoh suggested and require child items to have a ref. Is there any documentation that we should use to address this new requirement in v10?

dakahn:
@emyarod 👋 Didn't want this one going `wontfix`. Was there an update on this PR?

emyarod:
possibly replaced by https://github.com/carbon-design-system/carbon-components-react/pull/2324 cc @asudoh

emyarod:
#2324 is a stop gap for now, this PR can be marked as low priority while I continue experimenting with possible solutions. @asudoh and I discussed maybe using render props to control which menu items get assigned `ref`s

PR #1930, opened 2019-02-27
feat(RadioButton): support labelPosition prop

### LINKED ISSUE #1913: Radio Button - Label position
### Summary

This issue is for tracking an effort of e.g. top-positioned radio button label, as reported in IBM/carbon-components#1721, to the React variant.

### Justification

For better inconsistency between `carbon-components` and `carbon-components-react`. Above feature can be used in `carbon-components-react` today by adding CSS class names, but:

* There is no example in this repo
* Using a prop instead of CSS class names, makes the feature easier to use

Provide an easy way to use the feature in IBM/carbon-components#1721.

### "Must have" functionality

* At least a Storybook example of top-positioned label.

`v10`, or soon after `v10`.

asudoh:
Fixes #1913.

#### Changelog

**New**

- `labelPosition` prop to `<RadioButton>` and a Storybook knob for it

**Changed**

- Code to use `bx--radio-buttton-wrapper` class instead of `radioButtonWrapper` class for `v10`, as the latter seems not used anywhere.


package.json
```diff
@@ -179,7 +179,7 @@
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-react-docgen": "^2.0.0",
     "bowser": "^1.6.1",
-    "carbon-components": "^9.69.0",
+    "carbon-components": "^9.78.0",
```
emyarod: is this PR dependent on https://github.com/IBM/carbon-components/pull/1933? both netlify and local build don't appear to reposition the label when the knob is modified

asudoh: @emyarod Using the knob in experimental requires 
[IBM/carbon-components#1933](https://github.com/IBM/carbon-components/pull/1933) and that's why it doesn't work with Netlify link. That said, to test it locally, it needs to be with non-experimental mode and pull in `9.78.0`, which contains IBM/carbon-components#1722. Please don't hesitate to ping me if you still see the problem - Thanks!

emyarod (approved):

asudoh:
@IBM/carbon-developers Can somebody do another review...? Thanks!

src/components/RadioButton/RadioButton-story.js
```diff
@@ -9,15 +9,27 @@ import React from 'react';
...
-import { withKnobs, boolean, text } from '@storybook/addon-knobs';
+import { withKnobs, boolean, select, text } from '@storybook/addon-knobs';
 import RadioButton from '../RadioButton';
 import RadioButtonSkeleton from '../RadioButton/RadioButton.Skeleton';
 
+const labelPositions = {
+  'Left (left)': 'left',
+  'Top (top)': 'top',
+  'Right (right)': 'right',
+  'Bottom (bottom)': 'bottom',
```
aledavila: If we are no longer supporting labels to be at the top and bottom on v10 do we want to still add this or can these be removed for v10 ?

aledavila (changes requested):

asudoh:
@aledavila Thank you for pointing this out! Fixed.

src/components/RadioButton/RadioButton-story.js
asudoh: @aledavila Thank you for pointing this out! Fixed.

aledavila (approved):
Thank you! Looks great

Issue #1944, opened 2019-02-28
Allow ProgressIndicator to be interactive.  Support passing an onChange prop


scottdickerson:
Use this template if you want to request a new feature, or a change to an existing feature.

If you are reporting a bug or problem, please use the bug template instead.

### Summary

Enhancement request to allow the ProgressIndicator component to be interactive and support updating the progress by clicking on a ProgressStep.

### Justification

Our users expect to be able to click on the ProgressStep and have the Progress indicator update.

### Desired UX and success metrics

Describe the full user experience for this feature. Also define the metrics by which we can measure success/failure for the user.

### "Must have" functionality

Publish the selected Step Index to the onChange property to allow listeners to update the currentIndex prop of the ProgressIndicator

### Available extra resources

I have a Pull Request that I'll submit to add this feature if approved.

PR #1945, opened 2019-02-28
feat(ProgressIndicator): add optional interactivity

scottdickerson:
Closes IBM/carbon-components-react#1944

{{short description}}

#### Changelog

**New**

- Added new onChange prop to ProgressIndicator that will attach a keyboard/click handler to each ProgressStep. It will callback with the selected index.  This should be accessible and clickable.


src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -37,6 +38,14 @@ export const ProgressStep = ({ ...props }) => {
     [className]: className,
   });
 
+  const handleKeyDown = e => {
+    const key = e.key || e.which;
+
+    if (key === 'Enter' || key === 13 || key === ' ' || key === 32) {
```
emyarod: we may be able to use our [keyboard event handler helpers](https://github.com/IBM/carbon-components-react/blob/master/src/tools/key.js) here

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -88,21 +97,37 @@ export const ProgressStep = ({ ...props }) => {
...
-      <ProgressStepLabel className={`${prefix}--progress-label`}>
-        {label}
-      </ProgressStepLabel>
-      {componentsX &&
-      secondaryLabel !== null &&
-      secondaryLabel !== undefined ? (
-        <p className={`${prefix}--progress-optional`}>{secondaryLabel}</p>
-      ) : null}
-      <span className={`${prefix}--progress-line`} />
+      <div
```
emyarod: it appears that the experimental version of this component is broken with this change. I think we may have to incorporate the new markup and add new style rules to the vanilla component before we can merge this into the React repo

emyarod:
thanks for jumping in @scottdickerson! I think functionally this PR is solid, but it looks like these changes will break the appearance of the experimental version, so we will probably first implement these changes on the vanilla side before merging this PR in

scottdickerson:
thank you for catching these issues @emyarod I've updated the PR and I think fixed both, can you take another look?

emyarod:
thanks for the quick update @scottdickerson! It looks like overflowing content is not showing the ellipsis and most likely requires tweaks from the vanilla side if I'm not mistaken, which I'm happy to help out with once we get a ticket opened up for style and markup updates over there

image

image

I also wonder if the design team wants to modify the focus state so that it's not the browser default 
cc @IBM/carbon-designers 

image

image

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -88,21 +96,37 @@ export const ProgressStep = ({ ...props }) => {
...
+        style={{
+          display: 'inline-flex',
```
asudoh: Wondering what this style is for, but if this style is absolutely necessary, could you please create a PR to https://github.com/IBM/carbon-components/pulls with the corresponding style ruleset?

joshblack:
Hey @scottdickerson! 👋 Just wanted to check in on this PR, would you have time to address @asudoh's comment over in: https://github.com/IBM/carbon-components-react/pull/1945#discussion_r261785240 ?

scottdickerson:
sure @joshblack I'll adjust the styles under this PR for the vanilla comp

https://github.com/IBM/carbon-components/issues/1652

scottdickerson:
ok added the requested changes, please take a look @asudoh and @emyarod when you get a chance.  If the carbon-components PR is accepted and merged, please rev the dependency to match to pickup the new styles.

scottdickerson:
@joshblack I updated the dep to point to the latest carbon-components that has the new styles, can you take a look?  Thanks!

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -88,21 +96,35 @@ export const ProgressStep = ({ ...props }) => {
...
-        <p className={`${prefix}--progress-optional`}>{secondaryLabel}</p>
-      ) : null}
-      <span className={`${prefix}--progress-line`} />
+      <div
+        className={classnames(`${prefix}--progress-step-button`, {
+          [`${prefix}--progress-step-button--unclickable`]: !onClick || current,
+        })}
+        role="button"
+        tabIndex={!current && onClick ? 0 : -1}
+        onClick={!current ? onClick : undefined}
```
emyarod: the PR description states that each progress step will have a click handler attached, but it looks like the current progress step will not be tabbable or receive the click handler. I just want to confirm, is this the expected behavior?

scottdickerson: Yeah @emyarod in our user testing we got feedback that we shouldn't allow them to click on the current step.

emyarod: gotcha, thanks for clarifying @scottdickerson !

src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -199,25 +231,37 @@ export class ProgressIndicator extends Component {
...
 
-  renderSteps = () =>
-    React.Children.map(this.props.children, (child, index) => {
+  renderSteps = () => {
+    const { onChange } = this.props;
+
+    return React.Children.map(this.props.children, (child, index) => {
+      // only setup click handlers if onChange event is passed
+      const onClick =
+        onChange && index > -1 ? () => onChange(index) : undefined;
```
asudoh: Do you anticipate `index` being negative in `React.Children.map()` callback?

scottdickerson: you're right, great catch I've removed that in the latest commit.

Issue #1445, opened 2018-10-15
Add FooterItem component


vpicone:
### Summary

I think it'd be cool to easily add additional Footer items without needing to go look up the appropriate class names and dom heirarchy.

This is an example of how writing a Footer looks now with the default styling in place:

https://codesandbox.io/s/2x6453ypxp

### Justification

The "labelOne, linkTextOne, linkHrefOne, labelTwo,linkTextTwo,linkHrefTwo" could signal to individuals that the Footer component is more strict than it actually is. Or someone might wonder "is there a label/link/linkText Three?

If someone wants to add additional links, the need to go and mimic the styling of the first to elements or start from scratch.

From the carbon-react slack:
> `<Footer> Component not supporting more than two link to add. Do we have any solution to fix this issue. Please share your in case any solution or workaround to this issue.`

### "Must have" functionality

User's should be able to supply a link, label, and linkText to a FooterItem and have it handle the styling to match the default link/label/linktext styles

Additionally, users could be able to select a `type` property from `['link', 'cta']`

### Available extra resources

I could implement this if folks think it's a good idea.

joshblack:
Definitely like the suggestion! I'd be down for adding it in if others are 👍

vpicone:
Cool, I'll give it a shot then!

joshblack:
Thanks for the help @vpicone! 🙏

xrissot-ibm:
Hi,
If I can add one item to the wish list, please add also the target attribute to the link in order to be able to open the link in a new window for example.

vpicone:
@xrissot-ibm I think that's a nice thought. I think the best API for this would allow folks to supply their own Link component. I'm gonna revisit this.

vpicone:
Closing this since the footer is being handled in addons now.

PR #1954, opened 2019-03-03
feat: Add footer component

vpicone:
Closes IBM/carbon-components-react#1445

At the moment, the footer takes four separate props to create 2 footer items. If more than 2 items are needed, devs need to copy the markup. The FooterItem is intended to be used as a child of the Footer component.

#### Changelog

**New**

- ++FooterItem

**Changed**

- Footer stories now include an example using children (FooterItem)

### DIFF
--- a/src/components/Footer/Footer-story.js
+++ b/src/components/Footer/Footer-story.js
@@ -11,6 +11,8 @@ import { action } from '@storybook/addon-actions';
 
 import { withKnobs, text } from '@storybook/addon-knobs';
 import Footer from '../Footer';
+import FooterItem from '../FooterItem';
+import Link from '../Link';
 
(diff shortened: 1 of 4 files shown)


asudoh:
Thank you for your attempt to address this @vpicone! My apologies for not telling about #1950 as soon as I should have, which is, `<Footer>` is being deprecated. Given footer comes from Cloud BU requirement, I think this PR should go to https://github.com/carbon-design-system/carbon-addons-cloud-react repo. Interesting to see if people the wants to take over `<Footer>` itself, too.

vpicone:
No worries! Honestly, I wasn’t a fan of the Footer props from the start so this is great haha.

Issue #1899, opened 2019-02-19
Datepicker component uses non-accessible Flatpickr calendar dropdown


dakahn:
### Problem

The Datepicker relies on the inaccessible Flatpickr component to display a calendar dropdown. This calendar dropdown has [strange or missing keyboard navigation](https://github.com/flatpickr/flatpickr/issues/1636) and is being [flagged by our automated accessibility testing system](https://github.com/IBM/carbon-components/issues/1471) (DAP). 

### Solution

Two possible paths forward: we can submit a PR to Flatpickr, but given the release schedule and that work on their end [has slowed](https://github.com/flatpickr/flatpickr/issues/1646) a better long term (though more painful potentially short term) solution may be to convert our date input to a[ native html 5 calendar dropdown](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date).

bakerTX:
We're also having a _lot_ of trouble with this component. Not only is the datepicker throwing DAP violations, but it straight up **trapping** keyboard focus: When using the 'range with calendar' option, it seems impossible to tab out of the input in the forward direction.

trying to tab forward:
![feb-25-2019 16-34-27](https://user-images.githubusercontent.com/8573243/53373508-4bcae880-391b-11e9-828e-e241a311db53.gif)

chmln:
Hey guys, flatpickr author here.

@bakerTX I can't fix problems that are not reported. This is especially pertinent to plugins that have a "beta" status. I have gone ahead and fixed the trapping issue. The patched version is available right now (release v4.5.5).

@dakahn this sounds disingenuous. The accessibility issue you've opened not only contains incorrect information but is not actionable at all. The project has always been receptive to pull requests; and contrary to your tone is very well maintained with 4 developers onboard at this point.

asudoh:
Hi @chmln :wave: thank you so much for chiming in! And our apologies that the issue description here may have been too general against Flatpickr. In fact, though I wasn't on Carbon team when Carbon started using Flatpickr, I could see at then that Flatpickr choice for our date picker allows us to develop the component very quickly meeting our design requirement as well as g11n requirements at that time.

OTOH our design/UX requirements around range mode date picker and its difference from what Flatpickr provides out-of-the-box (or with a plugin) have required some tweaks wrt how our code interacts with Flatpickr. Most of such portion has been gone by using range plugin, though as you mentioned it's in "beta" status. But there are still some, and the combination of those yielded several issues opened still, where:

[...]

PR #1957, opened 2019-03-04
fix(DatePicker): fix tabbing through in range mode

asudoh:
Refs https://github.com/IBM/carbon-components-react/issues/1899#issuecomment-467212036. Accordion to Flatpickr author, this issue has been fixed in `4.5.5` and I can see it in our Storybook environment.

#### Changelog

**Changed**

- Flatpickr dependency to `4.5.5`


asudoh:
@emyarod Yeah the plan is opening up tickets. I think it starts with request for supporting our UX pattern, especially in range mode.

Issue #1965, opened 2019-03-04
Date picker does not match experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1927

https://github.com/IBM/carbon-components/pull/1789

PR #1966, opened 2019-03-04
fix(DatePicker): match experimental spec

emyarod:
Closes IBM/carbon-components-react#1965

Related:

IBM/carbon-components#1927

IBM/carbon-components#1789

#### Changelog

**Changed**

- update `carbon-components` version
- update flatpickr calendar SVG icons
- update flatpickr config option for previous month arrow in calendar view
- wrap date picker input in container `div` to match spec


asudoh (approved):
LGTM 👍, just one question - Thanks @emyarod!

src/components/DatePicker/DatePicker.js
```diff
@@ -329,14 +329,26 @@ export default class DatePicker extends Component {
   };
 
   rightArrowHTML() {
-    return `
+    return componentsX
+      ? `
+      <svg width="16px" height="16px" viewBox="0 0 16 16">
```
asudoh: Does `carbon-elements` have a corresponding icon...? If so, would it be a option using React string renderer, or would it be a too much overhead for client-side codebase...?

emyarod: I believe the `carbon-elements` source has hard coded ID and class attributes, but I think we could use `@carbon/icon-helpers` to convert it to a string

asudoh: OK if we can do it build-time instead of browser run-time it'd be great, but we can do it after `v10`. Thank you for your insights @emyarod!

aagonzales (commented):
When open the end range calendar, the calendar should be aligned to the edge of the start range input not the end range input.

This is wrong:
image

asudoh:
Hi @aagonzales thank you for pointing this out! Given the issue happens in `v9` as well and it has to do with what Flatpickr does behind the scene, created #1987 to track this.

emyarod:
since this exists as a carry over from `v9`, is this something we need to address before `v10`?

joshblack:
@emyarod if it's in v9, then it should be fine to carry over to v10 and we can fix in there 👍

aagonzales:
Ok, so its not to spec so I'm going to wait to approve it until that change is made.

asudoh:
> When open the end range calendar, the calendar should be aligned to the edge of the start range input not the end range input.

Filed flatpickr/flatpickr#1693 as it's how Flatpickr works with range mode.

dakahn (approved):
Code looks good to me. 👍🏽

asudoh:
Enqueued #1987 to `v10-rc1` bucket - @aagonzales Is the dropdown location the only issue? If so, can we go ahead and track the dropdown location issue in #1987? Thanks!

aagonzales:
Yes, that's the only issue. If we open up another issue to keep track of this problem then we can move forward with the PR.

asudoh:
Thanks @aagonzales for clarifying!

PR #1975, opened 2019-03-05
fix(number-input): use required prop `iconDescription` in CarbonX icons

### LINKED ISSUE #1974: [Carbon X] Unused required prop `iconDescription` in NumberInput
## Detailed description

The Carbon X version of the `NumberInput` no longer uses the required prop `iconDescription`.

See this block where the icons are generated: https://github.com/IBM/carbon-components-react/blob/master/src/components/NumberInput/NumberInput.js#L276

jendowns:
Closes #1974 

Use required prop `iconDescription` if it is provided (similar to: https://github.com/IBM/carbon-components-react/blob/master/src/components/NumberInput/NumberInput.js#L334)

#### Changelog

**Changed**

- use `iconDescription` or Carbon X icons
- updated `this.props.iconDescription` to `iconDescription`


jendowns:
Thank you for the suggestions @asudoh! 🎉

Issue #1750, opened 2019-01-15
Support "icon as a React component" in various components


asudoh:
### Summary

Currently there are several components that allows consumers to pass along the icon. We added support to "icon as SVG structure" some months ago. Given we are adding notion to "icon as React component" notion in `v10`, we'd want to support it, too. An alternate interface there may be a render-function.

### Justification

Better `v10` new feature support

### "Must have" functionality

"Icon as React component" support in above components

### Specific timeline issues / requests

Before `v10` release (Or soon after)

### Available extra resources

Here's the list from a quick search:

```
/carbon-components-react/src/components/Button/Button.js:
  131     * representing the SVG data of the icon
  132     */
  133:   icon: PropTypes.oneOfType([
  134      PropTypes.shape({
  135        width: PropTypes.string,

/carbon-components-react/src/components/DataTable/TableToolbarAction.js:
   36       * Specify the icon for the toolbar action
   37       */
   38:     icon: PropTypes.shape({
   39        width: PropTypes.string,
   40        height: PropTypes.string,

/carbon-components-react/src/components/Icon/Icon.js:
  182       * The icon data.
  183       */
  184:     icon: PropTypes.shape({
  185        width: PropTypes.string,
  186        height: PropTypes.string,

/carbon-components-react/src/components/ModalWrapper/ModalWrapper.js:
   23      handleSubmit: PropTypes.func,
   24      disabled: PropTypes.bool,
   25:     triggerButtonIcon: PropTypes.string,
   26      triggerButtonIconDescription: PropTypes.string,
   27      triggerButtonKind: ButtonTypes.buttonKind,

/carbon-components-react/src/components/Notification/Notification.js:
   42       * if something but regular "close" icon is desirable
   43       */
   44:     icon: PropTypes.shape({
   45        width: PropTypes.string,
   46        height: PropTypes.string,

/carbon-components-react/src/components/OverflowMenu/OverflowMenu.js:
  200       * The icon.
  201       */
  202:     icon: PropTypes.shape({
  203        width: PropTypes.string,
  204        height: PropTypes.string,

/carbon-components-react/src/components/Tooltip/Tooltip.js:
  164      /**
  165       * The the default tooltip icon.
  166       */
  167:     icon: PropTypes.shape({
  168        width: PropTypes.string,
  169        height: PropTypes.string,

/carbon-components-react/src/components/TooltipSimple/TooltipSimple.js:
   98    /**
   99     * The the default tooltip icon.
  100     */
[...]

PR #1977, opened 2019-03-06
chore(components): support SVG title attribute for icon description

asudoh:
Refs #1750.

#### Changelog

**New**

- `<title>` for experimental icons (and custom icons via `renderIcon`).
- `aria-label` for experimental icons, where there are _not_ `aria-hidden`.

**NOTE**: Didn't do above for `<NumberInput>` as it's covered by: #1975
**Removed**

- `aria-label` from experimental icons, where there are `aria-hidden`.


src/components/CopyButton/CopyButton.js
```diff
@@ -99,8 +99,9 @@ export default class CopyButton extends Component {
           <Copy16
             className={`${prefix}--snippet__icon`}
             aria-label={iconDescription}
-            alt={iconDescription}
-          />
+            alt={iconDescription}>
```
emyarod: is `alt` a valid attribute here? I think the `<title>` element is supposed to replace our invalid usage of `alt` attribute on SVGs

asudoh: Great catch @emyarod! - Fixed.

src/components/AccordionItem/AccordionItem.js
```diff
@@ -138,8 +138,9 @@ export default class AccordionItem extends Component {
           {componentsX ? (
             <ChevronRight16
               className={`${prefix}--accordion__arrow`}
-              aria-label={iconDescription}
-            />
+              aria-label={iconDescription}>
```
joshblack: Do we need to double-up on `aria-label` and `title`?

asudoh: Good question @joshblack - `<title>` is for hover-over text whereas `aria-label` is for screen reader, but don't hesitate to speak up if you believe one covers another.

emyarod:
@asudoh it looks like we there are a few conflicts that need to be resolved now

src/components/AccordionItem/AccordionItem.js
joshblack: @asudoh we've gotten feedback around the tooltips themselves from `title`, might be helpful for icons to default to not having it. If someone would like a `title`, they should place it on the interactive node (like `<button>`) that it's associated with.

joshblack:
bump @asudoh for merge conflicts, PR looks great! 👍

asudoh:
Thanks everyone for all the review and @joshblack for pointing out the surrounding `<button>` thing! Given @vpicone worked on that area I see many of my changes here are redundant. I'm planning to sort it out next week - Which may end up with updating this PR or withdrawing this PR.

asudoh:
Removed `<title>` in `v10` icons where there are surrounding `<button>` or `<div role="button">` with `title` attribute.

Issue #1979, opened 2019-03-06
Add ability to render any content for HeaderMenu


davidicus:
Use this template if you want to request a new feature, or a change to an existing feature.

If you are reporting a bug or problem, please use the bug template instead.

### Summary

Currently HeaderMenu component only renders the string provided as aria-label. This is a feature request to allow us to pass anything to render whether its an icon or text and an icon. 

This would just be a development request

### Justification

Our product uses icons and sometimes multi line strings and icons in our header. This change would allow for us to have these using the base carbon components

### "Must have" functionality

Having the flexibility of what we can render as a navigational component would allow us to stick with base carbon components and take advantage of built in accessibility and keyboard controls

### Available extra resources

What resources do you have to assist this effort?

I have issued a PR to make this change. https://github.com/IBM/carbon-components-react/pull/1980

_Carbon is a collaborative system. We encourage teams to build components and submit them for integration as either add-ons or core components._

PR #1980, opened 2019-03-06
[Feat] Add ability to render item other than string for HeaderMenu

davidicus:
Closes https://github.com/IBM/carbon-components-react/issues/1979

Currently HeaderMenu component only renders the string provided as aria-label. This PR allows the user to pass anything to render whether its an icon or text and an icon, etc.

#### Changelog

**Changed**

- added prop to HeaderMenu
- in render method check for content prop value. If one is available will render as menu item, otherwise it renders the aria-label
- updated the snapshot
- added test for keyboard interactions and render prop functionality


davidicus:
@asudoh Thanks for reviewing so quickly. I think this latest commit is what you are expecting. Please confirm.

src/components/UIShell/HeaderMenu.js
```diff
@@ -15,6 +15,13 @@ import { AriaLabelPropType } from '../../prop-types/AriaPropTypes';
 
 const { prefix } = settings;
 
+const defaultMenuContent = ({ ariaLabel }) => (
```
asudoh:
```suggestion
const defaultRenderMenuContent = ({ ariaLabel }) => (
```

src/components/UIShell/HeaderMenu.js
```diff
@@ -37,6 +44,15 @@ class HeaderMenu extends React.Component {
...
     tabIndex: PropTypes.number,
+
+    /**
+     * Optional component to render instead of string
+     */
+    renderMenuContent: PropTypes.func,
+  };
+
+  static defaultProps = {
+    renderMenuContent: defaultMenuContent,
```
asudoh:
```suggestion
    renderMenuContent: defaultRenderMenuContent,
```

Issue #1973, opened 2019-03-05
CopyButton component alt text


mekaufman:
Related issue: https://github.ibm.com/Bluemix/core-dev/issues/7928

## Title line template: CopyButton component in carbon-components-react alt text

We are using the CopyButton component in the resource list details pane. When a user hovers on the icon, the alt text is not displayed. However, when you inspect fully, you do see there is alt text on the svg. I'm not sure if it's just in the wrong place to be able to be displayed via hover. Also, I haven't tested using JAWS (screen reader), but if it's not read out, then we have an a11y violation. I'll try to test that and post a comment in the issue.

![image](https://user-images.githubusercontent.com/13648625/53834033-054b3e80-3f58-11e9-900a-8d4e2ce9f78b.png)

Can we get an update to the component to have alt text show on hover?

dakahn:
Possibly related to #1964

Issue #1964, opened 2019-03-04
FireFox: close icon tooltip inside modal component is not show 


jingxxue:
## Environment

> Operating system : WINDOWS 10

> Browser : FIREFOX 65.0.2

> Automated testing tool and ruleset

> Assistive technology used to verify

## Detailed description

> What version of the Carbon Design System are you using?

    "carbon-components": "^9.82.13",
    "carbon-components-react": "^6.99.1",
    "carbon-icons": "^7.0.7",

> What did you expect to happen?
when hoover on the close icon inside modal, Tooltip with "close modal" should be show
> What happened instead?
not tooltip show when hoover on the close icon
> What WCAG 2.1 checkpoint does the issue violate?

## Steps to reproduce the issue

1. Step one 
open a modal
2. Step two
hoover on the close icon on top right corner
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
> https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox

## Additional information

- Screenshots or code
- Notes
in the new version of firefox, Firefox ignore the tittle element inside the svg, if the svg element is inside a button element, suggest to move the title as a attribute of button.

jingxxue:
the codeSandBox test case:  https://codesandbox.io/s/526moz78ok

dakahn:
Looks like the aria-label is being set properly. Cross checked the bug on Chrome as well and yeah, there's no tool tip there either. Wondering if this is a recent regression -- did this work in the past?

jingxxue:
> Looks like the aria-label is being set properly. Cross checked the bug on Chrome as well and yeah, there's no tool tip there either. Wondering if this is a recent regression -- did this work in the past?

is working in previous version of FireFox, but not the new version

vpicone:
I believe the issue is the SVG doesn't take up the entirety of the button. So the outside edges of the button are clickable but do not show the SVG title tooltip.

In the test case, if you hover over the center of the icon, you'll see the tooltip show up.

dakahn:
Thats a great catch @vpicone!

vpicone:
Working on a PR now :)

jingxxue:
> I believe the issue is the SVG doesn't take up the entirety of the button. So the outside edges of the button are clickable but do not show the SVG title tooltip.
> 
> In the test case, if you hover over the center of the icon, you'll see the tooltip show up.

PR #1982, opened 2019-03-06
fix: Add titles to buttons that encapsulate icons

vpicone:
Closes IBM/carbon-components-react#1973 and closes IBM/carbon-components-react#1964

When contained in a button element, the title tag of an svg element won't show in some instances. Specifically, when a user hovers over the outside edges of the button. 

By giving the buttons the same title tag as the child svg icon, the tooltip will now pop up for the entire component.

### Alternative solutions

This seems like a common enough motif that it could warrant a new button component.

#### Changelog

**New**

- add title attribute to buttons that encapsulate icons


vpicone:
@joshblack The two issues this is for are in regards to the tooltip.

Issue #2018, opened 2019-03-15
Match notifications to experimental spec


emyarod:
Related

https://github.com/IBM/carbon-components/issues/1845

https://github.com/IBM/carbon-components/pull/1931

PR #2020, opened 2019-03-15
fix(Notification): match experimental spec

emyarod:
Closes IBM/carbon-components-react#2018

This PR incorporates feedback from the February 2019 notification component audit

depends on https://github.com/IBM/carbon-components/pull/2119 and https://github.com/IBM/carbon-elements/issues/437

#### Changelog

**New**

- support legacy `iconDescription` prop in experimental icons

**Changed**

- upgrade icon package and replace warning icon (IBM/carbon-elements#374)
- render icons for toast notifications (except for info notifications)
- refactor duplicate notification icon logic


emyarod:
looks like CI is unable to fetch the beta version of `@carbon/icons-react`

src/components/Notification/Notification.js
```diff
@@ -229,6 +228,44 @@ export class NotificationTextDetails extends Component {
   }
 }
 
+const useIcon = kindProp =>
```
joshblack: Maybe something like:

```js
const iconTypes = {
  error: componentsX ? ErrorFilled16 : iconErrorSolid,
  success: componentsX ? CheckmarkFilled16 : iconCheckmarkSolid,
  warning: componentsX ? WarningFilled16 : iconWarningSolid,
  info: componentsX ? null : iconInfoSolid,
};
```

Then later on:

```jsx
	          <Icon
              description={iconDescription}
              className={`${prefix}--inline-notification__icon`}
              aria-label="close"
              icon={iconTypes[kind]}
            />
```

Since the info seems to be static. Is the closure for closing over `componentsX`?

src/components/Notification/Notification.js
```diff
@@ -361,6 +390,7 @@ export class ToastNotification extends Component {
 
     return (
       <div {...other} role={role} kind={kind} className={classes}>
+        {NotificationIcon({ notificationType, kind, iconDescription })}
```
joshblack: Any reason for the eager evaluation? Could this be a component instead?

src/components/Notification/Notification.js
emyarod: @joshblack this pattern is a holdover from https://github.com/IBM/carbon-components-react/pull/888/files#diff-005a7fa10228412b5f153e9889723cf4R158

I'm alright with refactoring but would first like some input from @asudoh on the intention behind the original pattern

src/components/Notification/Notification.js
emyarod: no not without some test refactoring, which I'm willing to look into at a later time if that's alright

src/components/Notification/Notification.js
asudoh: `keyValueMap[key]` evaluation here seems merely because `useIcon` is referred by multiple places and I didn't feel like doing such evaluation in multiple places. But if you feel like doing the evaluation in multiple places it'd be fine.

src/components/Notification/Notification.js
asudoh: @emyarod `<NotificationIcon>` seems to work as a component already in your PR and thus I guess @joshblack is suggesting `<NotificationIcon notificationType={notificationType} kind={kind} iconDescription={iconDescription} />`?

src/components/Notification/Notification.js
```diff
@@ -486,34 +508,10 @@ export class InlineNotification extends Component {
...
-            />
-          );
-      }
-    };
-
     return (
       <div {...other} role={role} kind={kind} className={classes}>
         <div className={`${prefix}--inline-notification__details`}>
-          {NotificationIcon(kind)}
+          {NotificationIcon({ notificationType, kind, iconDescription })}
```
asudoh: And from here I see you turned `NotificationIcon()` function to a React component, without doing it on-purpose 😉

src/components/Notification/Notification.js
emyarod: yeah @asudoh  I was initially rendering the icon that way but our test suite was throwing errors which I didn't investigate closely yet

emyarod: yeah so as we discovered the node tree that Enzyme tests against changes when we render `<NotificationIcon {...props} />` instead of `NotificationIcon(props)`. I just went ahead and refactored the test suite to account for those changes

asudoh (approved):
LGTM 👍 - Thanks @emyarod!

shixiedesign:
Looks good! Only 1 small issue:
- [ ] Success/Warning/Failure icons here should be 20x20
Current: 
image
Should be: 
image

emyarod:
got it @shixiedesign, does that mean we should change it in vanilla as well? looks like it is 16x16 there currently

shixiedesign:
From what I saw here, looks like vanilla is already 20x20. https://the-carbon-components.netlify.com/?nav=notification

emyarod:
@shixiedesign ok I see, so I think we need to refactor the vanilla component. In the [vanilla HTML](https://github.com/IBM/carbon-components/blob/master/src/components/notification/notification.hbs#L14), we are using icon helpers for 16px icons in both toast and inline notifications, but we have some [CSS rules only on the inline notification icons](https://github.com/IBM/carbon-components/blob/master/src/components/notification/_inline-notification.scss#L170-L173) which upscale the icon to 20x20. This is fine because they are SVGs, but it might be less confusing to just directly use the 20x20 icon helpers in the markup and remove those style rules, especially since they are [different in the current toast notifications](https://github.com/IBM/carbon-components/blob/master/src/components/notification/_toast-notification.scss#L151-L154) when they should be the same

asudoh:
@emyarod Given this PR is of React one would it make sense to go ahead with 20x20 one? We can do a separate work to change vanilla to use 20x20 icon. Is it what you meant?

emyarod:
@asudoh yes that is what I meant, these changes can be independent of each other

emyarod:
actually @asudoh this does depend on the vanilla repo because we need to remove the CSS rules that are modifying the SVG size, tracking in https://github.com/IBM/carbon-components/issues/2118

chrisconnors-ibm:
https://github.com/IBM/carbon-elements/pull/438 adds the 20x20 assets for `error--filled.svg`

src/components/Notification/Notification.js
joshblack: @asudoh not sure what you mean by `keyValueMap[key]` evaluation. Looking at usage, it seems like calls are parameterized by a `kind` param like in: `useIcon(kind)`. The `useIcon` method itself seems to create a brand new object each time and then do the lookup. Seems like we could just statically refer to this and keep the same signature?

tw15egan (approved):
Looks good once icons are updated 👍

emyarod:
the holdover styles from old `carbon-components` rules are now removed

shixiedesign (approved):
Looks great! 🎉

src/components/Notification/Notification.js
asudoh: @joshblack I just wanted to double-check if repeating something like `iconTypes[kind]` is what the dev wants to do - Anyhow it's just a matter of dev's taste.

emyarod: I left it as a holdover from previous commits but I am alright with refactoring as well, since makes sense to have it as just an object now

PR #2021, opened 2019-03-15
fix(datepicker): pin flatpickr to 4.3 and add migration docs

### LINKED ISSUE #2011: Datepicker range calendar cannot be controlled with arrow keys
## Environment

> Operating system
macOS Mojave 10.14.3

> Browser
Firefox 65.0.1

## Detailed description

carbon-components-react 6.106.3

> What did you expect to happen?
After tabbing into the datepicker range inputs, I expect to navigate inside the calendar with the arrow keys. 

> What happened instead?
Instead, I found that when trying to use the arrow keys, there is some kind of focus state that appears around the calendar itself, but then the arrow keys don't do anything else.

> What WCAG 2.1 checkpoint does the issue violate?
Success Criterion 2.1.1 Keyboard: All functionality of the content is operable through a keyboard interface without requiring specific timings for individual keystrokes, except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints. 
https://www.w3.org/TR/WCAG21/#keyboard

1. Visit the Carbon-Components-React storybook
2. Navigate to the Datepicker range option
3. Use your tab key to focus onto the datepicker input
4. Attempt to use the arrow keys to navigate the popup calendar. 

I'm posting the issue here and not directly to Flatpickr because the issue doesn't arise until the Carbon implementation: the Flatpickr example handles keyboard accessibility well: (https://flatpickr.js.org/plugins/#rangeplugin-beta) I can see in the source code that Carbon adds a bit of extra functionality onto the Flatpickr dependency. I think the issue might be found there.

Note: I checked the experimental v10 component page here, too (https://next.carbondesignsystem.com/components/date-picker/code) and the issue exists there as well.

dakahn:
Closes #2011 

Pins Flatpickr to v4.3. Addresses accessibility and usability regressions introduced in 4.4 -- mainly, disabling left and right arrow input when `allowInput` is set to `true` in the Flatpickr config.

#### Changelog

- added some migration documentation as migrate.10.x.md
- add escape keyhandler that returns focus to input on keydown in the calendar


bakerTX:
Would this reintroduce the focus trapping behavior that Flatpickr resolved last week? https://github.com/flatpickr/flatpickr/releases/tag/v4.5.5

dakahn:
In my testing I didn't see any loss of functionality. @bakerTX seeing as how we're rolling it back to 4.3 -- any updates in 4.5 wouldn't be included.

dakahn:
@joshblack testing testing

seems like focus is lost when the user hits escape on the calendar instead of properly returning to the input. i'll add a keyhandler and amend the migration docs.

bakerTX:
@dakahn thankfully no focus trapping in the netlify preview build. The arrow key navigation is a little buggy on any usage *after* the first setting of the date value, ie: returning focus to an already-set field and trying to change it. But other than that, at least the arrow keys work again!

dakahn:
I think we get a lot from this PR _as is_, but to @bakerTX's comment about the navigation being "buggy" the keyboard needs a little more work. I'll move that work to another PR though

asudoh:
From digging the history - In addition to the tabbing discussion, another thing that is good to test is SSR: #897

dakahn:
@asudoh Great catch! Is there a way to patch SSR support in -- a check for the window object maybe? 🤔

asudoh:
Good question @dakahn! If you mean doing it in Flatpickr code you are correct. Something like before-change versions in below should be fixed:

* Code creating `flatpickr` global: https://github.com/flatpickr/flatpickr/commit/1882ea322e11bdfd5376bce6affb0d7d8562c38b
* Flatpickr's implementation of rAF polyfill: https://github.com/flatpickr/flatpickr/commit/9618cbd8e476a2db38523ca827a296c68afb6efe

asudoh:
Adding a note; Both issues I mentioned above happens in Flatpickr module initialization logic, instead of in Flatpickr constructor. And thus the issue is likely to happen regardless of if we instantiate Flatpickr or not. Hopefully our SSR test catches this.

dakahn:
I'm gonna shut up shop on this for now. I'll get something submitted to FP for v10-update

Issue #2023, opened 2019-03-15
Match link to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1263

https://github.com/IBM/carbon-components/issues/1801

https://github.com/IBM/carbon-components/issues/1900

https://github.com/IBM/carbon-components/issues/1995

https://github.com/IBM/carbon-components/pull/1802

PR #2024, opened 2019-03-15
feat(Link): add disabled prop

emyarod:
Closes IBM/carbon-components-react#2023

This PR adds support for disabled link classes to match the experimental spec


designertyler (approved):
Looks good!

asudoh:
Just want to double-check with @dakahn - Would removal of `href` is OK for `<a>`? Thanks!

dakahn:
Oh interesting. Yeah, an anchor tag without an href attribute is a _placeholder link_ -- which is valid. It will be removed from the tab order by default as well. 👍

dakahn (approved):
Looks cool

Issue #2028, opened 2019-03-15
hidden prop of Tabs component is ignored


scottdickerson:
## Detailed description

> Describe in detail the issue you're having.
I can't hide the Tab Content rendered by the Tabs component when the Tab isn't currently selected.  This causes problems because it forces all my child components to load even if the user isn't viewing them.

> Is this issue related to a specific component?
Tabs

> What did you expect to happen? What happened instead? What would you like to see changed?
I assumed that the "hidden" prop of the Tabs component would keep the child Tab content from rendering if it's not currently showing from the description of the prop?  Is that the intention of that property?

I checked the source of the Tabs component but it appears that the hidden property is ignored, and no matter what all the Tab Content components are rendered into the Dom with aria-hidden attached to them.

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
6.106.0

> Please create a reduced test case in CodeSandbox
https://codesandbox.io/s/nk30xz0p3m

Notice that tab2 content is actually in the DOM even though I pass the hidden prop

JSoet:
Just responding to this, after it was linked here: https://ibm-safer-payments.slack.com/archives/C2K6RFJ1G/p1554229626107000 https://ibm-safer-payments.slack.com/archives/C2K6RFJ1G/p1554229740107400

We've had a similar issue, that the components still technically existed and were just hidden, so we created a wrapper which only adds the react component to the tab if the tab is actually selected (may need to be generalized a bit before being used)

```
export function TabsWithRenderingWrapper(
  Components: Array<React.ReactElement<ITabbableComponent>>,
  selectedTab: number,
  handleTabSelection: (tabNumber: number) => void
) {
  const tabs = Components.map((element, index) => {
    return (
      <Tab
        key={element.props.tabLabel}
        label={element.props.tabLabel}
        tabIndex={index}>
        {selectedTab === index && element}
      </Tab>
    )
  })
  return (
    <Tabs selected={selectedTab} onSelectionChange={handleTabSelection}>
      {tabs}
    </Tabs>
  )
}
```

asudoh:
@JSoet To see what to do with this issue, would like to elaborate why you wanted to avoid rendering non-selected tab contents (panels)? Thanks!

JSoet:
[...]

PR #2029, opened 2019-03-15
fix(tabs): should respect hidden property and filter out tab content that's not selected

scottdickerson:
Closes IBM/carbon-components-react#2028

The hidden property was being ignored by the Tabs component.  I made a change to respect that property and actually hide the TabContent

#### Changelog

**Changed**

- Added a filter step to check the hidden property of Tabs and stop any non-selected Tab Content from rendering


asudoh:
Hi @scottdickerson thank you for reporting the issue and your attempt to fix it! It appears that [the UA stylesheet for `[hidden]`](https://html.spec.whatwg.org/multipage/rendering.html#hiddenCSS), which has a very weak selector priority by definition, is overridden by [our reset style](https://github.com/IBM/carbon-components/blob/v9.68.0/src/globals/scss/_css--reset.scss#L136-L139). That said, would you try fixing it in application-side CSS or `carbon-components` CSS? Thanks!

scottdickerson:
Actually from a component perspective we really want to skip the React component render for optimization so I can’t fix from just a css perspective

asudoh:
Thank you for elaborating @scottdickerson! If you want to avoid react-render would you do:

```javascript
(!avoidRender && (
  <Tabs>
    <Tab label="tab1">tab1 content</Tab>
    <Tab label="tab2">tab2 content</Tab>
  </Tabs>
}
```

scottdickerson:
hey @asudoh here's my usecase.  Inside the tab I have my tabContent component... And I typically don't want my tabContent to do anything unless it's actually visible (especially like fire expensive server calls, render components, etc).  However, currently the tabContent renders whether the tab is selected or not.  This forces the componentDidMount method of tabContent to fire (which would be ideally where I would put my loading behavior).  I assumed (maybe incorrectly) that the Tabs "hidden" prop was intended to hide the child content if the Tab wasn't currently selected.  But that's basically what I'm trying to accomplish.  Theres an internal slack going on about this discussion.  While it's certainly possible to pass these extra props through to the child content myself and then decide whether to load or not, it would be nicer if this kind of "lazy loading" was build into the Tabs proper through a property.

asudoh:
IIUC what @scottdickerson is talking about, `cDM` of `<TabContent>` is not something consuming application should use. In case `<TabContent>` being rendered regardless of selection really is the concern, we can provide custom rendering API for `<TabContent>`.

dakahn:
stalebot no! ⚡️ 

@scottdickerson 👋

asudoh:
@scottdickerson @dakahn My apologies for not updating here. The series of investigations to better support the use case (not rendering tab content unless it's select) has found that letting application decide what to do there would be the best solution. #2142 addresses that. I appreciate the contributions you have been making BTW!

scottdickerson:
this is the perfect fix thanks @asudoh

Issue #2026, opened 2019-03-15
Match select to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1958

https://github.com/IBM/carbon-components/pull/1978

PR #2030, opened 2019-03-15
fix(Select): match experimental spec

emyarod:
Closes IBM/carbon-components-react#2026

Closes https://github.com/IBM/carbon-components-react/issues/2037

This PR incorporates feedback from the February 2019 select component audit

Depends on https://github.com/IBM/carbon-components/pull/2114 and https://github.com/IBM/carbon-components/pull/2135


src/components/Select/Select.js
```diff
@@ -76,23 +74,51 @@ const Select = React.forwardRef(
...
+        </>
+      );
+    })();
+    return (
+      <div className={`${prefix}--form-item`}>
+        <div className={selectClasses}>
+          <label htmlFor={id} className={labelClasses}>
+            {labelText}
+          </label>
+          {componentsX && !inline && helper}
```
asudoh: The markup change seems to affect `v9`. Could you please double-check `v9` is not broken?

asudoh (commented):
LGTM basically, just one thing - Thanks @emyarod!

src/components/Select/Select.js
emyarod: it should be fixed now @asudoh

asudoh: OK thank you for adding that back! Actually `v9` topic came up as the helper used to be next icon whereas it's next to `<label>` in this PR. Wanted to make sure such move-around doesn't negatively affecting `v9`. Thanks!

asudoh: @emyarod You are right, I may have been seeing a stale commit, my apologies!

tw15egan (approved):
👍

emyarod:
currently in the middle of incorporating changes from IBM/carbon-components#2114 to add legacy support for inline helper text @tw15egan, will likely need another review in a bit

tw15egan:
Okay, just ping me when re-review is needed

emyarod:
@tw15egan https://github.com/IBM/carbon-components/pull/2135 is now ready for review

aagonzales (commented):
For disabled, the label and helper text need to be using the color "$disabled-02"

tw15egan:
@aagonzales seems like they are: https://github.com/emyarod/carbon-components/blob/8ef8e3ccd9eedfb1546cdf38b3b0b6b892d55e09/src/components/select/_select.scss#L287

aagonzales:
Oh ok the knob was only changing the field but I guess the labels are controlled by a different knob?

aagonzales (approved):
LGTM then!

emyarod:
@aagonzales can you double check again? I was still moving newly added changes from vanilla over when you reviewed, but disabling the component should also be changing label and helper text

aagonzales:
@emyarod yup! The knob is changing the label and helper text now so its good!

PR #2031, opened 2019-03-18
test(components): support v10 code

### LINKED ISSUE #2032: Test: Ensure test runs with feature flags flipped
### Summary/Justification

This issues is for tracking an effort to make sure our CI/test runs with feature flags (`componentsX`, `breakingChangesX`) flipped. We plan to ship `v10` with those feature flags flipped, and thus such effort ensures `v10` code passes the CI/test.

### "Must have" functionality

`yarn test` passes after manually editing feature flags.

Before we ship `v10`.

asudoh:
Fixes #2032.

#### Changelog

**New**

- CI script running tests with feature flags flipped.

**Changed**

- Tests to consider `v10` logic.
- Some misc changes to code under test, found during the test effort (details as comments to specific code).


src/components/Notification/Notification.js
```diff
@@ -135,7 +135,6 @@ export class NotificationButton extends Component {
           <IconTag
             aria-label={iconDescription}
             className={iconClasses}
-            icon={!icon && !name ? iconClose : icon}
```
asudoh: Found during the test, `icon` prop is for legacy `<Icon>` component and the code here is for new feature in `v10` supporting custom icon via render prop.

src/components/NumberInput/NumberInput.js
```diff
@@ -392,25 +392,21 @@ class NumberInput extends Component {
                       className={`${prefix}--number__control-btn up-icon`}
                       {...buttonProps}
                       onClick={evt => this.handleArrowClick(evt, 'up')}
-                      title={incrementNumLabel}
+                      title={incrementNumLabel || iconDescription}
```
asudoh: For getting in sync with `aria-label`.

src/components/NumberInput/NumberInput.js
```diff
@@ -392,25 +392,21 @@ class NumberInput extends Component {
...
-                        <title>{incrementNumLabel || iconDescription}</title>
-                      </CaretUpGlyph>
+                      <CaretUpGlyph className="up-icon" />
                     </button>
                     <button
                       className={`${prefix}--number__control-btn down-icon`}
                       {...buttonProps}
                       onClick={evt => this.handleArrowClick(evt, 'down')}
-                      title={decrementNumLabel}
+                      title={decrementNumLabel || iconDescription}
```
asudoh: For getting in sync with `aria-label`.

src/components/NumberInput/NumberInput.js
```diff
@@ -392,25 +392,21 @@ class NumberInput extends Component {
...
                       onClick={evt => this.handleArrowClick(evt, 'up')}
-                      title={incrementNumLabel}
+                      title={incrementNumLabel || iconDescription}
                       aria-label={incrementNumLabel || iconDescription}
                       aria-live="polite"
                       aria-atomic="true">
-                      <CaretUpGlyph className="up-icon">
-                        <title>{incrementNumLabel || iconDescription}</title>
-                      </CaretUpGlyph>
+                      <CaretUpGlyph className="up-icon" />
```
asudoh: Incorporating #1977.

src/components/NumberInput/NumberInput.js
```diff
@@ -392,25 +392,21 @@ class NumberInput extends Component {
...
                       onClick={evt => this.handleArrowClick(evt, 'down')}
-                      title={decrementNumLabel}
+                      title={decrementNumLabel || iconDescription}
                       aria-label={decrementNumLabel || iconDescription}
                       aria-live="polite"
                       aria-atomic="true">
-                      <CaretDownGlyph className="down-icon">
-                        <title>{decrementNumLabel || iconDescription}</title>
-                      </CaretDownGlyph>
+                      <CaretDownGlyph className="down-icon" />
```
asudoh: Incorporating #1977.

scripts/ci-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+yarn prettier:diff
+yarn lint
+yarn test --runInBand
+cp src/internal/FeatureFlags.js src/internal/FeatureFlags.js.orig
+sed -i -r 's/export *const *(components|breakingChanges)X *= *false/export const \1X = true/g' src/internal/FeatureFlags.js
```
emyarod:
```suggestion
sed -i -r 's/export *const *\(components|breakingChanges\)X *= *false/export const \1X = true/g' src/internal/FeatureFlags.js
```

I know the `-r` flag enables extended regex but no capture groups were being returned unless I escaped the parentheses when I tried running this. Not sure if I'm going down the right path with this correction but wanted to see if you also got errors when running the CI shell script locally

emyarod: should we also ignore the output `.js-r` file?

asudoh: @emyarod Have you tried `-E` instead of `-r`? From my research Mac and Linux seems to have different argument for regular expression. BTW the interim backup file will be erased before CI finishes.

emyarod: @asudoh I first tried with the `-E` flag before escaping the parentheses, but I received the same error: `sed: 1: "s/export *const *(comp ...": \1 not defined in the RE`

also thanks for clarifying on the backup file

asudoh: @emyarod I see Mac won't let me use `-i` without back-up file extension. Given it seems that you are reporting issues with Mac and I'm testing with CI environment (Linux), I changed `ci-check.sh` to support both.

emyarod (approved):
changes look good to me, thanks for adding macOS support for the CI script!

scripts/ci-check.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
joshblack: Instead of adding a custom shell script, should we add another setup file in `config/jest`? Could do a `config/jest/feature-flags.js` file:

```js

'use strict';

const FeatureFlags = require.requireActual('../../src/internal/FeatureFlags');

FeatureFlags.componentsX = true;
FeatureFlags.breakingChangesX = true;
```

And the task in yarn would look like:

```json
{
  "scripts": {
    "ci-check": "yarn prettier:diff && yarn lint && yarn test --runInBand && yarn test:experimental --runInBand && yarn test-ssr",
    "test": "jest",
    "test:experimental": "jest --setupFiles='<rootDir>/config/jest/setup.js' --setupFiles='<rootDir>/config/jest/feature-flags.js'",
  }
}
```

asudoh: Makes sense - Updated as such.

src/components/Select/Select.js
```diff
@@ -78,7 +78,9 @@ const Select = React.forwardRef(
             {children}
           </select>
           {componentsX ? (
-            <ChevronDownGlyph className={`${prefix}--select__arrow`}>
+            <ChevronDownGlyph
+              className={`${prefix}--select__arrow`}
+              aria-label={iconDescription}>
```
asudoh: @emyarod Guessing you actually wanted `aria-label` here, but please speak up if you had removed this deliberately. Thanks!

joshblack (approved):

tw15egan (approved):
👍 ✅

Issue #2033, opened 2019-03-18
Match slider to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1265

https://github.com/IBM/carbon-components/pull/1315

PR #2034, opened 2019-03-18
fix(Slider): add proper support for disabled state

emyarod:
Closes IBM/carbon-components-react#2033

This PR properly disables the slider when the disabled prop is applied

#### Changelog

**Changed**

- properly disable text input when component is disabled
- add disabled styles to component label


tw15egan (approved):
👍 ✅

tw15egan:
For @IBM/carbon-designers, since there is not a disabled story

Screen Shot 2019-03-19 at 12 11 19 PM

joshblack (approved):

aagonzales (approved):
Yup looks good!

Issue #2035, opened 2019-03-18
Match textarea to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1946

https://github.com/IBM/carbon-components/pull/1961

PR #2036, opened 2019-03-18
fix(TextArea): match experimental spec

emyarod:
Closes IBM/carbon-components-react#2035

This PR incorporates feedback from the March 2019 textarea component audit

depends on

https://github.com/IBM/carbon-components/pull/2151
https://github.com/IBM/carbon-components/pull/2153
https://github.com/IBM/carbon-components/pull/2158

#### Changelog

**New**

- disabled classes for form label and helper text
- experimental invalid state support

**Changed**

- refactor invalid textarea prop logic


tw15egan (approved):
👍

joshblack (approved):

laurenmrice (changes requested):
**Changes:**
~We need to implement the character count for text area. Is there a reason why we aren't right now? This has been something that is asked for a lot but other teams.~ will add as post v10 feature.
Screen Shot 2019-03-19 at 2 33 29 PM

- [x] For the disabled knob, the text needs to be `disabled-02`. This is happening in Firefox and Chrome.
Screen Shot 2019-03-19 at 2 13 27 PM

Validation knob
- [x] The error border is not equal on all sides. Make sure it is 2px all around. This is happening in Firefox and Safari.
Screen Shot 2019-03-19 at 2 36 20 PM

- [x] Double check to make sure we are pulling in the correct error icon (16/warning—filled) https://ibm.github.io/carbon-elements/icons/examples/esm/#16%2Fwarning--filled. The icon should also be placed 12px from the top and 16px from the left of the field.
- [x] Error message text should be calling `caption-01`
- [x] The spacing between the helper text and field should be 8px.

tw15egan:
Is adding the character count something that is needed to be added for v10? This seems like a feature enhancement and may be out of the scope of aligning vanilla / react styles. We would need to update the vanilla component again as well.

laurenmrice:
@tw15egan Yeah that is what I was wondering. We can add that as a post v10 thing for now!

tw15egan:
Great! Definitely something I want to add, but not a trivial addition and want to make sure we flesh it out for all use-cases

emyarod:
the error outline will be resolved in https://github.com/IBM/carbon-components/pull/2151

@laurenmrice just double checking, are the last 3 points found issues or just reminders? I believe they are already resolved since it directly inherits those from vanilla

laurenmrice:
@emyarod Sounds good about  error outline issue and counter. Yes the last three points were just to check on, but if they are correct then I will review the latest link.

emyarod:
@laurenmrice changes from https://github.com/IBM/carbon-components/pull/2151 are in, ready for another review!

laurenmrice (approved):
Looks good to me ! 👍

PR #2047, opened 2019-03-19
fix(Tooltip): support click-to-open with no-icon scenario

### LINKED ISSUE #2046: Click-to-open tooltip: Doesn't work without icon
## Detailed description

Click-to-open tooltip doesn't work when `showIcon={false}` is set.

`<Tooltip>`

Click-to-open tooltip should work even when `showIcon={false}` is set.

Chrome 73.

`9.105.0`

1. Open https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox
2. Replace the whole content of `index.js` with below:

```javascript
import React from 'react';
import { render } from 'react-dom';
import { Tooltip } from 'carbon-components-react';

const App = () => (
  <div style={{ marginTop: '2rem' }}>
    <Tooltip clickToOpen={true} showIcon={false} triggerText="Tooltip label">
      <p>
        This is some tooltip text. This box shows the maximum amount of text
        that should appear inside. If more room is needed please use a modal
        instead.
      </p>
      <div className="bx--tooltip__footer">
        <a href="/" className="bx--link">
          Learn More
        </a>
        <button
          className="bx--btn bx--btn--primary"
          type="button">
          Create
        </button>
      </div>
    </Tooltip>
  </div>
);

render(<App />, document.getElementById('root'));
```

3. Click on "Tooltip label"

BUG: Nothing happens. If you do `clickToOpen={false}` in `<Tooltip>` tooltip shows by hover-over.

asudoh:
Fixes #2046.

#### Changelog

**New**

- Added missing mouse/keyboard handlers for click-to-open `<Tooltip>` with no-icon scenario

**Changed**

- Knobs of various scenarios in click-to-open `<Tooltip>`, aligning to the typical one with icon.


joshblack:
Is this a supported use-case from design? Couldn't see an example in our usage tab

tw15egan:
cc @IBM/carbon-designers

asudoh:
@emyarod AFAIK definition tooltip is a hover-over tooltip, unless our designers want an alternate interaction method.

@joshblack Good question, the introduction of that option dates back to February 2017 (meaning it's before we started this repo) and I see our designer's sign-off to the contributor's PR (titled: `fix(Tooltip): Allow option to exclude info icon`).

shixiedesign:
Hi Akira, I noticed a few small things, though maybe some of these below are not intended to be fixed in this PR, and should be made into a new issue for Tooltip `v10-update`. Just let me know. I also couldn't find the definition tooltip 😬. We should still have a definition tooltip where hover triggers the tooltip to show. 

- [ ] Focus outline is incorrect. Should use $focus color token and same focus as link (I think...)
image
- [ ] Tooltip's link's hover underline color should be $link-01, same as link
image
- [ ] Tooltip's shark fin / triangle tip for Left, Top, and Right are all a couple pixels off 
image 
image
image

asudoh:
Thanks @shixiedesign for reviewing! Created #2061 for the visual issues. BTW definition tooltip can be found at `TooltipDefinition` in our storybook.

Issue #2016, opened 2019-03-14
Add support for current page in Breadcrumb


joshblack:
We added support for current page through `aria-current="page"` and a custom class over in carbon-components, we should sync up these changes over in this project.

PR #2054, opened 2019-03-19
feat(breadcrumb): add support for current page

joshblack:
Closes IBM/carbon-components-react#2016

Add support for current page in v10

#### Changelog

**New**

**Changed**

- Update `BreadcrumbItem` to be in `Breadcrumb` folder, maintain path for backwards-compat
- Update HTML for `Breadcrumb` behind `componentsX` flag
- Update logic for rendering in `BreadcrumbItem`
- Add `isCurrentPage` prop for `BreadcrumbItem`
- Add support for `aria-current="page"` for `BreadcrumbItem`
- Add story for `noTrailingSlash`
- Add stories behind flag for `isCurrentPage` and `aria-current="page"`
- Update test structure for breadcrumb

**Removed**


src/components/Breadcrumb/BreadcrumbItem.js
```diff
@@ -0,0 +1,78 @@
...
+  className: PropTypes.string,
+
+  /**
+   * Optional string representing the link location for the BreadcrumbItem
+   */
+  href: PropTypes.string,
+};
+
+if (componentsX) {
+  BreadcrumbItem.propTypes.children = PropTypes.oneOfType([
```
asudoh: `PropTypes.node`?

emyarod:
the deploy preview appears to be broken

image

src/components/Breadcrumb/Breadcrumb-story.js
```diff
@@ -42,10 +41,72 @@ storiesOf('Breadcrumb', module)
...
+  )
   .add('skeleton', () => <BreadcrumbSkeleton />, {
     info: {
       text: `
           Placeholder skeleton state to use when content is loading.
           `,
     },
   });
+
+if (FeatureFlags.componentsX) {
```
emyarod: I guess this breaks because of the way we are exporting our feature flags? `FeatureFlags` is undefined, so we need to directly import `componentsX`. other than that the changes look good to me

joshblack: Good call! Thanks @emyarod

src/components/Breadcrumb/BreadcrumbItem.js
joshblack: @asudoh good catch! Was going back and forth on this, essentially wanted to type it so that a single valid react element was the prop type, which I think element gets at (https://reactjs.org/docs/typechecking-with-proptypes.html#requiring-single-child)

shixiedesign:
Hi Josh, just 2 comments:
- [ ] Is current page supposed to be hover-able/clickable? Feels like it shouldn't since we are already on page. 
- [ ] If it is already current page, we should never have the `/` which indicates a level up on the hierarchy
image

joshblack:
@shixiedesign 

>  Is current page supposed to be hover-able/clickable? Feels like it shouldn't since we are already on page.

Since it's a link I don't think we should revert the tag or anything. One use-case could be, if wanting to share the link, to right click it and click "copy link address" in Chrome. But yeah, they most likely would not want to navigate to their current page.

>  If it is already current page, we should never have the / which indicates a level up on the hierarchy

This is behavior that's in v9 right now, mind making an issue to fix?

shixiedesign:
@joshblack Just discussed this with @aagonzales, we should remove the link behaviors from the current page breadcrumb. It's not spec'ed to be a link, and I worry that it will confuse users if it's clickable (user could be thinking: where does it take me? I thought I'm already here.) 

Made an issue for removing the `/` !

joshblack:
@shixiedesign the main dev issue here is that we use `aria-current="page"` to [tell screen readers that this link is the current page](https://www.youtube.com/watch?v=as2CZKc4Kx4) within a set of pages and it is only valid on a `<a>` tag.

While we don't have to follow WAI-ARIA examples, they're example also behaves in this way: https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

More info: https://tink.uk/using-the-aria-current-attribute/

shixiedesign:
@joshblack I see 🤔 so let me approve this so we can move forward. And let's see later if there's a good solution/alternative without compromising accessibility.

PR #2057, opened 2019-03-19
Audit: Pagination fixes (react)

lovemecomputer:
to be paired with https://github.com/IBM/carbon-components/pull/2136 for complete fixes!

note: there is one outstanding issue that is actually part of `Select`: once `Select` is updated & stable (and merged in), the icon here will automatically show the right one!


src/components/PaginationV2/PaginationV2.js
```diff
@@ -325,8 +326,9 @@ export default class PaginationV2 extends Component {
             <Select
               id={`${prefix}-pagination-select-${inputId + 2}`}
               className={`${prefix}--select__page-number`}
-              hideLabel
+              labelText={`Page number, of ${totalPages} pages`}
```
asudoh: g11n - Can we extract out this template as a prop (probably a form of function)? Thanks!

PR #2062, opened 2019-03-20
fix: deprecate the appendTo prop in date picker

vpicone:
Closes IBM/carbon-components-react#2068

This looks to be a relic of a one-to-one mapping with the flatpickr options. I find it really difficult to come up with a reason we'd want to traverse the dom for an arbitrary css selector to append to.

If it turns out folks really do need this behavior, a new feature that lets users pass a dom ref would be likely be a superior API choice I. Either way we'd still need to deprecate this imperitive css querying behavior.


asudoh:
Thanks you @vpicone for entering a quick fix! A couple of things:

1. We encourage folks to enter issue along with PR (probably except for internal minor chore)
2. Do you want to introduce a deprecation logic? (Searching for `componentsX` or `breakingX` will reveal some logic with `warning()` and you can do similar)

vpicone:
@asudoh Thank you for the feedback. I added a warning with recommendations and created an issue that better illustrated why this prop is so problematic.

src/components/DatePicker/DatePicker.js
```diff
@@ -251,10 +252,18 @@ export default class DatePicker extends Component {
...
-        typeof appendTo === 'string'
-          ? document.querySelector(appendTo)
-          : appendTo;
+
+      let appendToNode;
+      if (typeof appendTo === 'string' && !breakingChangesX) {
+        warning(
+          false,
+          `The DatePicker appendTo prop will be deprecated in Carbon X. 
+          - If you were appendTo this for styling, consider using a className.
```
emyarod: "using appendTo for ..."?

aledavila:
Ready to merge ? @vpicone

vpicone:
Yes!

vpicone:
@aledavila yes!

Issue #2063, opened 2019-03-20
[A11y] The Tooltip content can't be read by JAWS


icemanlily:
===== ENVIRONMENT ===== 
IBM CI 162 Checkpoint Violation: 1.3.1 Info and Relationships
IBM Guidance: https://www.ibm.com/able/guidelines/ci162/info_and_relationships.html
Build Level: Published - 02/27/2019
Test Tool Type: JAWS

===== DEFECT DESCRIPTION ===== 
The tooltip information cannot be read by screen reader, it only read as "tooltip button menu... ", for both "Select a resource group" and "Tags"

=====STEPS TO REPRODUCE ===== 

Login to IBM Cloud stage env
Navigate to Catalog -> Security and Identity -> Hyper Protect Crypto Service
Use Tab key to navigate to the tooltip icon

![image](https://user-images.githubusercontent.com/6970471/54661144-08c3f580-4b14-11e9-91ee-db914e6e74f8.png)

icemanlily:
Fixed and close this issue.

PR #2065, opened 2019-03-20
Fix tooltip a11y issue

icemanlily:
https://github.com/IBM/carbon-components-react/issues/2063


src/components/Tooltip/Tooltip.js
```diff
@@ -447,6 +447,11 @@ class Tooltip extends Component {
           'aria-owns': tooltipId,
         };
 
+    const ariaDescribedbyProps = !open
+      ? {}
+      : {
+          'aria-describedby	': tooltipId,
```
emyarod: just wondering is there a reason for the trailing space?

icemanlily:
@emyarod Removed the trailing space. Thank you!

src/components/Tooltip/Tooltip.js
icemanlily: Removed the trailing space. Thanks!

Issue #2012, opened 2019-03-13
Match button to experimental spec


emyarod:
Related:

IBM/carbon-components#1928

IBM/carbon-components#1935

https://github.com/IBM/carbon-components/pull/1930

https://github.com/IBM/carbon-components/pull/2087

https://github.com/IBM/carbon-components/pull/2106

PR #2067, opened 2019-03-20
fix(Button): add disabled state for other nodes

emyarod:
Closes IBM/carbon-components-react#2012

This PR incorporates feedback from the February 2019 button component audit

depends on IBM/carbon-components#2148 and https://github.com/IBM/carbon-components/pull/2161

#### Changelog

**Changed**

- apply disabled button styles to all button examples


tw15egan (approved):
🙌 ✅

laurenmrice (changes requested):
For the Button grouping:
Artboard

- [ ] When two buttons live together they both need to be the same size.
- [ ] When they are disabled, they kind of blend into each other, so I think adding a rule will help separate them.

emyarod:
button sets changes will come from https://github.com/IBM/carbon-components/pull/2161

laurenmrice (approved):
Looks good to me ! 👍

PR #2069, opened 2019-03-20
Pagination screen reader tweak

lovemecomputer:
setting new screen reader text after convo with @dakahn .

also removing some code smell of exposing the button text as a prop—should not be configurable.


lovemecomputer:
https://github.com/IBM/carbon-components/pull/2136 was merged while i was working on this, so i just made a new PR. files changed view still shows a good summary of the a11y tweaks here

src/components/PaginationV2/PaginationV2.js
```diff
@@ -139,9 +129,7 @@ export default class PaginationV2 extends Component {
   };
 
   static defaultProps = {
-    backwardText: 'Backward',
     itemRangeText: (min, max, total) => `${min}-${max} of ${total} items`,
-    forwardText: 'Forward',
```
tw15egan: Shouldn't we be setting the new text as default, not as an unchangeable hardcoded value? Thinking of internationalization

lovemecomputer: @tw15egan oh wow yes true!!!

src/components/PaginationV2/PaginationV2.js
```diff
@@ -338,32 +326,16 @@ export default class PaginationV2 extends Component {
           <button
             className={backButtonClasses}
             onClick={this.decrementPage}
-            aria-label={backwardText}
+            aria-label="previous page"
             disabled={backButtonDisabled}>
-            {componentsX ? (
-              <CaretLeft24 />
-            ) : (
-              <Icon
```
asudoh: `v9` need to keep the old icon?

lovemecomputer: @asudoh i think this whole code block is already inside of a `return componentsX ? ( … )` statement

lovemecomputer:
i updated the story to match the default props for better screen reader text!

src/components/PaginationV2/PaginationV2.js
asudoh: Got it - Thank you for clarifying @lovemecomputer!

Issue #2042, opened 2019-03-18
[Tabs] content need to be aligned to tabs type


shixiedesign:
Content inside tabs should start at the same position as tab's label text. This component should bleed into gutter so at any breakpoint, the content is 16px away from left and right edge of container.
Vanilla has this right except at 1 breakpoint: https://github.com/IBM/carbon-components/issues/2126
![image](https://user-images.githubusercontent.com/15144993/54563539-6ed73e00-4997-11e9-99ca-3feef4e32cef.png)

PR #2077, opened 2019-03-21
chore(tabs): add padding in tabs to content story

aledavila:
Closes IBM/carbon-components-react#2042

#### Changelog

**Changed**

- add padding to the content placeholder in storybook


lovemecomputer:
the padding is being added to the story? not sure if i understand why the padding isn’t being added to the component

PR #2079, opened 2019-03-21
fix: apply correct aria label prop for radio

nelsonchen90:
apply correct aria label prop for radio select on data table. resolving the dev warning about `Invalid ARIA attribute`

#### Changelog

**Changed**
- apply correct aria label prop for radio select on data table.

### DIFF
--- a/src/components/DataTable/TableSelectRow.js
+++ b/src/components/DataTable/TableSelectRow.js
@@ -25,13 +25,16 @@ const TableSelectRow = ({
     name,
     onClick: onSelect,
     checked,
-    ariaLabel,
     disabled,
   };
   const InlineInputComponent = radio ? RadioButton : InlineCheckbox;
   return (
     <td className={className}>
-      <InlineInputComponent {...selectionInputProps} />
+      <InlineInputComponent
+        {...selectionInputProps}
+        {...radio && { labelText: ariaLabel }}
+        {...!radio && { ariaLabel }}
+      />
     </td>
   );
 };
--- a/src/components/DataTable/TableSelectRow.js
+++ b/src/components/DataTable/TableSelectRow.js
@@ -32,7 +32,10 @@ const TableSelectRow = ({
     <td className={className}>
       <InlineInputComponent
         {...selectionInputProps}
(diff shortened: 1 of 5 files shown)


nelsonchen90:
error on dev env:
image

asudoh:
Also @nelsonchen90 do you want to take care of the snapshot? Thanks!

nelsonchen90:
@asudoh @emyarod thanks for the review,
yes I will take care of the snapshots!

nelsonchen90:
need more work, what's happening now is:
image

the 'Select row' is not expected

nelsonchen90:
@emyarod @asudoh 
I need your input here:
in RadioButton, we are mixing the `aria-label` with the text on the `<label>`:
https://github.com/IBM/carbon-components-react/blob/master/src/components/RadioButton/RadioButton.js#L125-L131

should I keep it as is? see the screenshot in the comment above

emyarod:
@nelsonchen90 I guess we can either pass in the value of `ariaLabel` as hyphen-case `aria-label` to RadioButton, or we can also add `hideLabel` as you suggested. Adding `hideLabel` would be consistent with other components like checkbox currently, and eventually we can refactor our components to avoid camelCase `ariaLabel` altogether and just use hyphen-case `aria-label` directly (related https://github.com/IBM/carbon-components-react/issues/1662). So for now I am leaning towards adding `hideLabel`, but would be interested in hearing others' thoughts on this

asudoh:
👍 for `hideLabel` - @nelsonchen90 Seems that snapshot update is still needed, could you take care of this? Thanks!

nelsonchen90:
@asudoh
sorry for late response!
yes I will add the hideLabel and then refresh the snapshot

nelsonchen90:
@asudoh @emyarod 
I updated the PR with the `hideLabel` and fixes the test snapshot

emyarod:
@nelsonchen90 it looks like there is a merge conflict, would you mind updating your snapshots?

nelsonchen90:
@emyarod done.

emyarod:
thanks @nelsonchen90 !

PR #2083, opened 2019-03-22
feat(DataTable): Add sorting params to header onClick callback

moczolaszlo:
With this change the callback of DataTable header's onClick gives back the params of sorting as an object.
Like:
```
{
  sortHeaderKey: 'name',
  sortDirection: 'DESC,
}
```

**New**

- Sorting params added to the callback function.


emyarod:
hi @moczolaszlo it looks like CI is failing because the bundle size threshold needs to be increased https://github.com/IBM/carbon-components-react/blob/master/package.json#L365

on a side note, is there an associated issue for this pull request for additional context?

moczolaszlo:
Hi @emyarod 
I updated the bundle size threshold, and I rebased the PR.
We need this parameters, because we want to sort a lot of data with paging.

src/components/DataTable/DataTable.js
```diff
@@ -173,10 +173,31 @@ export default class DataTable extends React.Component {
       isSortHeader: sortHeaderKey === header.key,
       // Compose the event handlers so we don't overwrite a consumer's `onClick`
       // handler
-      onClick: composeEventHandlers([this.handleSortBy(header.key), onClick]),
+      onClick: composeEventHandlers([
+        this.handleSortBy(header.key),
+        onClick
+          ? this.handleOnHeaderClick(onClick, header.key, sortDirection)
```
asudoh: Do you want to gather sortkey/sortdirection in an object here, too?

moczolaszlo: Like this?
```
this.handleOnHeaderClick(onClick, {
  sortHeaderKey: header.key,
  sortDirection,
});```

Comment (asudoh) [reply src/components/DataTable/DataTable.js]: Yes 👍

**Comment by akmalmzamri** [comment]
Hi @moczolaszlo and @asudoh. I'm pretty new to React and Carbon. I'm trying to get both header's key and sort direction for my custom sort function. I have this `onClick` function: 

```
<TableHeader
    {...getHeaderProps({header})}
    onClick={(e, sortParams) =>
        console.log(e, sortParams)
    }>
```

However, it seems that `sortParams` is `undefined` in the `onClick` function. Am I doing this wrong?

Issue #2078, opened 2019-03-21
[Carbon X] ProgressIndicatorSkeleton circle svg looks strange


jendowns:
## Detailed description

At first glance this doesn't look correct but let me know if I'm mistaken ---

In the latest Firefox, the `ProgressIndicatorSkeleton` for Carbon X has a strange circle shape:

![Screen Shot 2019-03-21 at 2 51 15 PM](https://user-images.githubusercontent.com/9057921/54780941-0da6a900-4be9-11e9-81f8-f7fbdfab1d01.png)

http://react-experimental.carbondesignsystem.com/?selectedKind=Progress%20Indicator&selectedStory=skeleton&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

Normally I would expect this circle shape to be a full circle.

PR #2084, opened 2019-03-22
fix(progressIndicatorSkeleton): Update html to fix styling

alisonjoseph:
Closes IBM/carbon-components-react#2078

Updates HTML for experimental progress indicator skeleton state to fix circle not displaying as a circle.


emyarod (approved):
looks good to me. should we be applying a fill color to the svg?

alisonjoseph:
Probably, but we don't have any updated skeleton state designs for v10 afaik.

shixiedesign (approved):

PR #2090, opened 2019-03-25
Multi select v9 selection behaviour

ConradSchmidt:
Closes https://github.com/IBM/carbon-components-react/issues/2091
These 2 commits should fix 2 things:
1. The multi select v9 should behave like the v10, by not moving selected items to top
2. The multi select v9 should move selected items to top, when the dropdown is closed (and reopened).

#### Changelog

**New**

- none

**Changed**

- Multi-select dropdown (+filtered): Selected items don't move to top
- Multi-select dropdown (+filtered): Selected items move to top after the closing the dropdown

**Removed**

- none


asudoh (commented):
HI 👋 thank you for jumping in! Seems that this changes the behavior for `v9`, which is a UX breaking change, though I can see your belief that the new behavior you proposed is better. We need to provide a option (and not change the default for `v9`) at least to go forward with this. Thanks!

ConradSchmidt:
@asudoh I added an attribute `selectionFeedback` with the values `fixed`, `top` and `mixed`. For v9 I set the default to 'top' and for v10 to `fixed`. Our suggestion would be `mixed`.

ConradSchmidt:
I'm not sure what to fix there - I ran `yarn prettier:diff` it fails, but it says not what exactly is broken... any advice?

emyarod:
just need to update your branch with the latest `master` to incorporate changes from https://github.com/IBM/carbon-components-react/pull/2099

asudoh (commented):
Thank you for your update @ConradSchmidt - One request - Can we rename `mixed` to something like `top-after-reopen`?

src/components/MultiSelect/MultiSelect.js
```diff
@@ -251,7 +267,16 @@ export default class MultiSelect extends React.Component {
                 {isOpen && (
                   <ListBox.Menu>
                     {componentsX
```
asudoh: @emyarod Are `v9`/`v10` code here identical besides sorting? Of so, should the code branch here be based on the "sorting mode"?

ConradSchmidt: I removed the branching, because the 2 branches are completely identical after extracting the sorting

ConradSchmidt:
Renamed the selection feedback `mixed` to `top-after-reopen`

ConradSchmidt:
The CI found a failed test in the sidebar, which I didn't change. How can this be fixed? Seems like that the snapshot file is dated/wasn't checked in properly...
```
 FAIL  src/components/UIShell/__tests__/SideNavMenu-test.js
  ● SideNavMenu › should render

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "SideNavMenu should render 1".
```

emyarod (approved):
thanks for tackling this @ConradSchmidt! these changes look good to me after changing the default sorting behavior for v10. and just as a heads up we most likely will hold off on merging until the first release of v10 is out

src/components/MultiSelect/MultiSelect.js
emyarod: yes @asudoh the code is identical besides sorting at this point. https://github.com/IBM/carbon-components-react/issues/2109 will introduce some v10 only changes but it is still WIP (https://github.com/emyarod/carbon-components-react/blob/2109-match-multiselect-to-experimental-spec/src/components/MultiSelect/MultiSelect.js) and shouldn't conflict with the sorting behavior we are adding in this PR

src/components/MultiSelect/MultiSelect.js
```diff
@@ -128,13 +136,15 @@ export default class MultiSelect extends React.Component {
     light: false,
     title: false,
     open: false,
+    selectionFeedback: componentsX ? 'fixed' : 'top',
```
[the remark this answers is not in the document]
ConradSchmidt: done :)

emyarod:
@ConradSchmidt thanks for being patient on this, would you mind resolving the new merge conflicts?

@asudoh I think we should merge this PR in before #2239 since it affects the component behavior, what do you think?

asudoh:
I think either way is fine - I see those two are independent. But @emyarod do you have any more specific reasons that your work depends on this PR here?

emyarod:
no just to make resolving merge conflict a little easier @asusoh that's all

emyarod:
@ConradSchmidt I went ahead and resolved the merge conflict myself, please let me know if you have any feedback regarding the merge

@asudoh this is now ready for review, and I think we should merge it in before #2239

asudoh (approved):
OK looks good to go - Thanks @ConradSchmidt!

Issue #2022, opened 2019-03-15
Match dropdown to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/1955

https://github.com/IBM/carbon-components/issues/1648

https://github.com/IBM/carbon-components/pull/1994

Depends on https://github.com/IBM/carbon-components/pull/2181

laurenmrice:
*Changes:*

- [ ] If the dropdown is scrollable could we show the drawer with the fourth option text cut off? I feel like this will better show that the dropdown is actually scrollable instead of showing an end rule after Option 3.
<img width="414" alt="Screen Shot 2019-03-19 at 1 36 28 PM" src="https://user-images.githubusercontent.com/43969356/54632534-51b07700-4a4c-11e9-8b08-b9ce6779d6f3.png">

- [ ] The top dropdown label is getting a hover color when selecting other options. Should not be getting hover.
<img width="350" alt="Screen Shot 2019-03-19 at 1 36 45 PM" src="https://user-images.githubusercontent.com/43969356/54632621-87edf680-4a4c-11e9-9956-ac6792f12eab.png">

- [ ] Options in drawer all still in Helvetica typeface. Needs to be Plex.
![Screen Shot 2019-03-19 at 1 42 04 PM](https://user-images.githubusercontent.com/43969356/54632773-d602fa00-4a4c-11e9-9f14-c23a1e3936a3.png)

- [ ] After selecting an option, the drawer closes but the bottom rule color should change back to ui-04 (#8C8C8C). Once you click onto the actual page it does change color again, but it should change right after the drawer is closed.
![Screen Shot 2019-03-19 at 1 44 03 PM](https://user-images.githubusercontent.com/43969356/54632936-1f534980-4a4d-11e9-8edb-59f904f1bf86.png)

- [ ] In Safari I am getting some weird glitches in the drawer. 
<img width="425" alt="Screen Shot 2019-03-19 at 1 45 46 PM" src="https://user-images.githubusercontent.com/43969356/54633079-6b9e8980-4a4d-11e9-82e1-d22e3692744a.png">

- [ ] The focus needs to cover the bottom rule. Right now it is showing on top.
<img width="1054" alt="Screen Shot 2019-03-19 at 1 46 51 PM" src="https://user-images.githubusercontent.com/43969356/54633142-9092fc80-4a4d-11e9-8a2b-59cee3667a38.png">

PR #2093, opened 2019-03-25
fix(DropdownV2): match experimental spec

emyarod:
Closes IBM/carbon-components-react#2022

depends on
https://github.com/IBM/carbon-components/pull/2203
https://github.com/IBM/carbon-components/pull/2206/
https://github.com/IBM/carbon-components/pull/2212

This PR incorporates feedback from the March 2019 dropdown component audit

#### Changelog

**Changed**

- Integrate v10 listbox styles

**Removed**

- `type` prop for inline listbox components (requires refactoring of multiselect and combobox as well)


emyarod:
https://github.com/IBM/carbon-components/pull/2203 only addressed the overflow issue for the trigger button label 🤦‍♂️ 

* need overflow styles on menu options

* inline list box menu options also need reduced side margins to match dropdown spec

image

laurenmrice (changes requested):
- [x] Just one change: The top rule line in the drawer is flush next to the focus state. It
Screen Shot 2019-03-26 at 3 25 10 PM
 needs to not be there when the focus border is active.

emyarod:
tests are failing because of improperly formatted issue templates https://github.com/IBM/carbon-components-react/issues/2098

blocked until https://github.com/IBM/carbon-components-react/pull/2099 is merged in

laurenmrice (changes requested):
**Changes:**

- [x] The disabled state shouldnt have a bottom rule .
Screen Shot 2019-03-28 at 10 13 13 AM

- [x] For the light variant there should call field-02 because it is the inverse.
Screen Shot 2019-03-28 at 10 26 55 AM

- [ ] I don't think the focus border should be triggered persistently upon selection of the label when you open the drawer. This is happening in Firefox, Safari, and Chrome and was not happening before.
Screen Shot 2019-03-28 at 10 28 33 AM

emyarod:
regarding the last point, this seems to be the default behavior of Downshift, which we depend on for all of our listbox components. I opened https://github.com/IBM/carbon-components-react/issues/2111 to start the discussion about this

laurenmrice (approved):
Looks good to me ! 👍

src/components/ListBox/ListBoxMenuItem.js
```diff
@@ -25,7 +26,16 @@ const ListBoxMenuItem = ({ children, isActive, isHighlighted, ...rest }) => {
   });
   return (
     <div className={className} {...rest}>
-      {children}
+      {componentsX ? (
+        <div
+          className={`${prefix}--list-box__menu-item__option`}
```
joshblack: Do we need keyboard handlers if this is `role="button"`? Or potentially another interactive prop like `onClick`?

emyarod: I believe that behavior is already handled by Downshift, these attributes are just holdovers from vanilla and can probably be removed altogether

src/components/DropdownV2/DropdownV2.js
```diff
@@ -130,87 +147,115 @@ export default class DropdownV2 extends React.Component {
...
-                ))}
-              </ListBox.Menu>
-            )}
-          </ListBox>
-        )}
-      </Downshift>
-    );
-    return title || helper ? (
-      <>
+      <div className={wrapperClasses}>
```
asudoh: Given the change here - Wanted to see it's tested in `v9`. Thanks!

emyarod: I have tested it in v9 but I will rope it off as we don't have a wrapper in v9

joshblack (approved):

Issue #2098, opened 2019-03-27
GitHub issue template formatting breaking CI


emyarod:
some unformatted changes were made to our GitHub issue templates which is causing build failures

output from CI:

```sh
$ #!/bin/bash -eo pipefail
  yarn ci-check

yarn run v1.9.4
$ yarn prettier:diff && yarn lint && yarn test --runInBand && yarn test:experimental -u --runInBand && yarn test-ssr
$ prettier --list-different "**/*.{scss,css,js,md}"
.github/ISSUE_TEMPLATE/accessibility-issue--.md
.github/ISSUE_TEMPLATE/bug-report.md
.github/ISSUE_TEMPLATE/feature-request.md
.github/ISSUE_TEMPLATE/question.md
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Exited with code 1
```

PR #2099, opened 2019-03-27
chore: format GitHub issue templates and update snapshots

emyarod:
Closes IBM/carbon-components-react#2098

Our issue templates were not properly formatted and caused all CI builds to fail. This PR formats them to pass CI


vpicone:
We should limit commits directly to master or alternatively loosen CI restrictions for MD.

Issue #2107, opened 2019-03-28
New story that shows how a DataTable can be nested into another DataTable


marcosimioni:
### Summary

It would be nice to have a new story that shows how a DataTable can be nested into another DataTable.

### Justification

We already have a nice story that shows how to use regular Table nested into an expandable row of another regular Table (see http://react.carbondesignsystem.com/?selectedKind=Table&selectedStory=Nested%20Table&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel).
It would be nice to support nested DataTables as well and have a story demonstrating its usage.

### Desired UX and success metrics

It should be possible to nest a DataTable inside the expandable row of another DataTable.

### "Must have" functionality

Styles shouldn't break and it should be possible to have all the functionality of the nested DataTable working as expected, as is if it was a standalone DataTable.

### Specific timeline issues / requests

No timeline! I have the feeling it should come out of the box with what we have, with a couple of minor styles adjustment such as IBM/carbon-components#2006

### Available extra resources

See this code pen https://codepen.io/marcosimioni/pen/bZqKxN
I also have a story ready to be committed, I will attach it here as a PR.

joshblack:
Hi @marcosimioni! 👋 Thanks for taking the time to make this issue, unfortunately nesting data tables is not a use-case we support. Going to close this for now, but thank you so much for your time here 🙏

marcosimioni:
👍 Np! I totally understand! thanks for your comment!

Just to make sure @joshblack - based on the fact that we have a storybook story for nested Tables (http://react.carbondesignsystem.com/?selectedKind=Table&selectedStory=Nested%20Table&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel), does that mean this use case for regular Tables is supported?

I'd rather stay within the "boundaries" of the supported features, so .... if nested Tables is supported, then I will switch from DataTables to regular Tables indeed. For that I'm doing, they should be enough.

joshblack:
So the `Table` component shown currently was deprecated in v9, `Table` will instead point to `DataTable.Table` in v10 👍 Unfortunately this means that that use-case is going away in our next release 😞

PR #2108, opened 2019-03-28
docs(DataTable): Add nested DataTable story (#2107)

marcosimioni:
#### Overview

Closes IBM/carbon-components-react#2107

Add nested DataTable story to show how one can use DataTables nested into an expandable row of another DataTable.

#### Changelog

**New**

- New story for DataTable showing how to use nested Data Tables

### DIFF
--- a/src/components/DataTable/DataTable-story.js
+++ b/src/components/DataTable/DataTable-story.js
@@ -153,4 +153,11 @@ storiesOf('DataTable', module)
         `,
       },
     }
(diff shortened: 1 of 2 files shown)


joshblack:
Hi @marcosimioni! 👋 Most likely we won't merge in stories showing nested data tables as it's not a pattern we support 👍 Thanks so much for taking the time to put this PR together!!

PR #2115, opened 2019-03-28
chore(pagination): use en dash instead of hyphen

lovemecomputer:
replaces hyphen (`-`) with en dash(`–`)

closes https://github.com/IBM/carbon-components/pull/2231 (carbon components)


lovemecomputer:
just pushed a commit which updates the test content with en dashes, hopefully that passes and this PR will be done!

Issue #1878, opened 2019-02-14
DropdownV2 has multiple tabbable element


nicoxxie:
## Environment

> Operating system
MACOSX

> Browser
Chrome

> Automated testing tool and ruleset
DAP

> Assistive technology used to verify

## Detailed description

> What version of the Carbon Design System are you using?
4.0.10

> What did you expect to happen?
Only one tabbable element on the main dom

> What happened instead?
two tabblable element

> What WCAG 2.1 checkpoint does the issue violate?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
> https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox
Directly viewable on http://react.carbondesignsystem.com/?selectedKind=DropdownV2&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel
## Additional information

- Screenshots or code
- Notes

![81d65928-f70d-11e8-896d-708051638e54](https://user-images.githubusercontent.com/17915105/52810641-6066df80-30d6-11e9-817a-eaeb591b8b9f.png)
![f7c2486a-2087-11e9-9ccb-caba4c2bdb06](https://user-images.githubusercontent.com/17915105/52810643-6066df80-30d6-11e9-9fea-442c48c780ed.png)

dakahn:
Too right! Thanks for submitting the issue. Currently working on this error rn. I'll link this issue in the subsequent PR. 👍🏽

dakahn:
Doing a round of testing and it looks like this issue is resolved. 👍 Happy to reopen if I need to though.

nicoxxie:
@dakahn I'm still seeing two tabbale item for Dropdown.  Think this should be removed from the listbox?

![Screen Shot 2019-03-27 at 3 22 41 PM](https://user-images.githubusercontent.com/17915105/55109726-7a67ea80-50a4-11e9-9884-ca1cbcba24e1.png)

dakahn:
Mega weird DAP doesn't catch that anymore. I'll push up a fix right quick 👍🏽

emyarod:
@nicoxxie can you verify if this issue still exists? I think I resolved it in https://github.com/IBM/carbon-components-react/pull/2158

eason123deng:
[...]

PR #2116, opened 2019-03-29
fix(Dropdownv2): remove extra tab stop

dakahn:
Closes IBM/carbon-components-react#1878

Gets rid of the extra tab stop in the Listbox container in Dropdownv2

#### Changelog

- adds an `isDropDown` prop on Listbox 
- adds a check if Listbox is used in a dropdown -- if it is removes Listbox from the tab order but otherwise handles tabindex as before


dakahn:
Oh sick -- i'll link your PR in the issue I was addressing and we'll go with what your stuff. 👍🏽

Issue #2126, opened 2019-03-29
DataTable: update to v10


vpicone:
Tasks:

- Update classNames
- Update stories
- Description
- Modern icons
- Expandable rows

PR #2127, opened 2019-03-29
fix(DataTable): update data-table styles and add description prop

vpicone:
Closes IBM/carbon-components-react#2126

- [X] Update classNames
- [X] Update stories
- [X] Description/title
- [x] Modern icons
- [x] Expandable rows


asudoh:
Hi @vpicone looking at the to-do list in the description - Does this mean this PR is still WIP? Thanks!

jnm2377:
Looking good so far! Thanks for doing this 🎉

jnm2377:
Looks great! I think you forgot to add `sortable: true` to the default table story bc it doesn't look like it got the sortable table class. Other than it looks good. 

I did notice some bugs semi-related but not caused by this PR. It looks like I didn't add a class for _sortable_ `td:first-of-type` to change the padding from the td to the btn instead. So the hover looks weird rn. That's on me, didn't even occur to me until seeing this example. Adding it to my to do list, but just wanted to document it here for other reviewers.

Screen Shot 2019-04-01 at 4 37 40 PM

Also, this is an existing bug it looks like, but for the expandable table in vanilla js, when you hover on the child row it adds a class to the parent row so that it'll receive the same hover styles (to appear like they're one row rather than 2). I noticed v9 wasn't doing that, but maybe we should update that?

Screen Shot 2019-04-01 at 4 38 27 PM

vpicone:
@jnm2377 sorted thank you!

vpicone:
First react hook in the codebase! 👀

src/components/DataTable/Table.js
```diff
@@ -48,6 +50,11 @@ Table.propTypes = {
    */
   short: PropTypes.bool,
 
+  /**
+   * `false` Applies styles for data tables with sorting functionality.
+   */
+  sortable: PropTypes.bool,
```
joshblack:
```suggestion
  isSortable: PropTypes.bool,
```

To match [`TableHeader`](https://github.com/IBM/carbon-components-react/blob/master/src/components/DataTable/TableHeader.js#L123)

vpicone: 👍

vpicone:
Need to take care of the toolbar stuff

jnm2377:
@vpicone do you wanna do the toolbar stuff in a different PR so this can get merged in? Will also just be easier for review. We've been sectioning out the data table PRs in vanilla for that reason.

vpicone:
@jnm2377 Sounds good to me!

src/components/DataTable/TableExpandedRow.js
```diff
@@ -7,20 +7,48 @@
...
   children,
+  colSpan,
   ...rest
 }) => {
-  const className = cx(`${prefix}--expandable-row-v2`, customClassName);
+  const rowRef = useRef(null);
+  const className = cx(`${prefix}--expandable-row`, customClassName);
+
+  const onMouseEnter = () => {
+    if (rowRef && rowRef.current) {
```
asudoh: Any reason why you did direct DOM access instead of state update then render?

vpicone: I tried a couple options, I found this one the easiest to reason about and simplest. 

Since the class is toggled on a sibling dom node, state would have to be pulled up to the common parent component. 

You’d then have to do introspection on the children to find the hovered element and it’s sibling. This would also  force a rerender of the whole table every time a new row is hovered. 

My thought was that until we need to perform other actions based on the hover state, introducing new state just to track the hover event unnecessarily complicated the component. 

Another alternative I considered was introducing a wrapper for the Row and Expanded Row pairs but the nesting structure of our table styles and the current amount of table components made this seem unwise for just this one usecase.

asudoh: Thanks @vpicone for elaborating what's behind this approach @vpicone! - Which made me realize that the code here changes DOM of other component, assuming such other components exists at a certain location in DOM relative to this component. We need to make our codebase robust considering future change in such DOM location - AFAIK this is one of the biggest reasons React preaches for avoiding direct DOM access wherever possible.

Making the fragment containing `<TableExpandRow>` and `<TableExpandedRow>` a component looked a most straightforward option, though we need support for event handlers on fragments (can be seen at issue# 12051 in React repo).

If we still want to go with direct DOM access approach, I'd recommend at walking up through the siblings to find one matching `.${prefix}--expandable-row` selector.

vpicone: the `TableExpandedRow` has many siblings with the `--parent-row` selector. Many of those parent rows might also have `--expanded-row` (meaning they are currently expanded). There could also be hundreds of rows each with an adjacent expanded rows and this gets fired onHover.

Querying for a specific css selector seems brittle in a different way. I could be wrong here, but I believe refactoring a className shouldn't break the behavior of the component. I think my assumption that the TableExpandedRow has a sibling parent row may be more robust than it having a that sibling and that sibling also containing a specific className.

I mentioned a wrapper component, maybe `ExpandableRow`, but I agree that it would make an already difficult API even more so.

I could add a check to ensure that the previousSibling is a `--parent-row` but that gets back to relying on classNames for behavior.

asudoh: @vpicone Change in class name is a great point. Another possible approach is marking parent row `<tr>` with an (`data-`) attribute and searching for that with an attribute selector. BTW when I talk about walking up through the siblings, I mean using the first one found, not about finding all matching. DOM element has `.matches(selector)` API where you can see an element matches the given selector or not.

vpicone: How would we mark the parent row without already having a mark on the parent row? Finding it to mark it has the same issues as finding it to style it.

We could have a required key prop but at that point we minus well just have a wrapper component.

asudoh: I was going to talk about just adding `data-` attribute on the parent row `<tr>`, but seems that we have one already: https://github.com/IBM/carbon-components-react/blob/3d6ec01297638d9a1d3b2ae6e537074b779defa8/src/components/DataTable/TableExpandRow.js#L39

vpicone: So do you think adding a check for that custom data attribute on the previousSibling is necessary/less brittle?

asudoh: Yeah - And again we should walk through the siblings until we find one.

src/components/DataTable/TableExpandedRow.js
```diff
@@ -22,17 +22,9 @@ const TableExpandedRow = ({
...
-      );
-    }
-  };
-
-  const onMouseLeave = () => {
-    if (rowRef && rowRef.current) {
-      rowRef.current.previousSibling.classList.remove(
+  const toggleParentHoverClass = () => {
+    if (rowRef && rowRef.current && rowRef.current.previousElementSibling) {
+      rowRef.current.previousElementSibling.classList.toggle(
```
asudoh: The attempt to converge two event listener is good! - Instead of using to args `.toggle()`, we can use three-args `.toggle()` with `evt.type === 'mouseenter'` for the 3rd arg, to make it explicit that we are adding the class on hover.

vpicone: I believe that argument is not supported by IE :/
Screen Shot 2019-04-05 at 6 30 32 PM

asudoh: Oh I thought we had a polyfill for that but it seems false. Good catch - A couple of approaches here:

* Bring in [the polyfill for vanilla](https://github.com/IBM/carbon-components/blob/v10.0.0/demo/polyfills/toggle-class.js) to the [polyfills directory](https://github.com/IBM/carbon-components-react/tree/v7.0.0/.storybook/polyfills) and update [the polyfills list](https://github.com/IBM/carbon-components-react/blob/v7.0.0/.storybook/polyfills.js) and [the migration guide](https://github.com/IBM/carbon-components-react/blob/v7.0.0/docs/migration/migrate-to-7.x.md#polyfills)
* `.classList[evt.type === 'mouseover' ? 'add' : 'remove'](className)`

vpicone: I just went ahead and added an argument to the event handler.

JessNah:
Thanks @vpicone! Do you have any idea when the new v7.0.1 should be available on npm?

Issue #2133, opened 2019-04-01
Disabled invalid text input is missing invalid outline styles


emyarod:
![image](https://user-images.githubusercontent.com/8265238/55343069-f2049380-546f-11e9-8a13-3f2e3cd1df5f.png)

PR #2134, opened 2019-04-01
fix(TextInput): add invalid attr on wrapper div

emyarod:
Closes IBM/carbon-components-react#2133

This PR adds the invalid state outline to text inputs with a data attribute which was previously missing


asudoh (approved):
LGTM 👍 - Thanks @emyarod!

jnm2377 (approved):

PR #2142, opened 2019-04-03
feat(Tabs): support custom tab content

### LINKED ISSUE #2028: hidden prop of Tabs component is ignored

asudoh:
Fixes #2028.

#### Changelog

**New**

- `renderTabContent` prop in `<Tabs>`, where you can replace `<TabContent>` with.


asudoh:
Storybook is a good point, and I now realized that the custom tab content is per-tab thing, instead of per-tabset thing. Will work on that.

asudoh:
Moved custom rendering React prop from `<Tabs>` to `<Tab>` and added a Storybook example for that.

lovemecomputer:
i think this custom tab content is a good feature! but also to directly address #2028 , i wonder if we should have a prop on `Tabs` that is a boolean named something like `shouldLoadHiddenTabContent`. it could default to `true`, but when set to false it would provide this behavior. does that make sense?

asudoh:
@lovemecomputer My apologies that I'm late getting back on this - I've been kind of on the fence on this - While the notion of providing render vs. no-render for content for inactive tab is plausible, I think it's easily achieved by creating a custom component for tab content rendering. That's where the notion of custom tab content came from. Another idea is making no-render approach the default for next major release. What do you think?

lovemecomputer:
@asudoh this custom component support does make sense to provide! so i definitely agree with this change.

however, the trick with `!selected ? null : <div/>`  is not something i would have immediately thought of. and it also might add an extra step of complexity for any app that needs to do this many times over. in the interest of reducing the friction of using Carbon, it seems like this no-render would be a good feature to also add built-in support for.

i can understand if we need to wait for the next version. no-render as default for v11 is an interesting idea! it might be the ideal behavior… this could use some thinking and testing though!

i can create a separate Issue to track this line of thinking, if you like! :)

asudoh:
Thank you very much for sharing your thoughts @lovemecomputer! I can see a pain-point without an out-of-the-box solution. Created: #2263

Issue #2148, opened 2019-04-03
[SideNav] Add a focus class when sidenav item has focus


davidicus:
### Summary

Focus class is needed to expand the sidenav when sidenav items contain focus. Should be similar to hover state for keyboard users.  Currently you are able to navigate through keyboard and even open subnavs. This causes a bug where the focused item is not visible. Additionally the SideNavMenu collapses and hides its icon when its child list is open. 

### Justification

For accessibility this should open when items are focused. especially hidden subnavs

### Available extra resources

I have opened a PR for this work.

### Extra notes
These full completion of this feature would need changes to the carbon component and carbon styles.

PR #2146, opened 2019-04-03
FEAT(SideNav):  add ability to change default expanded state

davidicus:
Closes IBM/carbon-components-react#2145

Added the ability via prop to declare the default expanded state of the sidenav component

#### Changelog

**Changed**

- Changed SideNav.js
- Changed SideNav-test.js
- Updated SideNav-test.js.snap

PR #2150, opened 2019-04-03
FEAT(SideNav): Side nav focus state

davidicus:
Closes IBM/carbon-components-react#2148

Add focus class name to component in order to open menu when one of its children holds focus.

#### Changelog

**Changed**

- SideNav.js
- SideNav-test.js
- SideNav-test.js.snap

###  Additional notes
~This PR contains work not yet merged from this other [PR](https://github.com/IBM/carbon-components-react/pull/2146) which I will pull in once it is merged. Changes from this [PR](https://github.com/IBM/carbon-components/pull/2261) to Carbon styles  are required to satisfy the issue stated above.~


src/components/UIShell/SideNav.js
```diff
@@ -46,16 +50,32 @@ export default class SideNav extends React.Component {
...
   };
 
-  state = {
-    isExpanded: true,
-  };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isExpanded: props.defaultExpanded,
```
vpicone: These states could be eventually conflicting, resulting in some weird bugs.

If the menu should always be expanded when it has focus, we don't want to create the opportunity for the state to fall out of sync. 

Perhaps we could stick to only using `isExpanded` and setting that based on the focus events.

asudoh: ^ Makes sense, if we want to go with expand-on-focus. In this way, no additional style is needed.

davidicus: I was initially going to use the isExpanded state. But using that would require that we keep track of whether the user had it expanded or collapsed before focus so we could respect their preference after the item blurs, wouldn't it?

src/components/UIShell/SideNav.js
```diff
@@ -64,24 +84,27 @@ export default class SideNav extends React.Component {
...
-      ? t('carbon.sidenav.state.open')
-      : t('carbon.sidenav.state.closed');
+    const assistiveText =
+      isExpanded || isFocused
+        ? t('carbon.sidenav.state.open')
+        : t('carbon.sidenav.state.closed');
     const className = cx({
       [`${prefix}--side-nav`]: true,
-      [`${prefix}--side-nav--expanded`]: isExpanded,
+      [`${prefix}--side-nav--expanded`]: isExpanded || isFocused,
```
davidicus: @asudoh @vpicone I did change the PR to remove the need for an additional className and styles. Is this preferable?

src/components/UIShell/SideNav.js
davidicus: As far as falling out of sync, I can imagine cases where the component can be in both states or just one so may be necessary for them to be out of sync for reason listed above.

asudoh:
Hi @davidicus do you want to split out the code for #2146? Thanks!

davidicus:
Hi @asudoh, I have removed the code form PR #2146 as requested

src/components/UIShell/SideNav.js
```diff
@@ -50,12 +50,23 @@ export default class SideNav extends React.Component {
...
   };
 
+  handleFocus = () => {
+    this.setState(state => {
+      if (!state.isFocused) return { isFocused: true };
+    });
+  };
+
+  handleBlur = () => {
+    this.setState(() => ({ isFocused: false }));
```
asudoh: Previous state check as `.handleFocus()` does?

davidicus: checking if isFocused is already true and only returning a value if needed. I need to return null for the case where it is.

asudoh: Oh I meant that we'd want to put a similar check here.

davidicus: Doh! Sorry, yeah. will do.

src/components/UIShell/SideNav.js
```diff
@@ -50,12 +50,31 @@ export default class SideNav extends React.Component {
...
+    isFocused: false,
   };
 
   handleExpand = () => {
     this.setState(state => ({ isExpanded: !state.isExpanded }));
   };
 
+  handleFocus = () => {
+    this.setState(state => {
+      if (!state.isFocused) {
```
vpicone: It seems like we’d always want this to be set to true no matter what. Same for handleBlur. Maybe just remove these checks since React won’t rerender unless there’s an update.

src/components/UIShell/SideNav.js
vpicone: Good call :)

src/components/UIShell/SideNav.js
davidicus: I thought that setState always triggers an update and causes a rerender, of the virtual DOM if not the actual one. As of [React v16.0 ](https://reactjs.org/blog/2017/09/26/react-v16.0.html#breaking-changes)  returning null will prevent the update.

vpicone: You’re super right, looking good, thanks for make the changes so quickly!

davidicus:
Thanks @vpicone and @asudoh for the quick review and suggestions

Issue #2113, opened 2019-03-28
Match combobox to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/issues/2186

https://github.com/IBM/carbon-components/issues/2210

#1763 

depends on:

https://github.com/IBM/carbon-components-react/pull/2093

https://github.com/IBM/carbon-components/pull/2241

emyarod:
feedback from @carmacleod: https://github.com/IBM/carbon-components-react/issues/1763#issuecomment-476252889

> Some notes:
* [ ] delete `tabindex="0"` from both of the outer divs, i.e. _only_ the input should take focus. So all keyboard events need to be on the input.
* [ ] the `role="button"` needs to go on the div containing the triangle icon. Do not add a tabindex. Mouse events can go on this, but not keyboard events.
* [ ] the `div.bx--list-box__field` should have `role="combobox"` and `aria-haspopup="listbox"`, `aria-expanded="true/false"`, and `aria-owns="unique-id-of-div.bx--list-box__menu"` (note: if using component-user-provided label for id, make sure id does not have spaces).
* [ ] the `div.bx--form-item.bx--combo-box.bx--list-box` should not have any role or any ARIA related markup - it's just a generic container.
* [ ] the `input` could have `role="textbox"` (which is the default for an input type="text", so technically, you can leave the role off), and it needs `aria-controls="unique-id-of-div.bx--list-box__menu"` (same id as for the aria-owns on the combobox div), `aria-autocomplete="list"` (because typing in the input filters the list), and `aria-label="Choose an item"` (delete `aria-label="ListBox input field"`). The input also needs `aria-activedescendant`, but your code there seems good.  :)
* [ ] the `div.bx--list-box__menu` needs `role="listbox"` and the same aria-label as the input (i.e. `aria-label="Choose an item"`).

PR #2158, opened 2019-04-04
fix(ComboBox): match experimental spec

### LINKED ISSUE #1763: ComboBox: Keyboard support needs minor improvements.
There seems to be some minor issues with the ComboBox's keyboard support.

## Detailed description

When I first tab into the component, a focus outline is drawn around the control but the expected keyboard functions (e.g. space, up/down, etc) do not work.  The focus outline also has a strange little bump near the drop arrow. I suspect there's a missing `overflow:hidden` in there somewhere.

From here, if I tab a second time, the component still has focus and some of the keys work but I cannot type in the filter.  In this state, there is no focus outline so I can't visually tell if the control has focus.

From here, if I tab a third time, the input gets focus and I can type a filter.

I think the ideal scenario would be to go straight to that third tab state when the control gets focus.  In this state, I can type in the filter and also use the up/down arrows to select an item from the dropdown, etc.

Latest Chrome 

v6.76.0

### LINKED ISSUE #2081: A11y: Dropdown component have more than one tabbale element
The issue comes from #1851 .

## Environment

> Operating system
MacOS

> Browser
Chrome

> Automated testing tool and ruleset
Manual

## Detailed description

6.58.0

> What did you expect to happen?
Dropdown list should be focused only one time to operate.

> What happened instead?
We must type tab key twice to focus on the dropdown list. When we focus on the dropdown list first time, we can't operate it. Then type the tab key again, We can see the focus is not changed and we can type space to operate the dropdown.

> What WCAG 2.1 checkpoint does the issue violate?
2.1.1 Keyboard

1. Go to https://dev.console.test.cloud.ibm.com/login
2. Type tab key to focus on the Dropdown list.

## Additional information
image

image

emyarod:
Closes IBM/carbon-components-react#2113

Closes #1763 

Closes #2081 

This PR adds the experimental combo box while also addressing a number of a11y and UX issues.

#### Changelog

**New**

- menu option to show overflow text behavior in component story
- require `id` for more accessible labels (affects all listbox components), related https://github.com/IBM/carbon-components-react/pull/1109/

**Changed**

- fix extra tab stops (found in all listbox components)

**Removed**

- unneeded or redundant attributes

#### Testing/Reviewing

Ensure combo box works as expected

Ensure there are no regressions with the listbox components


emyarod:
blocked by https://github.com/IBM/carbon-components-react/pull/2093 and https://github.com/IBM/carbon-components/pull/2241. These changes have not yet been released, but testing can still be done locally via `yarn link`/`npm link`

asudoh:
Removed blocked label given both PRs that @emyarod marked as dependencies seem to have been merged.

emyarod:
@asudoh the only blocker is that we haven't cut a vanilla release which includes the changes from those PRs

laurenmrice (changes requested):
When the dropdown is closed and there is an x icon, when you click the x icon, the drawer opens and the x goes away. If the dropdown is already closed, i think you should be able to click the x and the dropdown will remain closed.
Screen Shot 2019-04-09 at 4 20 06 PM

emyarod:
@laurenmrice it looks like right now we are intentionally opening the dropdown when the user clicks the x to clear selections (https://github.com/IBM/carbon-components-react/blob/master/src/components/ListBox/ListBoxSelection.js#L34-L36). Is this going to be a v10 only change or do we also want to modify the v9 behavior?

laurenmrice:
@emyarod Ok if this is happening in v9 lets just keep it as is for now and revisit this later.

emyarod:
@laurenmrice sounds good! this should be ready for review again then since no changes were made

laurenmrice (dismissed):
- Only change is that we do not need a hover on the label because you can type into the field.
Screen Shot 2019-04-17 at 12 01 46 PM

laurenmrice (approved):
looks good 👍

asudoh (approved):
LGTM 👍 - Thanks @emyarod!

Issue #2076, opened 2019-03-21
ToolTipDefinition needs to be dismissible, hoverable and persistent (WCAG 2.1 Checkpoint 1.4.13)


ajlohr:
## Environment

> Operating system

macOS 10.13.3 (17D47)

> Browser

Chrome Version 73.0.3683.75 (Official Build) (64-bit)

> Automated testing tool and ruleset

IBM Accessibility February 2019

> Assistive technology used to verify

Manual check

## Detailed description

[WCAG 2.1 Checkpoint 1.4.13 Content on Hover Focus](https://www.ibm.com/able/guidelines/ci162/content_on_hover_71.html) says when additional content is popped up on hover it must be dismissible, hoverable and persistent. 

1. dismissible
   * When I hover over the popup there is no way to dismiss it using the ESC key. To comply there has to be another way then moving the mouse or tabbing the keyboard focus out
1.  hoverable
   * When I try to hover over the popup it disappears instead of staying visible
1. Persistent
   * I think its already okay with this one, but would be nice to double check

> What version of the Carbon Design System are you using?

```
        "carbon-components": "9.84.14",
        "carbon-components-react": "6.108.0",
        "carbon-icons": "7.0.7",
```

> What did you expect to happen?

tooltip to be dismissed by clicking ESC key or similar and be persistent when moving mouse over content

> What happened instead?

tooltip popup stayed when ESC was pressed, and popup disappears when trying to hover over the text body

> What WCAG 2.1 checkpoint does the issue violate?

WCAG 2.1 Checkpoint 1.4.13- Content on Hover Focus.

## Steps to reproduce the issue

1. Add ToolTipDefinition and hover over tooltip, try to dismiss the popup without moving mouse or using keyboard focus (tab).

Then try to move the mouse over the popup text and the popup disappears when it should be persistent.

Jaorji:
I'm in the pipeline team, we fond that when we click the hover tooltip, it will disappear.

dakahn:
Hey -- doing some testing. Looks to me like the concerns here are addressed in the latest. 🎉 Happy to open up again if there's something I'm missing.

PR #2164, opened 2019-04-05
fix(tooltip): handle esc keydown event

dakahn:
adds a check in the `handleKeyPress` method that closes the tooltip if escape is pressed.

ref #2076 

### to test
open the tool tip and press escape. the tooltip should close.


src/components/Tooltip/Tooltip.js
```diff
@@ -374,6 +374,11 @@ class Tooltip extends Component {
   handleKeyPress = evt => {
     const key = evt.key || evt.which;
 
+    if (key === 'Escape' || key === 27) {
```
asudoh:
```suggestion
    if (key === 'Escape' || key === 'Esc' || key === 27) {
```

Talking with @vpicone revealed that `evt.key` yields `Esc` in IE and looks that we should cover that.

emyarod: would we want to also update and use these helper functions https://github.com/IBM/carbon-components-react/blob/master/src/tools/key.js?

asudoh: @emyarod Good point - I think it makes sense to me.

asudoh:
Thanks @dakahn for the update and my apologies for being late on this - Updated code basically looks good to me, could you please look at the test failure?

asudoh:
The updated code LGTM 👍 - Thanks @dakahn!

Issue #1953, opened 2019-03-01
Code snippet screen reader issues


dakahn:
@dakahn 
AVT 3 issues found with macOS Mojave, Chrome and VoiceOver.
For the inline code snippet VO does not:
- announce what the button is for. VO should announce that it is a "copy" button.
- announce that the text is a code snippet.
- announce "copied" after the button is pressed the first time. (note: copied is announced after the button is pressed twice).

For the single and multi-line code snippet VO:
- Incorrectly announces that the code snippet is an editable text field.
- Does not announce "copied" after the button is pressed the first time. (note: copied is announced after the button is pressed twice).

_Originally posted by @snidersd in https://github.com/IBM/carbon-components/issues/1531#issuecomment-458255961_

dakahn:
Open PR in vanilla repo [here](https://github.com/IBM/carbon-components/pull/1969)

dakahn:
New testing -- new issues. Reopening @sniders

dakahn:
Addressed inline issues with #2165

Multi-line code snippet issues forthcoming yall 🤠

dakahn:
not stale ⚡️

PR #2165, opened 2019-04-05
fix(code-snippet): add screen reader accessibility improvements

dakahn:
ref #1953 

### change log

- adds a role of alert to the copy verification popup on the code snippet making it accessible to screen readers.

  _this gets the react code snippet in line with changes made to the vanilla component in [this pull request](https://github.com/IBM/carbon-components/pull/1969) 🏄_

- adds more explicit inline code snippet example to storybook
  _this new `aria-label` demonstrates intended use and calls out to user the buttons function_

- removes now unused `copyLabel` prop

- adds `aria-describedby` and corresponding `id` so the code snippet is read aloud

### to test
Run a screen reader. Tab to the code snippet and hit space press space to copy. The copy verification should be read aloud.


dakahn:
There's a few more changes I can bundle up here, so I changed this to a WIP 👍

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -39,11 +39,6 @@ export default class CodeSnippet extends Component {
      */
     feedback: PropTypes.string,
 
-    /**
-     * Specify the label used for the Copy Button
-     */
-    copyLabel: PropTypes.string,
```
asudoh: The removal here seems a breaking change - Could we put a compatibility code to support `copyLabel` until next major?

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -192,15 +186,18 @@ export default class CodeSnippet extends Component {
...
         <Copy
           {...other}
           onClick={onClick}
+          aria-label={ariaLabel}
+          aria-describedby="description"
           className={codeSnippetClasses}
-          aria-label={copyLabel}
           feedback={feedback}>
-          <code>{children}</code>
+          <code id="description">{children}</code>
```
asudoh: We need to ensure ID uniqueness - We can generate a unique ID by: https://github.com/IBM/carbon-components-react/blob/v7.0.0/src/tools/uniqueId.js

src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -192,15 +186,18 @@ export default class CodeSnippet extends Component {
       />
     );
 
+    // TODO: Get the popup announced like an alert
```
tw15egan: Is this ready to be removed?

src/components/CodeSnippet/CodeSnippet.js
dakahn: Hmm, do you mean if they set the `copyLabel` prop we send a warning saying "This prop is non-functional and will be deprecated in Carbon 11 👋🏽"?

asudoh: I see, if your intent is creating a `v11` code then we can add `breakingChangesXI` feature flag to `src/internal/FeatureFlags.js` and we can run new code _only_ when the flag is true.

Another way to deal with this kind of thing is making `copyLabel` work (in addition to `ariaLabel`) until we ship `v11`.

dakahn: Got it. Thanks for clarifying 👍

asudoh: Thanks @dakahn for the update! My last comment to the latest code is that we don't want to remove `ariaLabel`, either (for the sake of backward-compatibility) - So I'm anticipating something like this:

```diff
--- CodeSnippet.js.orig	2019-04-10 23:31:26.000000000 -0500
+++ CodeSnippet.js	2019-04-10 23:37:02.000000000 -0500
@@ -55,6 +55,12 @@
      * Specify a label to be read by screen readers on the containing <textbox>
      * node
      */
+    ariaLabel: PropTypes.string,
+
+    /**
+     * Specify a label to be read by screen readers on the containing <textbox>
+     * node
+     */
     copyLabel: PropTypes.string,
 
     /**
@@ -114,7 +120,8 @@
       children,
       feedback,
       onClick,
-      copyLabel,
+      ariaLabel,
+      copyLabel, // TODO: Merge this prop to `ariaLabel` in `v11`
       copyButtonDescription,
       light,
       showMoreText,
@@ -170,7 +177,7 @@
         role="textbox"
         tabIndex={0}
         className={`${prefix}--snippet-container`}
-        aria-label={copyLabel ? copyLabel : 'code-snippet'}>
+        aria-label={ariaLabel || copyLabel || 'code-snippet'}>
         <code>
           <pre
             ref={codeContent => {
@@ -195,7 +202,7 @@
         <Copy
           {...other}
           onClick={onClick}
-          aria-label={copyLabel}
+          aria-label={copyLabel || ariaLabel}
           aria-describedby={this.uid}
           className={codeSnippetClasses}
           feedback={feedback}>
```

dakahn:
@asudoh Hi there! Just wanna make sure we're good to go. You left a few comments _after approving_. 👍🏽

asudoh:
Thanks a lot @dakahn for double-checking with me! Seems that the only remaining thing is https://github.com/IBM/carbon-components-react/pull/2165#discussion_r274255041, which is, the need to keep `ariaLabel`, but if you didn't do it deliberately with a reason please don't hesitate to bring up!

PR #2174, opened 2019-04-09
feat(OverflowMenu): add menuOptionsClass to overflowMenu

ashharrison90:
Added a new property, `menuOptionsClass`, to `overflowMenu`. This allows a class to be specified on the `overflowMenu` options. This is particularly useful when `floatingMenu` is set, as the menu is appended to the body and has no obvious selector.

#### Changelog

**Changed**

- Changed `overflowMenu` to accept a new property `menuOptionsClass` and pass it down to the menu body.

### DIFF
--- a/src/components/OverflowMenu/OverflowMenu.js
+++ b/src/components/OverflowMenu/OverflowMenu.js
@@ -297,6 +297,11 @@ class OverflowMenu extends Component {
     onClose: PropTypes.func,
 
     /**
+     * The class to apply to the menu options
+     */
+    menuOptionsClass: PropTypes.string,
+
+    /**
      * Function called when menu is closed
      */
     onOpen: PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -613,6 +618,7 @@ class OverflowMenu extends Component {
     );
 
     const overflowMenuOptionsClasses = classNames(
+      this.props.menuOptionsClass,
```
asudoh: Unless you put `menuOptionClass` to [the code destructing `props`](https://github.com/IBM/carbon-components-react/blob/v7.0.0/src/components/OverflowMenu/OverflowMenu.js#L562-L581), the props goes to [the trigger `<div>`](https://github.com/IBM/carbon-components-react/blob/v7.0.0/src/components/OverflowMenu/OverflowMenu.js#L696), too, which we don't want - That said, could you use the variable destructed from `prop` here instead? Thanks!

ashharrison90: good spot! have made the suggested changes 👍

Issue #2185, opened 2019-04-10
Update codepen example to v10


vpicone:
Allows for folks to more easily demonstrate v10 issues in codepen.

PR #2186, opened 2019-04-10
docs: update codepen example

vpicone:
Closes IBM/carbon-components-react#2185


asudoh:
Thank you for doing this @vpicone!

PR #2196, opened 2019-04-11
fix(tabs): dropdown label matches updated tabs

### LINKED ISSUE #2204: Tabs dropdown label references cached tab label
## Detailed description

Currently, updating props.children and props.selected at the same time uses prevProps.children's label in the dropdown. This occurs because getTabAt checks the refs cached in the class, which are not updated until ref is called (after selectedLabel is populated).

1. Update `selected` while also updating the component's `children`

https://codesandbox.io/s/jl71ql90l3

## Additional Information

I've opened #2196 with a simple fix.

Jinhei:
This PR fixes #2204 where the wrong selected dropdown label is shown when `props.children` and `props.selected` are both updated. 

Currently, updating `props.children` and `props.selected` at the same time uses `prevProps.children`'s label in the dropdown. This occurs because `getTabAt` checks the refs cached in the class, which are not updated until `ref` is called (after `selectedLabel` is populated).

### DIFF
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -251,7 +251,7 @@ export default class Tabs extends React.Component {
       }),
     };
 
-    const selectedTab = this.getTabAt(this.state.selected);
+    const selectedTab = tabsWithProps[this.state.selected];
     const selectedLabel = selectedTab ? selectedTab.props.label : '';
 
     return (


asudoh:
Hi @Jinhei could you create an issue with a reduced case based on https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox? We encourage contributors to enter issues accompanied with PRs except for small/internal chores. Thanks!

Jinhei:
@asudoh See #2204

asudoh:
Superseded by #2224 - Sorry for this, I found `tabsWithProps` won't be the best fit.

PR #2199, opened 2019-04-11
chore(ci): re-enable storybook, remove experimental

joshblack:
Re-enable CI deployment for React storybook. This should be merged in after a v6 route is setup.

TODO

- [x] v6 route for React is setup and published with v6 codebase


joshblack:
v6 published using the v6 branch at: http://v6-react.carbondesignsystem.com

PR #2200, opened 2019-04-11
fix(DataTableSkeleton): use new classnames

### LINKED ISSUE #2198: [DataTableSkeleton]: Skeleton still using carbon components v9 classnames

vpicone:
closes #2198


vpicone:
@jendowns As we start removing feature flags, more and more v9 stuff will break. Unfortunately we can't support using versions of carbon-components v9 with carbon-react v7 going forward.

PR #2208, opened 2019-04-12
refactor: update SideNav state management

vpicone:
Related to #2207 

This purpose of this refactor is to 
1. Allow for a `controlled` mode by using the `expanded` prop
2. Allow for a defaultExpanded state to be used without impacting future renders

#### Changelog

**New**

- `controlled`: this ref is set on the first render and does not change on subsequent renders (even as props change). It is enabled by setting the expanded prop.
- `forwardRef`: this allows the user to handle their own focus events should they activate the controlled mode
- `defaultExpanded`: This prop is now used to handle the initial render and only the initial render.

**Changed**

- `isExpanded` → `expanded`: I like `expanded` better but we should probably align on this.

**Removed**
- `getDerivedStateFromProps`
- `setState` from props


vpicone:
The tests I removed are good examples of testing implementation rather than behavior.

The behavior of this component is to change it's appearance based on props.

Our implementation was to set the props to state then use the state to apply classNames which, in turn, get applied to the component and impact it's appearance.

Rather than testing that the state was set (implementation), a better test might check if the classNames were applied. An even _BETTER_ test might ensure that those classNames changed the appearance of our component in the way we expect.

joshblack:
Looking at this quickly, any reason for going from `isExpanded` -> `expanded` as the prop name?

vpicone:
@joshblack I did that for parity with the `defaultExpanded` prop. To keep things aligned, the props would have been `isExpanded` and `defaultIsExpanded`. I could be crazy but `defaultIsExpanded` reads like a statement rather than a boolean to me and feels clunkier. We could leave `Is` out of the default prop(`isExpanded` and `defaultExpanded`) but I think that we should try and keep the prop names as intuitive as possible.

I think the most important think is that we align on a naming convention. If the `is` prefix is the chosen solution I can definitely swap it up.

vpicone:
@joshblack @emyarod

I discussed this morning in person with @asudoh the possibility of just exposing a single `onChange` prop for these components. In that way, we could expose future state changes without increasing the number of props.

```jsx
  // This could become reducer if more state changes were added
  const onChange = ([event, value]) => {
    if (!controlled) {
      setExpandedState(value);
    }
    if (onToggle) {
      onToggle(event, {expanded: value});
    }
  };
```

The exposed API would still just be onChange, however they could be supplied with new events describing the new state.

asudoh:
LGTM 👍 - Let me see how @emyarod responds and will sign-off once it happens - Thanks a lot @vpicone for gathering ideas here and making the updates!

src/components/UIShell/SideNav.js
```diff
@@ -5,131 +5,120 @@
...
+    expanded: expandedProp,
+    defaultExpanded,
+    'aria-label': ariaLabel,
+    'aria-labelledby': ariaLabelledBy,
+    children,
+    onToggle,
+    className: customClassName,
+    translateById: t,
+  } = props;
+  const { current: controlled } = useRef(expandedProp != null);
```
emyarod: should we use strict inequality check here?

vpicone: That's a good question, I assume you mean `!== undefined`? I think you're right since our stance is "using this prop means you control the component."

emyarod: I initially just meant the `!==` operator but yeah I guess it does make sense to look for `undefined`

vpicone: Yeah, I think !== null would miss the main case we want to identify (they don't provide any prop)

PR #2211, opened 2019-04-12
chore(project): update babel deps, remove unused plugins

joshblack:
Updates babel config, removes un-used plugins and updates current ones to latest.

#### Changelog

**New**

**Changed**

- `package.json` babel config and dependencies

**Removed**


vpicone:
How do you feel about just moving the babel config to a `.babelrc` in the root directory? That we we don't have to parse the package.json, and I think it might have the added benefit of making it more legible.

joshblack:
@vpicone in general we've tried to keep as few dotfiles as needed in preference of `package.json`. The reason being that we would have to support multiple dotfiles if we went down that route, or we would have to decide what is/is not suitable for a dotfile in which case it's unclear where you would look for configuration for something. In other words, we would have a situation where we might have:

```
/carbon-components-react/
  .babelrc
  .eslintrc
  .config-a-rc
  .config-b-rc
```

Which would mean too many files at the root level, which can be seen as undesirable. If we moved a couple of the config to `package.json` to alleviate this, for example:

```json5
// package.json
{
  "eslintConfig": {
    // ...
  },
  "config-a": {
    // ...
  },
}
```

And left root as:

```
/carbon-components-react/
  .babelrc
  .config-b-rc
```

Then we lose the ability to quickly understand what config we have for what projects.

As a result, we tended to drift towards `package.json` as the source of config truth given it's ability to colocate this information and be `require`'d by other modules, when needed, to use the configuration.

Unfortunately, with a root-level `.babelrc` we would also need to do a parsing step to normalize for storybook anyways since `./scripts/env` doesn't resolve correctly when used as its config 😞

.storybook/.babelrc.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const path = require('path');
+const packageJson = require('../package.json');
+
+const root = path.resolve(__dirname, '../');
+const babelConfig = Object.keys(packageJson.babel).reduce((acc, key) => {
+  const options = packageJson.babel[key].map(option => {
```
emyarod: we're assuming that `option` is always a string right? do we need to account for plugins like https://github.com/IBM/carbon-components-react/blob/master/package.json#L250-L255?

asudoh:
@vpicone Ping - Any further thoughts? Thanks!

vpicone:
I can definitely see Josh’s point, however I think this intermediary parsing might cause the same sort of confusion. Someone new to the project might see this and think we’re selectively modifying the config. Having two separate Babel configs might not be the worst thing in the world given the difference between the js storybook requires and the bundle we intend on exporting.

vpicone:
As far as the scripts/env I could be missing something but feel like we could configure webpack so that abstraction might not be necessary.

joshblack:
@vpicone would the associated comments help to eliminate any confusion? Left a comment detailing why it was doing what it was doing in order to get storybook to work as intended 👍

joshblack:
RE: the env script, most likely we could but we should pull that out into a separate PR instead of lumping it into this one 👍

vpicone:
Sounds good to me!

PR #2214, opened 2019-04-12
chore(project): update eslint to 5.16.0

joshblack:
Update eslint and associated dependencies. Includes update for react eslint plugin to detect react version.

#### Changelog

**New**

**Changed**

- `package.json` has been updated to include latest eslint deps

**Removed**


vpicone:
I wonder if we might want to revisit our eslint rules at some point. Would be nice to have jsx-a11y cookin.

PR #2216, opened 2019-04-12
chore(rollup): update rollup and config

joshblack:
Update rollup config to 1.10.0, also includes update from `ongenerate` to `generateBundle`

#### Changelog

**New**

**Changed**

- `package.json`
- `generateBundle` update in rollup config

**Removed**


package.json
```diff
@@ -346,5 +346,5 @@
       "enzyme-to-json/serializer"
     ]
   },
-  "bundleSizeThreshold": 120000
+  "bundleSizeThreshold": 150000
```
vpicone: Raising the debt ceiling 😢

joshblack: Yeah I'm not sure what's happening with the size increase, it feels like our bundling is off somewhere and we'll have to start digging into why. My hope is that this upgrade PR will get merged and then we can start updating our bundle output from transpiled babel files to a single entrypoint

scripts/rollup.config.js
```diff
@@ -60,13 +62,13 @@ module.exports = {
...
     resolve({
-      jsnext: true,
-      main: true,
+      mainFields: ['main', 'jsnext'],
     }),
     commonjs({
       include: 'node_modules/**',
       sourceMap: true,
       namedExports: {
+        'node_modules/carbon-components/umd/index.js': ['settings'],
```
asudoh: @joshblack Was your finding that `carbon-components-react` is picking up `umd` directory of `carbon-components` instead of the `es` directory? Just thought it may have contributed to increased bundle size.

joshblack: @asudoh ah that might be it actually, the `node-resolve` plugin updated their fields config so I'm sure I just need to add `module` to the list to pick up `es`, will try real quick

package.json
joshblack: @vpicone bumped it back down with the `module` change!

Issue #2192, opened 2019-04-10
AVT 1 - Modal has DAP violation for missing accessible name


snidersd:
## Environment
macOS Mojave version 10.14.2
Chrome version 72.0.3626.96 (Official Build) (64-bit)

## Detailed Description
Run DAP with Feb. 2019 ruleset on the Form Item Component
The following violations are reported:
An interactive element/widget must have an accessible name
[WAI-ARIA Specification for Widget Roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles)
[WAI-ARIA Specification for Accessible Name Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_name)
<img width="1056" alt="Screen Shot 2019-04-10 at 2 30 08 PM" src="https://user-images.githubusercontent.com/21676914/55907839-6e4c5480-5b9d-11e9-9ffb-e941b2db2369.png">

PR #2217, opened 2019-04-15
fix(Modal): adds text to modalAriaLabel on story.

paschalidi:
Hey 👋 you all! Amazing work here! Appreciate that you give us the opportunity to contribute in such a great project! 🕊 

Closes https://github.com/IBM/carbon-components-react/issues/2192

Currently there is no text added in the property `modalAriaLabel` when its being initialised on the story. 
Therefore there is a DAP violation that is being described on the issue above. 
To solve that I added an actual text on that property.

#### Changelog

**New**

- adds text to modalAriaLabel on story.

**Changed**

- The property `modalAriaLabel` has value `A label to be read by screen readers on the modal root node` on `Modal-story.js`


paschalidi:
@asudoh thanks for the quick review! :D

PR #2221, opened 2019-04-16
chore: update carbon-components for storybook

vpicone:
Brings in new styles and updates for data table work


package.json
```diff
@@ -242,8 +242,16 @@
       "@commitlint/config-conventional"
     ],
     "rules": {
-      "header-max-length": [2, "always", 50],
-      "body-max-line-length": [2, "always", 72]
+      "header-max-length": [
```
emyarod: just wondering, did we change our linter settings?

vpicone: Might be from a prettier update, the change happened with the linter/formatting. Did you not get the same change?

emyarod: 🤔not sure where it came from, just noticed that it was causing merge conflicts

PR #2224, opened 2019-04-17
fix(Tabs): avoid stale tab ref cache

### LINKED ISSUE #2204: Tabs dropdown label references cached tab label

asudoh:
The problem happens when `selected` and `children` changes at the same time.

Fixes #2204.

#### Changelog

**Changed**

- A change not to use the cache of tab's DOM ref when the `<Tabs>` code gets selected tab's label.


asudoh:
@emyarod Good question - #2196 calls a different internal method that has more overhead (notably copying props over). That's why I came up with a different PR. Hope it explains things well.

asudoh:
> so this would supersede #2196 then right?

Yes.

Issue #2230, opened 2019-04-17
Icon default description value masks DAP (accessbility) violations/problems


justindm234:
## Detailed description

An Icon component defined without a description does not cause a Dynamic Assessment Plugin violation, because the component is defined with a default description property value. This default description generates an invalid (from an a11y perspective) aria-label attribute value in the HTML. This leads to a false negative in DAP (no reported violation).

> What version of the Carbon Design System are you using?

7.1.0

> What did you expect to happen?

A DAP violation for missing aria-label should have been generated

> What happened instead?

No DAP violation was generated

> What WCAG 2.1 checkpoint does the issue violate?

1.1.1 Non-text Content - The default alternative text ("Provide a description that will be used as the title") is not "appropriate or equivalent".

## Steps to reproduce the issue

1. Define an Icon component without a description property

> https://codesandbox.io/s/x7nopkj8l4

PR #2235, opened 2019-04-18
fix(Icon): removes description default value

paschalidi:
✨ Hey you all ✨Hope you are doing all fine and coding happily!

Closes https://github.com/IBM/carbon-components-react/issues/2230

When the Icon component is being used without the `description` property the default is being assigned to `description`. The default value at the moment is  `Provide a description that will be used as the title`. This default description generates an invalid (from an a11y perspective) aria-label attribute value in the HTML, which can cause confusion to the end user. This leads to a false negative in DAP (no reported violation).

#### Changelog
**Removed**
Removed the following property from the Icon component.
`descripton: Provide a description that will be used as the title`

Please dont hesitate to raise questions! Peace out!


paschalidi:
@asudoh @emyarod thanks for merging this and taking the time to review it!

Issue #2109, opened 2019-03-28
Match multiselect to experimental spec


emyarod:
Related:

https://github.com/IBM/carbon-components/pull/2181

https://github.com/IBM/carbon-components/pull/2207

https://github.com/IBM/carbon-components-react/issues/1611

depends on:
https://github.com/IBM/carbon-components-react/pull/2090
https://github.com/IBM/carbon-components-react/pull/2093
https://github.com/IBM/carbon-components/pull/2228
https://github.com/IBM/carbon-components-react/issues/2113
https://github.com/IBM/carbon-components/pull/2265

emyarod:
because of the filterable multi select, this will depend on https://github.com/IBM/carbon-components-react/issues/2113

laurenmrice:
not stale

Issue #1611, opened 2018-12-05
MultiSelect a11y issues


nicoxxie:
## Detailed description

> Describe in detail the issue you're having.
When I tab onto a MultiSelect component, the focus is correct but cannot do anything by keyboard.
![screen shot 2018-12-05 at 2 06 01 pm](https://user-images.githubusercontent.com/17915105/49540951-f987a780-f896-11e8-95ac-5f155a47e895.png)

Then I tab again, the focus is on the dropdown arrow, I can use the keyboard to expand it but the focus is not shown at all.
![screen shot 2018-12-05 at 2 06 20 pm](https://user-images.githubusercontent.com/17915105/49540950-f987a780-f896-11e8-820b-a5bcc28ef15a.png)

> Is this issue related to a specific component?
MultiSelect

> What did you expect to happen? What happened instead? What would you like to see changed?
Hover on the first tab and able to expand by keyboard on there.

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
6.55.5

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cloud

## Steps to reproduce the issue
Can be tested on http://react.carbondesignsystem.com/?selectedKind=MultiSelect&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

> Please create a reduced test case in CodeSandbox
> https://codesandbox.io/s/github/IBM/carbon-components-react/tree/master/examples/codesandbox

## Additional information

- Screenshots or code
- Notes

frastlin:
Confirm using NVDA 2018.3.2 on Chrome Version 71.0.3578.80 (Official Build) (64-bit) and 63.0.3 (64-bit).

I would like to add that the element that is used to represent the combobox is not an element I would click on and is not labeled with aria role="combobox". A better element would be a `<input type="text" role="combobox">`. For a perfect implementation of the desired behavior, please see the [AccDC Example](http://whatsock.com/tsg/Coding%20Arena/ARIA%20Comboboxes/ARIA%20Comboboxes%20(Simulated,%20Readonly%20Multiselect)/demo.htm)

yacineMTB:
https://github.com/IBM/carbon-components-react/pull/1731

shinytoyrobots:
Closed by #1731 merge.

nicoxxie:
[...]

PR #2239, opened 2019-04-18
feat(MultiSelect): match experimental spec

### LINKED ISSUE #2009: MultiSelect.Filterable invalid: true styling bug.
## Detailed description

> While using `<MultiSelect.Filterable>` component with `invalid={true}` the style of bottom shadow has a bug. Here is couple of screenshots: 

not focused
Here you can see that red box shadow also has blue color over it.

focused
You can see it even better when input field used for filtering inside MultiSelect is focused

That's all because of `bx--text-input` style.
**Normal state:** while `bx--multi-select` has `data-invalid="true"`, `bx--text-input` which is inside still has `box-shadow: 0 1px 0 0 #5a6872;` 
**Focus state:** `box-shadow: 0 2px 0 0 #3d70b2;` for `.bx--text-input:focus` which causes the bug.

1. Open sandbox - https://xrmwxlyo5w.codesandbox.io/
2. Observe the bug
3. ?????
4. You are awesome

emyarod:
Closes IBM/carbon-components-react#2109

Closes #2009

This PR adds the v10 multiselect and filterable multiselect

Related https://github.com/IBM/carbon-components-react/issues/1611

depends on https://github.com/IBM/carbon-components/pull/2326 and corresponding version release

we should also merge https://github.com/IBM/carbon-components-react/pull/2090 before this PR

#### Changelog

**New**

- storybook knobs to expose all options
- v10 multiselect and filterable multiselect

**Changed**

- align checkbox classes to vanilla

**Removed**

- duplicated/unused props


emyarod:
since https://github.com/IBM/carbon-components-react/pull/2090 affects the behavior of the component, it might be good to merge that PR in before this one

dakahn (approved):

asudoh:
Interim comment: `v9` support code is no longer needed, other than that LGTM.

emyarod:
PR requires designer feedback but it seems our netlify integration is still down

src/components/MultiSelect/MultiSelect-story.js
```diff
@@ -20,15 +18,29 @@ import {
 import { withReadme } from 'storybook-readme';
 import readme from './README.md';
 import MultiSelect from '../MultiSelect';
+import { componentsX } from '../../internal/FeatureFlags';
```
asudoh: Just one remainder of `v10` feature flag ref.

emyarod: removed, thanks!

src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -253,15 +257,46 @@ export default class FilterableMultiSelect extends React.Component {
...
-        [`${prefix}--list-box--light`]: light,
+        [`${prefix}--multi-select__wrapper--inline`]: inline,
+        [`${prefix}--list-box__wrapper--inline`]: inline,
+        [`${prefix}--multi-select__wrapper--inline--invalid`]:
+          inline && invalid,
+        [`${prefix}--list-box__wrapper--inline--invalid`]: inline && invalid,
       }
     );
-    return (
+    const className = ({ selectedItem }) =>
```
asudoh: `className` as a function is not seen in other components. Probably move this to render callback?

asudoh (commented):
Thank you for the update! Just another review comment.

src/components/MultiSelect/FilterableMultiSelect.js
emyarod: this is how I implemented it for our [Dropdown](https://github.com/carbon-design-system/carbon-components-react/pull/2093/files#diff-700df727ae95bd5df9e9b8baa2eb6a1eR158) as well to get the dropdown state from downshift. Should I create a separate PR to change dropdown?

asudoh: Yes, it makes sense!

asudoh (approved):
LGTM 👍 - Thanks @emyarod!

dakahn:
Super psyched for this. Thanks for all your hard work @emyarod 🙏

Issue #2193, opened 2019-04-11
React Tooltip - Accessibility markup cleanup


carmacleod:
There are some unnecessary attributes in the accessibility-related markup for the React Tooltip.

For an overview, please see the screen capture of the markup, below.

Please note:
- if the user provides visible text content for the bx--tooltip__label, then the button should use `aria-labelledby` to point to its id, and if the user doesn't have visible text, then they need to provide an aria-label for the button
- the component should have a sensible default aria-label, like "Info" or "Help" (not "tooltip"), and this default should be published in the doc for the aria-label prop (although users should be encouraged to provide their own)

Please delete:
- title - it is completely unnecessary
- aria-owns - this attribute is not used in the tooltip pattern
- role="img" and aria-label="tooltip" on the svg - these are not necessary because the button label overrides the image label (or keep role="img" and aria-label[ledby] on the svg and remove aria-label[ledby] from the button)
- alt="tooltip" on the svg - just delete this - alt is not a valid attribute on svg elements
- aria-labelledby on the tooltip div - this is not used in the tooltip pattern

Please consider:
- try to use a button element instead of `div role="button" tabindex="0"` ... and use onclick instead of handling space and enter keys
- feel free to test with [this little test case](https://carmacleod.github.io/playground/tooltip-test.html) before switching to button. It uses a real button with (mostly) bx styles, and it looks ok.

![image](https://user-images.githubusercontent.com/3331913/55920628-8fc23600-5bc7-11e9-8de1-91b239d50323.png)

paschalidi:
@carmacleod will look into this tomorrow!

paschalidi:
>try to use a button element instead of div role="button" tabindex="0" ... and use onclick instead of handling space and enter keys

Is this part of the requirements 100 percent? I am happy to check this out then! Le me know!

paschalidi:
@carmacleod you did an amazing job explaining what needs to be changed here!! Thanks a lot for your time and effort you reeeeally helped me out realise what needs to be done here! Amazing job, thank you very much!

paschalidi:
@carmacleod can we please have a look what is left here?

I think in my understanding, only
>try to use a button element instead of div role="button" tabindex="0" ... and use onclick instead of handling space and enter keys

I am not sure if the above is a necessity of a "good to have" feature tho. 

Can you please when you get some time confirm?
Then I am happy to assign this to me and take care of it.

dakahn:
@carmacleod 👋

[...]

PR #2245, opened 2019-04-19
fix(Tooltop): a11y improvements

paschalidi:
🌳 👋 hello to you all 🌻  Hope you enjoying the great day!

Continues on https://github.com/IBM/carbon-components-react/issues/2193

#### Tackles the following
* removes `IconTitle` - its not needed for a11y.
* removes `aria-owns` -   this attribute its not needed for a11y in the `tooltip` according to https://github.com/IBM/carbon-components-react/issues/2193
* removes `aria-labelledby` this attribute its not needed for a11y in the `tooltip` according to https://github.com/IBM/carbon-components-react/issues/2193
* when `Icon` is present then the div wrapping the `Icon` has all the attributes are are relevant to the action of showing the `Tooltip`. When the `Icon` component is not present those attributes are now being parsed to the div that wraps the actual text. I call these attributes `properties` its not a very distinct name please feel free to suggest something more meaningful! 
* when `triggerText` is not being parsed user gets error message that the property `iconDescription` is required in order to have better user experience

#### Changelog

Tooltip a11y changes.

##Future 👨‍🚀 
The next PR would be regarding 
> try to use a button element instead of div role="button" tabindex="0" ... and use onclick instead of handling space and enter keys

from https://github.com/IBM/carbon-components-react/issues/2193


paschalidi:
@dakahn i can reproduce it. will take care of it as soon as possible. Thanks for taking care of this

paschalidi:
@dakahn hey thanks for reviewing!  In my FF the big is now fixed. Problem was that I created that `FinalIcon` functional component and `refs` cannot be passed into functional components just classes therefore clicking in the icon itself would feel like clicking also outside of the tooltip therefore would trigger it to close it.. 

Well hope that makes little bit of sense.
Have another look when your time allows it!

src/components/Tooltip/Tooltip.js
```diff
@@ -216,12 +216,14 @@ class Tooltip extends Component {
     /**
      * The description of the default tooltip icon, to be put in its SVG 'aria-label' and 'alt' .
      */
-    iconDescription: PropTypes.string,
-
-    /**
-     * The title of the default tooltip icon, to be put in its SVG `<title>` element.
-     */
-    iconTitle: PropTypes.string,
+    iconDescription: ({ triggerText, iconDescription }) => {
```
asudoh: Seems close to the use case of: https://github.com/IBM/carbon-components-react/blob/v7.1.0/src/prop-types/isRequiredOneOf.js

src/components/Tooltip/Tooltip.js
```diff
@@ -452,83 +451,58 @@ class Tooltip extends Component {
...
-                onBlur={this.handleMouse}
-                aria-haspopup="true"
-                aria-label={iconDescription}
-                aria-expanded={open}
-                {...ariaDescribedbyProps}
-                {...ariaOwnsProps}>
-                {finalIcon}
+              <div className={`${prefix}--tooltip__trigger`} {...properties}>
+                {IconCustomElement ? (
+                  <IconCustomElement {...iconProperties} />
```
asudoh: Need to set `triggerEl` for `<IconCustomElement>` in addition to `<Icon>`.

src/components/Tooltip/Tooltip.js
```diff
@@ -452,83 +451,58 @@ class Tooltip extends Component {
...
+    const properties = {
+      role: 'button',
+      tabIndex: tabIndex,
+      onClick: this.handleMouse,
+      onKeyDown: this.handleKeyPress,
+      onMouseOver: this.handleMouse,
+      onMouseOut: this.handleMouse,
+      onFocus: this.handleMouse,
+      onBlur: this.handleMouse,
+      ref: mergeRefs(ref, node => {
```
asudoh: This `ref` should be set to `<div className={triggerClasses}>` _only_ when its content is text-only. This is for locating tooltip relative to icon when it's present, and to the containing `<div>` otherwise.

paschalidi: thanks for explaining, was not very clear to me .. will fix that

paschalidi:
@asudoh maybe have another look

asudoh:
@paschalidi Have you had a chance to address my comments...? Thanks!

paschalidi:
I did but for some reason never pushed those changes. very sorry for notifying for no good reason @asudoh.

Issue #1764, opened 2019-01-17
Tooltip needs keyboard accessibility improvements


carmacleod:
Please see https://github.com/IBM/carbon-components/issues/1264

Please fix both problems:
1. the tab key not focusing the first element in the tooltip (the fix is to move the tooltip flyout DOM to be right after the tooltip trigger DOM)
2. Esc key not closing the tooltip

carmacleod:
This problem has come up for us again. It works in Vanilla, but not in React.

carmacleod:
Thanks!

PR #2248, opened 2019-04-19
fix(Tooltip): adds esc listener

paschalidi:
Hello 👋 ! 

Closes https://github.com/IBM/carbon-components-react/issues/1764

Adds event listener so that Esc key closes `Tooltip` when state is `open`.

#### Changelog

**New**

Adds event listener so that Esc key closes `Tooltip` when state is `open`.


src/components/Tooltip/Tooltip.js
```diff
@@ -266,6 +266,8 @@ class Tooltip extends Component {
     requestAnimationFrame(() => {
       this.getTriggerPosition();
     });
+
+    document.addEventListener('keydown', this.handleEscKeyPress, false);
```
asudoh: Please clean-up this event listener on un-mount.

src/components/Tooltip/Tooltip.js
```diff
@@ -391,6 +393,13 @@ class Tooltip extends Component {
     }
   };
 
+  handleEscKeyPress = event => {
+    const { open } = this.state;
+    if (open && keyDownMatch(event, [keys.ESC, keyCodes.ESC, keyCodes.IEESC])) {
```
dakahn: Great job grabbing this util 👍

paschalidi: Thanks @dakahn, appreciate it!

src/components/Tooltip/Tooltip.js
paschalidi: thanks for the quick review! Keep it up!

paschalidi:
@asudoh is something holding us back here?

asudoh:
@paschalidi Sorry for being late!

PR #2255, opened 2019-04-22
fix(notification): adds aria-label to button

### LINKED ISSUE #2246: AVT 1 - Toast Notification: DAP violation the form control is missing an associated
## Environment
macOS Mojave version 10.14.2

## Detailed Description
Run DAP with Feb. 2019 ruleset on the Expandable Tile component
See screenshot below for violation:
Screen Shot 2019-04-19 at 8 50 44 AM

dakahn:
closes #2246 

adds aria-label to close button on toast notification so that the button's purpose can be known to screen reader users.

#### Changelog

- adds `aria-label={iconDescription}` to the `<button>` component

### DIFF
--- a/src/components/Notification/Notification.js
+++ b/src/components/Notification/Notification.js
@@ -665,7 +665,6 @@ class Notification extends Component {
         className={notificationClasses.toast}>
         <NotificationTextDetails
           title={title}
-          aria-label="catcatcat"
           subtitle={subtitle}
           caption={caption}
           notificationType="toast"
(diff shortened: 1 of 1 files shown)


src/components/Notification/Notification.js
```diff
@@ -664,6 +665,7 @@ class Notification extends Component {
         className={notificationClasses.toast}>
         <NotificationTextDetails
           title={title}
+          aria-label="catcatcat"
```
asudoh: Did you mean to use an existing prop or introduce a prop here?

dakahn: 🤦‍♂️ Ugh, my bad. That was from testing. Will remove.

Issue #2242, opened 2019-04-18
AVT 1 - Expandable Tile: DAP violation the form control is missing an associated label


snidersd:
## Environment
macOS Mojave version 10.14.2
Chrome version 72.0.3626.96 (Official Build) (64-bit)

## Detailed Description
Run DAP with Feb. 2019 ruleset on the Expandable Tile component
See screenshot below for violation:
<img width="1060" alt="Screen Shot 2019-04-18 at 3 34 21 PM" src="https://user-images.githubusercontent.com/21676914/56389730-e3e4a000-61ef-11e9-80e8-ad7d71b376a8.png">

PR #2258, opened 2019-04-22
fix(tile): various DAP fixes

### LINKED ISSUE #2238: AVT 1 - Tile Multi-Select: DAP violations
## Environment
macOS Mojave version 10.14.2

## Detailed Description
Run DAP with Feb. 2019 ruleset on the Tile Multi-select component
See screenshot below for violations related to id's and labels that are not unique:
Screen Shot 2019-04-18 at 11 22 55 AM

dakahn:
closes #2238, #2242 

A few quick fixes getting rid of a handful of DAP errors reported in the aforementioned issues. 

#### Changelog

- adds unique IDs to the tile-story for tile group
- adds `aria-labelledby` attribute to the `button` in expandable tile


src/components/Tile/Tile.js
```diff
@@ -464,9 +464,12 @@ export class ExpandableTile extends Component {
         {...other}
         onClick={this.handleClick}
         tabIndex={tabIndex}>
-        <button className={`${prefix}--tile__chevron`}>
+        <button
+          className={`${prefix}--tile__chevron`}
+          aria-labelledby="tileExpandIcon">
           {componentsX ? (
             <ChevronDown16
+              id="tileExpandIcon"
```
asudoh: Need a way to avoid ID duplication, if you search for [`uniqueId.js`](https://github.com/IBM/carbon-components-react/blob/v7.1.0/src/tools/uniqueId.js) references you'll see how other components in `carbon-cmoponents-react` solves this problem.

src/components/Tile/Tile.js
```diff
@@ -453,6 +454,10 @@ export class ExpandableTile extends Component {
     const content = this.getChildren().map((child, index) => {
       return React.cloneElement(child, { ref: index });
     });
+
+    // a unique ID generated for use in aria-labelledby
+    this.uid = uid();
```
asudoh: The `uid` is created in creation time instead of rendering time - Otherwise every `render()` changes the unique ID and thus redundant rendering will happen.

src/components/Tile/Tile.js
```diff
@@ -423,6 +424,9 @@ export class ExpandableTile extends Component {
     return React.Children.map(this.props.children, child => child);
   };
 
+  // a unique ID generated for use in aria-labelledby
+  uid = uid();
```
emyarod: should this match `this.props.id` (when it is provided)?

asudoh: @emyarod Couldn't find `id` prop for this component? (Though I could find ones in other tile variants)

emyarod: @asudoh is that intentional or does this variant not need an `id` then? I assumed it did for a11y purposes

asudoh: Are you referring to `<SelectableTile>`? As far as I can see it didn't take auto-ID-generation route.

emyarod: sorry I mean `ExpandableTile`, should we allow consumers to set their own `id`s rather than just always auto generating one? something like `uid = this.props.id || uid()`?

asudoh: OK you mean introducing `id` prop? If so it sounds good to me?

dakahn:
Sorry for letting this go stale! Added the `id` prop to ExpandableTile 👍

src/components/Tile/Tile.js
```diff
@@ -453,6 +462,7 @@ export class ExpandableTile extends Component {
     const content = this.getChildren().map((child, index) => {
       return React.cloneElement(child, { ref: index });
     });
+
```
asudoh:
```suggestion
    const buttonId = this.props.id ? `${this.props.id}__button` : this.uid;
```

src/components/Tile/Tile.js
```diff
@@ -464,9 +474,12 @@ export class ExpandableTile extends Component {
         {...other}
         onClick={this.handleClick}
         tabIndex={tabIndex}>
-        <button className={`${prefix}--tile__chevron`}>
+        <button
+          className={`${prefix}--tile__chevron`}
+          aria-labelledby={this.uid}>
```
asudoh:
```suggestion
          aria-labelledby={buttonId}>
```

src/components/Tile/Tile.js
```diff
@@ -464,9 +474,12 @@ export class ExpandableTile extends Component {
...
           {componentsX ? (
             <ChevronDown16
+              id={this.uid}
```
asudoh:
```suggestion
              id={buttonId}
```

src/components/Tile/Tile.js
```diff
@@ -423,6 +429,9 @@ export class ExpandableTile extends Component {
     return React.Children.map(this.props.children, child => child);
   };
 
+  // a unique ID generated for use in aria-labelledby if one isn't providedj
+  uid = this.props.id || uid();
```
asudoh:
```suggestion
  uid = uid();
```

asudoh:
Thanks @dakahn for the latest update! @emyarod Would you be able to review? Thanks!

src/components/Tile/Tile.js
```diff
@@ -416,6 +422,9 @@ export class ExpandableTile extends Component {
     return React.Children.map(this.props.children, child => child);
   };
 
+  // a unique ID generated for use in aria-labelledby if one isn't providedj
+  uid = uid();
```
emyarod: originally I meant instead of creating a new variable wouldn't we just be able to do this?

```suggestion
  uid = uid();
  uid = this.props.id || uid();
```

asudoh: @emyarod Doing so prevents change in `props.id` after initialization being ignored.

PR #2260, opened 2019-04-22
fix(pagination): add noLabel prop to Select

### LINKED ISSUE #2236: AVT 1 - Pagination: DAP violations
## Environment
macOS Mojave version 10.14.2

## Detailed Description
Run DAP on the pagination component
See screenshot below for violations:
Screen Shot 2019-04-18 at 11 06 57 AM

Note: The label ID's are not unique as displayed in the screenshot.

dakahn:
closes #2236 

DAP was throwing an error for having multiple labels on the same control. This adds `noLabel` prop to Select for use when select is an embedded component/descendant to an existing label and doesn't needs it's own.

#### Changelog

- add `noLabel` prop to Select
- add logic to show or hide `<label>` in render method of select
- set `noLabel` on nested Select in Pagination component


src/components/Pagination/Pagination.js
```diff
@@ -300,7 +300,7 @@ export default class Pagination extends Component {
             id={`${prefix}-pagination-select-${inputId}`}
             className={`${prefix}--select__item-count`}
             labelText=""
-            hideLabel
+            noLabel
```
asudoh: This is a breaking change, need to preserve the name or need to introduce a compatibility layer.

emyarod: @asudoh could we look into https://github.com/IBM/carbon-components-react/issues/1662 again for v11?

src/components/Select/Select.js
```diff
@@ -26,6 +26,7 @@ const Select = React.forwardRef(
       labelText,
       disabled,
       children,
+      noLabel,
```
asudoh: Uh noticed that my earlier comment to `<Pagination>` was just to the `<Select>` usage and the API discussion should have been here. So `<Select>` currently has `hideLabel` which visually hides this, and `noLabel` which removes the label from DOM. @dakahn Do you think we should have both, or do you think just having a way to removing label from DOM is sufficient?

src/components/Pagination/Pagination.js
asudoh: @emyarod Yes, just need an API proposal with research of what components are affected and how.

src/components/Select/Select.js
dakahn: There are definitely use cases for _visually_ hiding the label, but preserving it in the DOM. There is some question as to whether we should expose noLabel as part of this components API though. I guess if they were building their own widget with Select and (like Pagination) don't want redundant labeling causing errors in AAT noLabel is useful.

asudoh: Thanks @dakahn for your thoughts! Where we should expose `noLabel` or not is a good point -  Now I'm wondering if `<Pagination>` not using `<Select>` would solve the issue 🤔?

src/components/Pagination/Pagination.js
emyarod: sounds good @asudoh, will update that ticket body with more information from my earlier audits

src/components/Select/Select.js
dakahn: Something we can plan to do -- I know a few components are due for a bit of a cleanup/refactor. In the meantime we can use this prop to get right with DAP.

asudoh: OK given removal of prop will be a breaking change in future, can we add a word to `noLabel` prop saying that it's reserved for `<Pagination>`? Thanks!

asudoh: Thanks @dakahn for the update! Can we add `noLabel` in `Select.propTypes` with such wording, too?

vpicone:
Maybe a dumb question, but is there a reason we’re not just using the Select component’s labelText prop as intended? This seems like a weird way about it as a one off. Would we want to give every form input the ability to not render a label for consistency?

asudoh:
@vpicone Good question, I imagined that empty `labelText` should work the same as `hideLabel` (something @emyarod is interested in tackling), but no-rendering route there may be an interesting option.  CC @dakahn

Issue #1607, opened 2018-12-04
Disabled row will be selected by select all in data table


nelsonchen90:
## Detailed description

> Describe in detail the issue you're having.

In DataTable, we have to disable some rows for multi-selection as they don't make sense in the row context. The checkbox for that row is disabled and user can not interact with it. But if user use the select all check box from the table header, the disabled rows will be selected and added to the `selectedRows` in the `render` props

> Is this issue related to a specific component?

`<DataTable />`

> What did you expect to happen? What happened instead? What would you like to see changed?

1. Disabled rows should not be selectable. 
2. it should also not counted as the select all tri-state checkbox, meaning if all the selectable rows are selected, the select all checkbox should be checked, not indeterminate state.

> What browser are you working in?

Chrome, but I believe it's not browser specific

> What version of the Carbon Design System are you using?

6.51.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

ICP4D, UG&I
2019 Q1

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in code sandbox
https://codesandbox.io/s/wq7nx362z5

## Additional information
https://ibm-analytics.slack.com/archives/C2K6RFJ1G/p1543941572048700

Issue #1247, opened 2018-08-28
[Enhancement] Add ability to pre-select rows in DataTable


joshblack:
Tried checking before making this to see if this was already an issue, but didn't see any ones specifically for creating the enhancement. The closest I could find was: #1088.

This issue is for discussion around adding in the ability to pre-select rows in the DataTable, and then for figuring out who can take on this feature and built it out in a PR 👍

javadude24:
This was not an enhancement request, but a more details to explain how I can pre-set selected row. 

This is a write up on renderProps, but not sure how to implement.  I've tried setting isSelected in my base table data. But how does it get converted into the array called  'rowsById', which also has a isSelected attribute. However, the mystery is where this reduction of row data happen and why my isSelected does Not have an impact.

https://medium.com/carbondesign/data-table-updates-in-carbon-react-v5-22-0-6da0c24a96d6

marklindsay24:
There is a very simple fix for this, allowing for raw data rows to have isSelected. In the normalize function, pull initial value from there.

![datatable-isselected](https://user-images.githubusercontent.com/42783294/44744053-27a85f00-aad2-11e8-8e92-497309dfab2f.png)


Please let me know if I can fix for team.

joshblack:
Oh duh, thanks @javadude24!

David-Yuen:
Just adding to the post: I had originally requested this feature on slack

David-Yuen:
What kind of variable name do you want to be exposed externally?

Currently the PR is set using `isSelected`, which is what is used internally. 

From an external's perspective `row.isSelected` directly equals "is the row selected?", which should be answered with a boolean `true` or `false`. My opinion, besides the exposed external name having the same name as the internal name, `isSelected` is a straight-forward and valid API name and wouldn't cause any confusion for developers using it.

Josh raised that he wasn't comfortable with exposing internal variables, I agree but at the same time, that is not the intended pattern/format that will follow for all variables going forward. Some variables (used internally and externally) are generic or very basic for the reason of understanding, and I think the most important thing is the variable name is straight forward and self-describes the purpose within the given scope (such as internally or externally). So I personally don't think it is an issue in this case to having it called `isSelected`. 

Otherwise, the other options I can think of is `selected` or `checked` but to me, doesn't feel as strong as `isSelected` or `isChecked` because adding `is` is asking yes or no. 
[...]

PR #2271, opened 2019-04-24
fix(DataTable): selection fix

harshzalavadiya:
Closes IBM/carbon-components-react#1607
Closes IBM/carbon-components-react#1247

- As of now there's no functionality to specify `pre-selected`, `pre-expanded` or `pre-disabled` rows
- If user specifies `_disabled`, `_expanded` and `_selected` key with true or false it will set it as initial value
- If row is disabled checkboxes will be automatically disabled and will not be selectable by selectall checkbox trigger

#### Testing
- Apply attached patch/diff file [testing-selection.patch.txt](https://github.com/IBM/carbon-components-react/files/3112050/testing-selection.patch.txt)

#### Changelog

**Changed**

- DataTable selection utils


src/components/DataTable/tools/normalize.js
```diff
@@ -26,8 +26,9 @@ const normalize = (rows, headers, prevState = {}) => {
     // expansion
     rowsById[row.id] = {
       id: row.id,
-      isSelected: false,
-      isExpanded: false,
+      isSelected: row._selected || false,
```
vpicone: We typically only want to use the underscore to prefix internal methods/properties. Did you intend on leaving the underscore here?

I think keeping the same property names is cleaner and a more intuitive API.

```js
  const {id, isSelected, isExpanded, disabled} = row;
  rowsById[row.id] = {
    id,
    isSelected,
    isExpanded,
    disabled
  }
```

harshzalavadiya: Hi @vpicone ,

I guess you are right keeping underscore anyhow represents internal so having same column names are good to go. I have made changes as suggested.

harshzalavadiya:
> hi @harshzalavadiya thanks for jumping in! would you like to expose these features in our storybook docs as well?

Hi @emyarod 
I have made changes for docs, let me know if any changes are required.

Issue #2264, opened 2019-04-22
Pagination backward and forward buttons incorrectly marked as disabled


austinbruch:
## Detailed description

> Describe in detail the issue you're having.

When using the `Pagination` component, when specifying the `pageInputDisabled` prop as `true`, not only is the select subcomponent removed (which is expected), but the backward and forward buttons also have the `bx--pagination__button--no-index` class applied, causing the cursor to indicate that these buttons are disabled. As I understand the purpose of the `pageInputDisabled` prop, this prop is supposed to only control whether the `select` which allows for direct page change is rendered. I do not expect this prop to apply classes to the backward and forward buttons, making them appear disabled.

However, even in this state, the backward and forward buttons _still work_, even though when mousing over them, the cursor indicates that they are disabled. I think that is a separate but related issue.

> Is this issue related to a specific component?

Yes, `Pagination`.

> What did you expect to happen? What happened instead? What would you like to see changed?

See above description.

> What browser are you working in?

Chrome 73

> What version of the Carbon Design System are you using?

"carbon-components-react": "7.1.0",

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM Cloud Identity

## Steps to reproduce the issue
Render `Pagination` component, set `pageInputDisabled` prop to true.

> Please create a reduced test case in CodeSandbox

https://codesandbox.io/s/30pj14pkjm

PR #1800, opened 2019-01-28
feat(PaginationV2): add experimental Pagination

emyarod:
Closes IBM/carbon-components-react#1724

This PR will add the experimental pagination component

Related: 

https://github.com/IBM/carbon-components/issues/1251

https://github.com/IBM/carbon-components/issues/1304

https://github.com/IBM/carbon-components/pull/1518

https://github.com/IBM/carbon-components/issues/1638

https://github.com/IBM/carbon-components/pull/1679

---

question for reviewers: will we be deprecating `PaginationV1`?

PR #2278, opened 2019-04-28
fix(Pagination): pageInputDisabled won't append disabled class to buttons

kleyu:
Closes IBM/carbon-components-react#2264

Pagination component with `pageInputDisabled: true` prop would display forward and backward buttons as disabled, while they'd still work when clicking.

This PR disables check for `pageInputDisabled` so the buttons' styling is not affected by it.

`disabled` prop disables next/previous page buttons so it can be used in parallel with `pageInputDisabled` if that functionality is needed.


asudoh:
@emyarod Could you review this, as the original code seems to have come from #1800? Thanks!

PR #2280, opened 2019-04-29
fix(NumberInput): avoid validation for space

### LINKED ISSUE #2106: NumberInput is no longer able to be controlled by a parent component
## Detailed description

Previously, NumberInput was ‘prop driven’. Prop copied into state and used, thus consistent.
Now, there is a check around min/max value - meaning the provided value from the parent may not be used.
It seems like the behaviour of `value` has changed to be more like `defaultValue` as per other carbon components. Where we use a parent component to drive the value from props, we are now getting inconsistent behaviour where the component is ignoring the props we give in. I.e when a user hits backspace to start typing a a new value, it automatically resets to the minimum value.
Should value be `defaultValue`, and never read again post mount?

NumberInput

Empty string should be able to be applied and accepted as an input without being overwritten with the minimum value

All

V6.97.2

See sandbox, hit backspace to try and edit input

> Please create a reduced test case in CodeSandbox

https://codesandbox.io/s/z62vy7y463

asudoh:
Fixes #2106.

#### Changelog

**Changed**

- `gDSFP` code in `<NumberInput>` to avoid `min` to be applied. Also added missing `max` support.


asudoh:
@emyarod Right - That's what `Math.max(min, value)`/`Math.min(max, value)` does.

emyarod:
@asudoh on the netlify preview when I manually edit the number to something that is invalid, the error state does not appear

asudoh:
@emyarod This issue/PR is about the way input (prop) from application code instead of user-input. Does this make sense to you? We can track your point with a separate issue.

PR #2281, opened 2019-04-29
chore(components): better name with forwardRef

asudoh:
This change updates React component name for various forward-ref'ed components from just `<ForwardRef>` to  something like `<ForwardRef(TextInput)>`.

Refs #1768.

#### Changelog

**Changed**

- Change the 1st arg of `React.forwardRef()` of various components to named functions.

#### Reviewing

Please use "Diff settings" - "Hide whitespace changes" for reviewing.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -18,99 +18,97 @@ const { prefix } = settings;
 
 let didWarnAboutDeprecation = false;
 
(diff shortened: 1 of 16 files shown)


vpicone:
Awesome! This has been a pain for quite a bit, thanks for taking it on.

Issue #1952, opened 2019-03-01
DropdownV2: Usability: support translatable "Open menu", "Close menu" text.


JessNah:
Hovering over the arrow icon on dropDownV2 displays text "Open menu"/"Close menu", would be nice if we could set this text ourselves, so that we can support translations.

dakahn:
Hi! We actually have full internationalization implemented but don't expose that prop in this component yet. Great catch and a much needed improvement. 👍

PR #2290, opened 2019-04-30
fix(DropdownV2): v6 - passes through props including translations to menu icon

ZacharyStair:
for v6 branch: allow menu icon within dropdown to be i18n-ized using menu icons props

~Closes IBM/carbon-components-react#1952~ I suppose this doesn't technically close it since there's still master and v5 but i'll open prs against those once we iron out this one

it was a tossup between passing `rest` which includes potentially a lot of unnecessary props vs enumerating the properties of the child component, this way is more resilient to future changes while leaving the DropdownV2 as uncluttered as possible

### DIFF
--- a/src/components/DropdownV2/DropdownV2.js
+++ b/src/components/DropdownV2/DropdownV2.js
@@ -153,6 +153,7 @@ export default class DropdownV2 extends React.Component {
       light,
       invalid,
       invalidText,
+      ...rest
     } = this.props;
     const inline = type === 'inline';
     const className = ({ isOpen }) =>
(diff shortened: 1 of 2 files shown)


ZacharyStair:
@asudoh @joshblack thoughts?

ZacharyStair:
i see in https://github.com/carbon-design-system/carbon-components-react/pull/2268 that the solution there was to pass through the same `translateWithId` function but would that still apply if the current component doesn't use one?

ZacharyStair:
closing this in favor of https://github.com/carbon-design-system/carbon-components-react/pull/2293

PR #2293, opened 2019-04-30
fix(DropdownV2): v6 - explicitly passes through translateWithId to menu icon

ZacharyStair:
*against v6 branch

so consumers of dropdownv2 can specify translations for title text of the icon

Related to https://github.com/carbon-design-system/carbon-components-react/issues/1952

updated version of my previous pr, i noticed that FilterableMultiSelect has a directly analogous instance of this so I figured I would just copy that pattern instead of making a new one.

### DIFF
--- a/src/components/DropdownV2/DropdownV2.js
+++ b/src/components/DropdownV2/DropdownV2.js
@@ -113,6 +113,11 @@ export default class DropdownV2 extends React.Component {
     titleText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
 
     /**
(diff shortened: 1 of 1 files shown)


ZacharyStair:
Worth noting that if a component defines a 'default' implementation of this function it will need to include all of the child component's strings. I think that is easy enough to accomplish, but it is a bit of development overhead to be aware of.

ZacharyStair:
@asudoh @joshblack this is a better version of the above closed pr, could you give it a review?

PR #2294, opened 2019-04-30
feat: update SideNav and Menu

vpicone:
- Use new PropTypes.elementType to allow for `Link` from third party libraries
- Make menu icons optional


vpicone:
PropType.node is not appropriate here because some valid nodes would not work (numbers).

PR #2299, opened 2019-04-30
fix(modal): screen reader improvements

### LINKED ISSUE #1659: Screen Reader Issues With Modal
## Detailed description

Looking at the [latest version of modal](https://deploy-preview-1657--carbon-components-react.netlify.com/?selectedKind=Modal&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel) there are several issues:

1. Escape does not exit the modal (I don't know if the exit modal is blocked on the example for some reason though)
2. The button does not close the modal (I don't know if the exit modal is blocked on the example for some reason though)
3. The close button is wrapped in a graphic that makes the screen reader announce: "Button Close the modal graphic Close the modal". The graphic should be hidden so the redundant information is not announced.
4. The "label" heading is level 4 and is above the "Modal heading" heading level 2. The "Label" should be heading level 1 or 2 (it doesn't matter), and the "modal heading" should be either the same or deeper than the "label" level.

It may be a good idea to have a button that triggers the modal as when a screen reader clicks on the "Modal" component, they are instantly placed in the modal. It is a little confusing.

1. Using NVDA go to the [Modal component](https://deploy-preview-1657--carbon-components-react.netlify.com/?selectedKind=Modal&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel)
2. Notice that focus is placed in the modal on page load
3. Hit ctrl+home and arrow down listening to the headings. The levels are out of order.
4. Hit escape
5. Arrow over the exit button and hear the redundant information.
6. hit enter on the exit button and nothing happens.

dakahn:
closes #1659 

We were inappropriately setting the `title` attribute which was causing confusing screen reader behavior (NVDA). We now use the same prop to set the `aria-label` which communicates clearly the close button's purpose while ignoring the SVG.

Also, we're now using more semantic markup to distinguish the label vs the modal header 👍 ♿️


src/components/Modal/Modal.js
```diff
@@ -305,7 +305,7 @@ export default class Modal extends Component {
         className={`${prefix}--modal-close`}
         type="button"
         onClick={onRequestClose}
-        title={iconDescription}
```
asudoh: Seems that we lost hover-over text on the close button. Probably we'd keep both?

emyarod: would it be better to keep the `title` on the button and add the `aria-label` attribute to the SVG itself instead of replacing the attribute entirely?

PR #2304, opened 2019-05-02
[WIP] - chore(storybook): add locale support + Farsi locales for components

theiliad:
#### Changelog
- Added locale support for storybooks
- Added `en` and `fa` locales
- Updated button stories to use locales

**New**
image

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -8,3 +8,4 @@ import '@storybook/addon-options/register';
 // Community addons
 import 'storybook-readme/register';
 import 'storybook-addon-rtl/register';
+import 'storybook-addon-intl/register';
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -5,7 +5,35 @@ import { withOptions } from '@storybook/addon-options';
 import { configureActions } from '@storybook/addon-actions';
 import { initializeRTL } from 'storybook-addon-rtl';
 // import { checkA11y } from 'storybook-addon-a11y';
+import { addLocaleData } from 'react-intl';
+import { setIntlConfig, withIntl } from 'storybook-addon-intl';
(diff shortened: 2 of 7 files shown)


theiliad:
Only removing the `draft` status to get a review. Josh is somewhat aware of the context here, but overall my goal here is to enhance our RTL storybooks by switching to Farsi translations in the individual stories.

Currently with the `rtl` plugin the storybooks are using, we're not able to detect a change in the status and also trigger a locale change, however that shouldn't be a problem. The 2 can function separately, also developing a custom RTL switcher could be a solution as well.

asudoh:
Hi 👋 @theiliad thank you for your contribution! For RTL stuffs we rely on build-time CSS conversion, e.g. replacing `left` with `right` and vice versa. Given Storybook RTL add-on does not seem to take care of that, I have created #2316 to remove that Storybook add-on, and more importantly, creates an environment to actually test the RTL stuffs.

Wrt translations themselves - Do you have your resource who actually does the translation? If so it'll be great and we can demonstrate the translated version. Otherwise, we need to rely on application teams to do the translation and thus our demo can only be of pseudo translations at the most.

Lastly, for general g11n stuff, please refer to: https://github.com/carbon-design-system/carbon-components-react/blob/v7.2.0/docs/g11n.md

Hope these give you some context on the current state of RTL stuffs. Don't hesitate to ping me if you have any questions. Thanks!

joshblack:
Hey @theiliad! 👋 We recently moved this project over to https://github.com/carbon-design-system/carbon ([Announcement](https://medium.com/carbondesign/carbon-is-moving-to-a-monorepo-c6bfcbe87de0)) Going to close this for now, but would you want us to help move this PR over to there?

PR #2308, opened 2019-05-02
fix(tag): add filter tag keyboard navigation

emyarod:
Closes IBM/carbon-components-react#2307

depends on a vanilla release that includes https://github.com/carbon-design-system/carbon-components/pull/2360

This PR adds keyboard navigation to the filter tag. When focused, the filter tag will appear as though it is hovered

#### Changelog

**New**

- `disabled` storybook knob
- filter tag story


emyarod:
superseded by https://github.com/carbon-design-system/carbon/pull/2654

Issue #2314, opened 2019-05-03
[Storybook] RTL feature on the Toggle component is messed up


camilasb:
## Detailed description

Steps:
1. Open the storybook
2. Click on Toggle component
3. Click on RTL sub-tab
4. Click on the RTL toggle - untoggled

Actual Result: Toggle label is hidden under the component
Expected Result: Toggle label must be visible

NOTE: all 3 types of toggle present problems

![image](https://user-images.githubusercontent.com/20360162/57157926-5e005180-6db8-11e9-9c37-06e077d308cb.png)

> What browser are you working in?
Seen on Chrome, Safari, FF

> What version of the Carbon Design System are you using?
Latest available at http://react.carbondesignsystem.com/

asudoh:
Hi 👋 thank you for reporting! You need to get RTL version of the CSS following the instruction at: https://github.com/carbon-design-system/carbon-components/blob/v10.2.0/docs/g11n.md - Once I did that I see the problem reported here doesn't happen.

Issue #2286, opened 2019-04-30
Pagination: Need better RtoL support for the pagination component


scottdickerson:
### Summary

If you enable the RtoL support in the storybook story for Pagination you can see that there's lots of overlap and incorrect buttons when you flip to support internationalized languages that go right to left. Screenshot attached below

### Justification

Need to support Arabic, Hebrew and other RtoL languages

### Desired UX and success metrics

No overlap, button layout that makes sense in the reversed mode

![image](https://user-images.githubusercontent.com/6663002/56945688-4743cb80-6b28-11e9-9bfd-525c4f3233ac.png)

emyarod:
@IBM/carbon-designers do we have RTL spec for this component (and others)?

asudoh:
Ref: https://github.com/carbon-design-system/carbon-components/blob/v10.2.0/docs/g11n.md - Not sure if RTL conversion of the CSS has been done when above screenshot is made.

dakahn:
Not sure about priority set on this one. Technically those components are broken for right to left users, but are we saying that the onus is on the devs implementing these components to handle right to left support in their apps @asudoh?

I can swap this to a higher priority if this is actually a bug

asudoh:
@dakahn Support for RTL requires an additional build step mentioned in https://github.com/carbon-design-system/carbon-components/blob/v10.2.0/docs/g11n.md. Not sure if the person who introduced RTL knob to our Storybook env was aware of such additional build step. Otherwise, we should remove the RTL knob for now because it causes a confusion (of broken RTL support).

Issue #2053, opened 2019-03-19
[Storybook] Add an RTL (Right-to-Left) toggle or feature for components


camilasb:
### Summary

Since IBM is world-wide and we have to be compliant with many languages including bi-di, it would be nice to have an RTL (Right-to-Left) toggle or feature so we could see how the components would behave on a bi-di env.

### Justification

We're working on a project to be released on November (as of now) of a really mature and multi-language product which will start using Carbon React components now. We're having many doubts about how the components must behave on such language environments and we don't know what would be the correct look and feel.

### Desired UX and success metrics

In our own storybook we are adding this toggle, but sometimes we cannot know if the behavior is correct, since the original Carbon storybook doesn't provide that info for us.
![Screen Shot 2019-03-19 at 1 26 47 PM](https://user-images.githubusercontent.com/20360162/54623649-b5d23b80-4a4a-11e9-8e69-b6576173bb4a.png)

In the screenshot above, we don't know if the checkboxes should be on the right or left side of the label.

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an upcoming release?

As stated above, we're targeting end of year, so if you guys have any input for us, would be great!

### Available extra resources

What resources do you have to assist this effort?

Don't know how many or which resources we would be able to provide, but we can discuss it internally if you're up to help us =)

asudoh:
@vpicone Would you be interested in adding something like https://www.npmjs.com/package/storybook-addon-rtl to our Storybook environment? Thanks!

vpicone:
@asudoh totally!

PR #2316, opened 2019-05-03
chore(Storybook): remove RTL plugin

asudoh:
This change removes RTL plugin of Storybook. Though it's a nice add-on, we require an additional build step to generate RTL version of CSS and that Storybook add-on does not seem to kick off an altern ate build.

Refs #2053.
Refs #2314.
Refs #2286.

#### Changelog

**New**

- New environment variable in our Storybook env, `CARBON_REACT_STORYBOOK_USE_RTL`. (Updated our contribution doc for that)

**Removed**

- RTL Storybook plugin.
- `CARBON_USE_EXPERIMENTAL_FEATURES`/`CARBON_USE_BREAKING_CHANGES` mentions in our contribution doc given we no longer use it.


dakahn:
Great! 🎈

asudoh:
@emyarod Good question - As of today, those feature flags are gone. We'll introduce new feature flags as necessary, my bet is one per one breaking change.

PR #2318, opened 2019-05-06
fix(tabs): replace nav with div tags

### LINKED ISSUE #2171: Tabs component's tablist does not need to be in a nav
It is not necessary to have the tabs of a Tabs component inside of a `<nav role="navigation" ...>`. 
I could not find another "Tabs/Tabfolder" component that wraps its tablist in a nav, so it is uncommon, and users would not expect it.
It results in extra verbosity when entering the Tabs component using a screen reader.
If there is another `<nav>` on the page (most pages have at least one), then the Tabs nav would need to have an app-specific aria-label (i.e. another prop) in order to tell the navs apart.
The tab items behave more like buttons (and are often implemented with buttons) than links, and so don't really belong in a nav.

I think it would be cleaner to just delete the wrapper nav.

eldain:
Closes #2171

Replace the `nav` tags with `div` tags. We want to reduce verbosity when using a screenreader and multiple `nav` elements are on the same page.

#### Changelog
**Changed**

- replace `nav` tags with `div` tags


asudoh:
@eldain Could you please update the snapshot? Thanks!

eldain:
@asudoh Snapshot updated! Happy to make any additional changes if necessary, especially if the Enzyme selectors aren't the preferred style

carmacleod:
Thanks, @eldain !  :tada:

Issue #1635, opened 2018-12-11
Accessibility: the labelText prop for DatePickerInput should be required


jendowns:
## Detailed description

The `labelText` prop for the DatePickerInput is currently not required: https://github.com/IBM/carbon-components-react/blob/master/src/components/DatePickerInput/DatePickerInput.js#L27

End users are therefore able to use `DatePicker` and the child `DatePickerInput` without providing any assistive text.

So you can have a situation like in the range variation of the `DatePicker` in Carbon React's storybook, where there are no labels provided to describe what the form inputs do: http://react.carbondesignsystem.com/?selectedKind=DatePicker&selectedStory=range%20with%20calendar%20and%20min%2Fmax%20dates

i.e. -- 
```html
<DatePicker
  id="date-picker"
  onChange={anonymous}
  minDate="1/10/2020"
  maxDate="1/20/2020"
  datePickerType="range"
>
   <DatePickerInput id="date-picker-input-id" light={false} onChange={anonymous} />
   <DatePickerInput id="date-picker-input-id-2" light={false} onChange={anonymous} />
</DatePicker>
```

For some reason, the `a11y` addon isn't flagging this component variation in Carbon React's storybook (perhaps due to the way the props object is passed in, or different storybook versions?), but the `a11y` addon *does* flag it in our environment:
```
#date-picker-input-id
    ✘ aria-label attribute does not exist or is empty
    ✘  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible
    ✘ Form element does not have an implicit (wrapped) <label>
    ✘ Form element does not have an explicit <label>
    ✘ Element has no title attribute or the title attribute is empty

#date-picker-input-id-2
    ✘ aria-label attribute does not exist or is empty
    ✘ aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible
    ✘ Form element does not have an implicit (wrapped) <label>
    ✘ Form element does not have an explicit <label>
    ✘ Element has no title attribute or the title attribute is empty
```

Inputs need to have labels to go with them: https://www.w3.org/WAI/tutorials/forms/labels/
Even if the label text is implied, there needs to be some assistive text present (even if it's visually hidden).

By comparison, the `labelText` prop for the `TextInput` component is required: https://github.com/IBM/carbon-components-react/blob/master/src/components/TextInput/TextInput.js#L111

i.e. - 
```
  /**
[...]

PR #2319, opened 2019-05-06
fix(date-picker-input): require label

jendowns:
Closes IBM/carbon-components-react#1635

Require `labelText` prop for `DatePickerInput` for accessibility purposes.

#### Changelog

**Changed**

- made `labelText` a required prop (made it a `node` type like `TextInput` counterpart: https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/TextInput/TextInput.js#L135)
- removed `labelText` default prop


src/components/DatePickerInput/DatePickerInput.js
```diff
@@ -24,14 +24,18 @@ export default class DatePickerInput extends Component {
      * The description of the calendar icon.
      */
     iconDescription: PropTypes.string,
+    /**
```
asudoh:
```suggestion

    /**
```

jendowns:
Thanks @asudoh! I just committed your suggestion 👍

PR #2320, opened 2019-05-06
feat: add carbon-context component

vpicone:
closes https://github.com/carbon-design-system/carbon-components-react/issues/2321
relevant: #1575

CarbonContext offers utilities to  allows the user to provide information
to carbon components throughout their application.

This is meant to offer an easier interface for testing and using experimental/optional features in a `carbon-components-react` application.

**Possible uses:** breaking changes, style variations, I18n, or experimental components.

***

#### `CarbonProvider` 
This allows users to enable flags at the root of their App. All the flags are disabled by default
```jsx
import { CarbonProvider } from 'carbon-components-react';

const Root = (
  <CarbonProvider flags={['UIShell']}>
    <App />
  </CarbonProvider>
);

ReactDOM.render(Root, document.getElementById('root'));
```

#### `useCarbonFlag` 

A React hook that takes the name of a flag

```jsx
import { useCarbonFlag } from 'carbon-components-react';

const Component = () => {
  const experimenting = useCarbonFlag('UIShell');
  return experimenting ? <ExperimentalUIShell /> : <UIShell />;
};
```

#### `CarbonContext`

This allows us to use context in Class components as well

```jsx
import { CarbonContext } from 'carbon-components-react';
class ClassComponent extends React.Component {
  componentDidMount() {
    let experimenting = this.context;
    /* perform a side-effect at mount using the value of MyContext */
  }
  render() {
    let experimenting = this.context;
    /* render something based on the value of MyContext */
  }
}

MyClass.contextType = CarbonContext;
```


asudoh:
Hi @vpicone thank you for gathering a thought! The current feature flag mechanism is created to avoid doubling Carbon code in app bundle, which means compile-time switch. I think your suggestion is run-time switch, right?

vpicone:
@asudoh That's correct, the advantage of doing a run-time swap is that dev's could enable flags for certain regions/pages of their application and not for others. There is a risk of duplicating code, but I believe this additional feature would be worth the marginal increase in bundle size as we test out new components or features. It also depends on how the flag is used (rendering an entire component vs enabling a specific prop/feature).

It also means not needing to eject applications or customizing webpack configs in order to use experimental features. I believe this would smooth  and increase adoption/testing of new features.

asudoh:
@vpicone I see. The thoughts around easier adoption of new features as well as adoption to specific portions in application are good and plausible. Having those in mind back then, we made a conscious decision that non-production code shouldn't affect application's bundle/runtime. Another strategy we had during `v9`/`v10` concurrent development was that we strived to separate `v9`/`v10` code rather than intermingling them wherever it makes sense, for easier code reviews/analysis (by non-authors) and avoid introducing bugs.

The code in this PR is close to an option we had for CSS prefix, though. And I think it's still viable for that area.

vpicone:
@asudoh the context could be accessed at any level, selectively importing or exporting entire components or making granular changes. the `useCarbonFlag('breakingChangesX)` would be used the same as breakingChangesX except with the advantages you mentioned included.

asudoh:
@vpicone From your series of responses, I have a feeling that we should rename the mechanism in this PR here, something that tells "global options". Wanted to make sure that what you created is used for global options that user can easily changed back-and-forth, and feature flags are used for big changes (e.g. `v9`/`v10` migration).

vpicone:
@asudoh The idea is to accommodate both of those use cases so that users could enable breaking changes (v11+ migration) not just per component, but also per region or page of their application.

For example, if we make breaking changes to the DataTable component in v11, they could write any new DataTables using the v11 version by developing under a Carbon Context Provider with the `DataTablev11` flag enabled. This means they can try out the new breaking changes _now_ while not needing to re-write their v10 DataTables.

asudoh:
Let's get back to the earlier discussion - Does the new mechanism introduced here support tree-shaking?

vpicone:
@asudoh My idea is that [React.lazy()](https://reactjs.org/docs/code-splitting.html#reactlazy) could be used to dynamically load bundles containing the experimental or v11 components. Note: `React.lazy` doesn't work for server side rendered applications yet, but there are libraries that do: [loadable-components](https://github.com/smooth-code/loadable-components).

If that fails, I'd be curious to hear whether you think the advantages I mentioned (no custom webpack, per-component, and per-app-region flags) would be worth increasing the size of some components pre-release.

asudoh:
Thank you for your thoughts @vpicone - Seems that `React.lazy()` is for dynamic imports which comes with additional HTTP transaction (tends to incur more cost than byte size). My perspective is that the features for easier development like no WebPack customization and (easier) granular application of feature flags won't outweigh the experience applications' end-users get (less time/energy spent on loading/parsing/running the code).

That said, another option on this problem space is similar to what we took in the past, which is, shipping new version as a different component. In this way, tree-shaking works without customizing flags.

vpicone:
The new component version was by many accounts abroad we didn’t want to pursue again. It also doesn’t allow for small,  component level changes like prop depreciations or any of the other uses mentioned in the original comment.

vpicone:
After talking with @asudoh I think releasing v11-alpha's more often with just the feature flagged enabled might be a better solution for experimental components. I'm gonna close this until we have more time to work on the intricacies of the other applications (focus management, style variations, I18n).

Issue #2328, opened 2019-05-07
MultiSelect.Filterable doesn’t support `useTitleInItem` (tooltip on hover)


hyungskim:
MultiSelect.Filterable from carbon-component-react doesn’t support `useTitleInItem` (tooltip on hover). It will show tooltip fine if we don’t use `Filterable`  option but this Filterable option is required for our UI search functionality. Can you add `useTitleInItem` functionality? It’s very crucial for Watson Knowledge Catalog UI

Version info:
carbon-components-react@6.97.0
carbon-components@9.79.0

<img width="272" alt="Screen Shot 2019-05-07 at 8 15 05 AM" src="https://user-images.githubusercontent.com/50377275/57320599-1a596080-70b4-11e9-9e08-c4dcdb6e02fb.png">

asudoh:
Hi 👋 thank you for reporting! If you search for `useTitleInItem` in our `<MultiSelect>` code, you'll find that it simply translated to our `<Checkbox>`: https://github.com/carbon-design-system/carbon-components-react/blob/v7.1.0/src/components/MultiSelect/MultiSelect.js#L300. You can try doing so in your fork, and if it works well and you'll see it valuable, please be sure to contribute it back!

hyungskim:
Hi @asudoh thanks for responding quick.  Yeah, I've noticed `useTitleInItem` in MultiSelect.js but not in FilterableMultiSelect.js.  FilterableMultiSelect.js seems to be the actual source code where we need `useTitleInItem`.

asudoh:
You are right - Would you try applying the `<MultiSelect>` code to `<MultiSelect.Filterable>`? Thanks!

hyungskim:
I love to apply my code change but I'm getting access denied error when trying to push my changes.  Can I request access?  

<img width="626" alt="Screen Shot 2019-05-07 at 7 07 35 PM" src="https://user-images.githubusercontent.com/50377275/57344351-c2dee300-70fb-11e9-9cfa-88fdb694b28c.png">

asudoh:
Hi @hyungskim the process is creating your own fork and submit a PR based on that. Please see https://github.com/carbon-design-system/carbon-components-react/blob/v7.2.0/.github/CONTRIBUTING.md#setup for more details. Thanks!

PR #2330, opened 2019-05-08
Add support for useTitleInItem (tooltip on hover)

hyungskim:
- MultiSelect.Filterable doesn’t support `useTitleInItem` (tooltip on hover) #2328
- https://github.com/carbon-design-system/carbon-components-react/issues/2328


vpicone:
What about itemTitleText for parity with the titleText prop?

asudoh:
@vpicone I may be missing something, but any reference on `itemTitleText`?

vpicone:
I was offering it as a naming alternative that more closely aligns with the parent titleText. I’m on my phone, so if that name doesn’t make sense in this context that’s okay too!

asudoh:
@vpicone OK the prop name is for the parity with `<MultiSelect>` and changing the name altogether is a breaking change. We can consider the name change in future.

hyungskim:
Hi @asudoh, do you the published version of carbon-component-react that contains this fix?

asudoh:
Hi 👋 @hyungskim we recently switched to monthly-ish release cycle. Stay tuned!

hyungskim:
@asudoh Hi, do you know which version has my fix? Looks like our production env still doesn't have this fix.

asudoh:
Hi 👋 @hyungskim if it's the case, this PR may have slipped from our monorepo move. Would you mind creating an PR at https://github.com/carbon-design-system/carbon/pulls? Thanks!

hyungskim:
@asudoh Looks like my fix is in there for MultiSelect.Filterable by checking at this sample but not sure which published version this one uses.

Screen Shot 2019-06-25 at 8 17 43 AM

emyarod:
@hyungskim looks like this was merged into v10.3.2 https://github.com/carbon-design-system/carbon/commit/33795afae9bd24babe3fc49de3102e29c0d69d60#diff-e8c89a1867f093cba993a4c0fd1eb37d

hyungskim:
@emyarod @asudoh Hmm, that makes sense as our production currently uses carbon 9 (v6.xxx). It would take some time to migrate to carbon 10 (v.10.x).  Do you think I should create a PR for carbon 9 but not sure which branch I should use?

emyarod:
@hyungskim yeah you can target https://github.com/carbon-design-system/carbon-components-react/tree/v6

hyungskim:
@emyarod Do I need a permission to push my change?  I didn't have this issue using master but I'm having an issue right now with v6. 

Screen Shot 2019-06-25 at 1 59 04 PM

PR #2333, opened 2019-05-08
feat(PresentationalNumberInput): introduction

asudoh:
`<PresentationalNumberInput>` is presentational-only version of `<NumberInput>` to let application handle its state rather than handling it internally.

Refs #2207.

#### Changelog

**New**

- `<PresentationalNumberInput>`, presentational-only version of `<NumberInput>` to let application handle its state rather than handling it internally.

**Changed**

- `<NumberInput>` to use `<PresentationalNumberInput>` internally.

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -6,13 +6,12 @@
  */
 
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


asudoh:
Superseded by: carbon-design-system/carbon-components#2427

PR #2338, opened 2019-05-08
fix(code-snippet): update skeleton story info and spacing

jendowns:
The `style` wasn't really adding any space between the 2 skeleton components in the `CodeSnippetSkeleton` story, so I removed it and added a `<br />` instead.

Also noticed the `info` was missing some text, so added that.

#### Changelog

**Changed**

- remove `style` from `CodeSnippetSkeleton`, use `<br />` instead
- update `CodeSnippetSkeleton` info.


dakahn:
👋 Hello

I'm on react.carbondesignsystem and I can't seem to see that styling bug. I'm seeing the 8px margin being applied (at cursor). But I could totally just be missing something.

2019-05-08 15_53_43-Carbon Components React

jendowns:
Hey @dakahn! I think my wording was a little ambiguous -- there's a padding being applied, not a margin. As a result, there's no "visual" spacing that I think was originally intended here?

dakahn:
Ah, I see. Should we add a margin attribute instead of the less-semantic `<br>` element? Could be easier to debug in the future.

jendowns:
@dakahn Update is in 👍

PR #2342, opened 2019-05-08
fix(Tooltip): remove ARIA attribute

emyarod:
Closes IBM/carbon-components-react#2335

This PR removes the `aria-labelledby` attribute on the tooltip to avoid a DAP violation


asudoh:
@emyarod I think you are still working on this, is it correct?

emyarod:
@asudoh wanted to confirm first with the issue creator if this addresses the issue

closed and superseded by https://github.com/carbon-design-system/carbon/pull/2656

Issue #2344, opened 2019-05-08
Radio button selection in the Carbon data table should support table actions


emyarod:
[Originally reported by Tom Ginader in IBM Slack](https://ibm-cloudplatform.slack.com/archives/C2K6RFJ1G/p1556634610314900):

>Hello. How would I use radio buttons in DataTable to have actions? I was able to get regular selection with batch actions working properly, but we realized that we only want our users to select one item at a time. I'm not seeing any examples that combines radio button selection with some kind of action.

asudoh:
Reduced case: https://7142j221yq.codesandbox.io/

More details describing what the issue is about:

> You can see that clicking the checkbox triggers the batch action options, but clicking on a radio button does not.

PR #2347, opened 2019-05-08
feat(DataTable): support radio button row actions

emyarod:
Closes IBM/carbon-components-react#2344

#### Changelog

**New**

- support batch table actions for radio button row selection in data table
- storybook example

### DIFF
--- a/src/components/DataTable/DataTable.js
+++ b/src/components/DataTable/DataTable.js
@@ -389,7 +389,7 @@ export default class DataTable extends React.Component {
           return p;
         }, {});
         return {
-          shouldShowBatchActions: false,
+          shouldShowBatchActions: !row.isSelected || selectedRowsCount > 0,
           rowsById: {
             ...rowsById,
             [rowId]: {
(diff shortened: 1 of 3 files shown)


src/components/DataTable/DataTable.js
```diff
@@ -389,7 +389,7 @@ export default class DataTable extends React.Component {
           return p;
         }, {});
         return {
-          shouldShowBatchActions: false,
+          shouldShowBatchActions: !row.isSelected || selectedRowsCount > 0,
```
asudoh: Seems that what `handleOnSelectRow()` returns is almost identical between radio/checkbox with your change. Would you be able to try reducing the duplicated code? Also it appears that the code tries to access `selectedRowsCount` before it's defined.

Also, just found that the existing code (above `reduce()` logic) seems to mutate an object, is it true? If so, we should create a clone with the new `isSelected` prop given the nature of React (where its change detection logic relies on immutable objects).

emyarod:
superseded by https://github.com/carbon-design-system/carbon/pull/2657

PR #2349, opened 2019-05-09
fix(tag): update reduce method in story

jendowns:
Small thing --

I wanted to suggest updating the first argument in the `reduce` method here from `o` to `accumulator`. I think that in a case where there's nothing relevant to name the accumulator, it would be best to just use a default / recognizable word to help with readability. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Syntax

#### Changelog

**Changed**

- change first argument of `reduce` method from `o` to `accumulator`

### DIFF
--- a/src/components/Tag/Tag-story.js
+++ b/src/components/Tag/Tag-story.js
@@ -12,8 +12,8 @@ import Tag, { types as typesList } from '../Tag';
 import TagSkeleton from '../Tag/Tag.Skeleton';
 
 const types = typesList.reduce(
-  (o, type) => ({
-    ...o,
+  (accumulator, type) => ({
+    ...accumulator,
     [`${type} (${type})`]: type,
   }),
   {}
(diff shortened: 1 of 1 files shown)


asudoh:
Heads-up @jendowns - The repo has been merged to the vanilla repo. Good to have your future contributions there - Thanks!

src/components/Tag/Tag-story.js
```diff
@@ -12,8 +12,8 @@ import Tag, { types as typesList } from '../Tag';
 import TagSkeleton from '../Tag/Tag.Skeleton';
 
 const types = typesList.reduce(
-  (o, type) => ({
-    ...o,
+  (accumulator, type) => ({
```
joshblack: Super super, uber, mega nitpicky, but we tend to abbreviate as `acc`:

```suggestion
  (acc, type) => ({
```

jendowns: No problem @joshblack! I just committed your change 👍

jendowns:
Hey it looks like the build for my latest commit was cancelled as a redundant build. Is anyone able to restart it for me? 😅

asudoh:
Superseded by carbon-design-system/carbon#2683 - @jendowns Our apologies for the hustle!

PR #2359, opened 2019-05-10
fix: added the legend to TileGroup, resolves #2358

### LINKED ISSUE #2358: TileGroup never renders the legend attribute
Look at the TileGroup example here:  http://react.carbondesignsystem.com/?selectedKind=Tile&selectedStory=Selectable&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

The example has a `legend` attribute with a value of `Selectable Tile Group`, but that attribute is never added to the `fieldset` element for the TileGroup.  This means there's no way to set a legend for the TileGroup and that results in the following accessibility issue:

Screen Shot 2019-05-10 at 7 40 26 AM

zgrossbart:
Closes IBM/carbon-components-react#2358

Adds the legend attribute to the TileGroup if it was set by the calling code.

### DIFF
--- a/src/components/TileGroup/TileGroup.js
+++ b/src/components/TileGroup/TileGroup.js
@@ -111,6 +111,12 @@ export default class TileGroup extends React.Component {
     }
   };
 
+  renderLegend = legend => {
+    if (legend) {
+      return <legend>{legend}</legend>;
+    }
+  };
+
   render() {
     const {
       disabled,
@@ -119,7 +125,8 @@ export default class TileGroup extends React.Component {
     } = this.props;
 
     return (
-      <fieldset className={className} disabled={disabled} legend={legend}>
+      <fieldset className={className} disabled={disabled}>
+        {this.renderLegend(legend)}
         {this.getRadioTiles()}
       </fieldset>
     );
--- a/src/components/TileGroup/TileGroup.js
+++ b/src/components/TileGroup/TileGroup.js
@@ -107,10 +107,14 @@ export default class TileGroup extends React.Component {
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


zgrossbart:
I added the `legend` property explicitly to the property list.

joshblack:
Is the DAP violation for the `legend` attribute, or the `<legend>` tag? Had a hard time finding the attribute on MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#Attributes

src/components/TileGroup/TileGroup.js
```diff
@@ -106,11 +111,22 @@ export default class TileGroup extends React.Component {
     }
   };
 
+  renderLegend = legend => {
```
joshblack: Would we need to render `null` to prevent the: "Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null." error? Or is this gone in newer versions of React?

src/components/TileGroup/TileGroup.js
```diff
@@ -106,11 +111,22 @@ export default class TileGroup extends React.Component {
...
+    if (legend) {
+      return <legend>{legend}</legend>;
```
joshblack: Does this node need to be styled at all?

zgrossbart: This will inherit the style from the parent element.

src/components/TileGroup/TileGroup.js
zgrossbart: The newer version of React is happy with this.  It doesn't show any errors if the legend isn't there.

PR #2365, opened 2019-05-13
feat: lazy loading DataTable

areddon:
The current implementation of the `DataTable` component requires that you have all of the data up front and is problematic when display large lists of data in a table. The `LazyDataTable` allows you to display the first set of rows in the table and then update the rows when the user interacts with the header (sorting) or pagination options.

#### Changelog

**New**

- `LazyDataTable` component with `onHeaderChange` and `onPaginationChange` properties to update the table row data.

**Changed**

- Update require React version to 16.8 as the component uses hooks.


asudoh:
Hi 👋 thank you for considering the contribution! First, please be advised that this repo has been merged to https://github.com/carbon-design-system/carbon repository.

Given `<DataTable>` is involves non-trivial amount of code, I personally would be more comfortable the code is maintained outside core Carbon repo, especially considering our maintenance effort. I think we can definitely advise the lazy version if you are an IBMer, have a GH repo and gives us a link!

areddon:
This component is essentially the existing `DataTable` component with two extra props, I could see it being a replacement.

The `-addons` libraries which I use and contribute to are not public, and this component should be shared with others. They also share the same base in `carbon-components-react`, so it makes sense to contribute back so there is no code duplication and I do not think it makes sense to have a single repository for a single component.

dakahn:
@areddon can you re-open this PR on our /carbon repo here so we can continue the conversation?
https://github.com/carbon-design-system/carbon

areddon:
I would like to put this into version 6 of `carbon-components-react` which is not available in in the Carbon monorepo.

dakahn:
@areddon please see @asudoh comment above about maintaining this outside the core Carbon repo 👍

PR #2372, opened 2019-05-14
fix(downshift): allow passing props to Downshift

art0rz:
This change will allow users to pass additional props to Downshift. This is useful for when for example the app is running in a shadowRoot and the user needs to pass an `environment` to Downshift (which is an issue I'm currently running into)

See downshift-js/downshift#536

#### Changelog

**New**

**Changed**

The following files have the same changes applied to them:
- Combobox.js
- Dropdown.js
- FilterableMultiselect.js
- Multiselect

Added optional prop `downshiftProps` which is merged with component's own Downshift props (component's props take precedence) and passed to Downshift.

**Removed**

### DIFF
--- a/src/components/ComboBox/ComboBox.js
+++ b/src/components/ComboBox/ComboBox.js
@@ -146,6 +146,11 @@ export default class ComboBox extends React.Component {
      * should use "light theme" (white background)?
      */
     light: PropTypes.bool,
+
+    /**
+     * Additional props passed to Downshift
+     */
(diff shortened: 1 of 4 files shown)


joshblack:
Hi there! This project has been moved over to: https://github.com/carbon-design-system/carbon

More details: https://medium.com/carbondesign/carbon-is-moving-to-a-monorepo-c6bfcbe87de0

dakahn:
I'll close this down, but would you down to re-open in our /carbon repo? @art0rz

PR #2392, opened 2019-06-10
fix(Modal): allow differnt roles for modal

Zhao-Yi-Peng:
Issue:
- Current modal component has a accessibility checkpoint violation. `All content must reside within a WAI-ARIA landmark or labelled region role`
- The role "presentation" in the modal is not a WAI-ARIA role.

Details in: https://aat.w3ibm.mybluemix.net/token/98e2765a-de76-49d2-9f5c-605778089422/b7118222-21ab-441d-9037-25deacdefdf0/archives/2018AugustDeploy/doc/w3/help/en-US/idhi_accessibility_check_g1157.html

Solution:
- adding prop that allow modification of the role of the container so no more violation in the content of the modal

### DIFF
--- a/src/components/ComposedModal/ComposedModal.js
+++ b/src/components/ComposedModal/ComposedModal.js
@@ -28,6 +28,7 @@ export default class ComposedModal extends Component {
   static defaultProps = {
     onKeyDown: () => {},
     selectorPrimaryFocus: '[data-modal-primary-focus]',
+    role: 'region',
   };
 
   outerModal = React.createRef();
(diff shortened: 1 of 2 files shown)


psachs21:
@asudoh is there a schedule for when v6 releases will be done? Just waiting on this fix and wanted to know when to expect it.

emyarod:
v6.115.4 is now available https://github.com/carbon-design-system/carbon-components-react/releases/tag/v6.115.4

PR #2397, opened 2019-06-25
Add support for useTitleInItem (tooltip on hover) for carbon 9 (v6)

hyungskim:
- MultiSelect.Filterable doesn’t support `useTitleInItem` (tooltip on hover) #2328
- https://github.com/carbon-design-system/carbon-components-react/issues/2328

Ref: https://github.com/carbon-design-system/carbon-components-react/pull/2330


hyungskim:
Question. When do I expect to see my change? In 1st week of July or 1st week of August?

emyarod:
I'm cutting a release today that will include some v9 changes from vanilla as well

emyarod:
@hyungskim should be in https://github.com/carbon-design-system/carbon-components-react/releases/tag/v6.115.5

PR #2399, opened 2019-07-05
fix(TextArea): add ref forwarding to TextArea

webadam1:
I've added ref forwarding to the TextArea component the same way it's done in the TextInput component.

### DIFF
--- a/src/components/TextArea/TextArea.js
+++ b/src/components/TextArea/TextArea.js
@@ -13,19 +13,22 @@ import WarningFilled16 from '@carbon/icons-react/lib/warning--filled/16';
 
 const { prefix } = settings;
 
-const TextArea = ({
-  className,
-  id,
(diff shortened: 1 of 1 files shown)


emyarod:
just one thing, would you mind setting the `displayName` to be `TextArea`? similar to our other components that are forwarding refs

webadam1:
Added displayName 👍

emyarod:
@webadam1 did you need this for `v6`? if so we can revert this and open the PR to `v6` branch instead

webadam1:
@emyarod No, I only need this for v10 Carbon, so I think I don't need it in the v6 branch.

emyarod:
@webadam1 gotcha we can revert it in this legacy repo and merge it into `v6` branch since v10 changes are all contained in `@carbon/react`

PR #2401, opened 2019-07-25
fix(SelectableTile): remove console warnings

emyarod:
Closes https://github.com/carbon-design-system/carbon/issues/3537

ref: https://github.com/carbon-design-system/carbon/issues/2959 https://github.com/carbon-design-system/carbon/pull/3431

This PR removes console warnings created by invalid props and invalid prop types in the `<SelectableTile>`

#### Testing/Reviewing

Ensure no console errors are emitted when viewing the component


emyarod:
do you mean we should instead be setting `readOnly` to these by default?

PR #2403, opened 2019-08-01
feat(Notification): add icon prop for notification

webadam1:
The following issues are addressed in this PR:

- Notification icons are not customizable in any way. Now it's possible to pass an icon as an elementType prop, which will overwrite the default icon that's based on the notification kind. (if the 'icon' prop is unset, the functionality and look of the component remains the same)
- The 'iconDescription' prop was mistakenly supplied both to the close icon and the decorative left-side icon. Now these are separated into iconDescription for the left side decorative icon and closeIconDescription for the close icon.

### DIFF
--- a/src/components/Notification/Notification.js
+++ b/src/components/Notification/Notification.js
@@ -186,8 +186,13 @@ const useIcon = kindProp =>
     warning: WarningFilled20,
   }[kindProp]);
 
-const NotificationIcon = ({ notificationType, kind, iconDescription }) => {
-  const NotificationIconX = useIcon(kind);
+const NotificationIcon = ({
+  icon,
+  notificationType,
+  kind,
+  iconDescription,
+}) => {
+  const NotificationIconX = icon || useIcon(kind);
   return !NotificationIconX ? null : (
     <NotificationIconX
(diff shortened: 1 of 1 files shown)


webadam1:
The ci/circleci check seems to fail because the linter does not recognise PropTypes.elementType. Could this be because it's using an older version of eslint-plugin-react? Locally the linter runs without errors.

joshblack:
Hi @webadam1! 👋 Thanks so much for taking the time to make this PR, unfortunately this codebase is in maintenance mode and is only accepting bugfixes as a result. We can definitely look into adding the fix for `iconDescription`, though 👍

webadam1:
Allrigth, clear as day, thanks for the clarification @joshblack!
Should I close this PR and open another one for the iconDescription fix, or should I change the code this one to only include the fix?

joshblack:
Feel free to update! Definitely don't want to make things harder on your end 😄

asudoh:
BTW given this PR seems to be against `master` - Please be advised that the `master` equivalent (latest `10.x`) is maintained in https://github.com/carbon-design-system/carbon repo instead of this repo.

PR #2410, opened 2019-08-29
fix: pass parent aria-label to <ListBox.Field>

emyarod:
Closes https://github.com/carbon-design-system/carbon/issues/3481

Typically, downshift sets a default `"open menu"` or `"close menu"` aria-label. This PR passes the listbox `aria-label` or `ariaLabel` attribute into `<ListBox.Field>`

#### Testing / Reviewing

Ensure the listbox components (`<ComboBox>`, `<DropdownV2>`, `<MultiSelect>`, `<MultiSelect.Filterable>` meet the WCAG 2.1 guidelines for aria-labels


tw15egan:
For some reason, I'm seeing two `aria label` attributes when testing. When I pass `aria-label="testing"`, I get:

Screen Shot 2019-08-29 at 10 45 59 AM

When I send in `ariaLabel`, I'm getting

Screen Shot 2019-08-29 at 10 45 08 AM


🤔

emyarod:
should be resolved now @tw15egan

emyarod:
should be resolved now @tw15egan

emyarod:
Downshift adds `aria-label` to `<ListBox.Field>`, and by default its values are `'open menu'` and `'close menu'`. the `ariaLabel` prop is only added to the `<input>` if I'm not mistaken

asudoh:
Thanks @emyarod for clarifying. What Downshift does wrt `aria-label` seems simply adding `aria-label` to what `getButtonProps()` returns. That said, would we want to override that with the `ariaLabel` we have, instead of providing `aria-label` to `getButtonProps()` arg?

Basically my concern is only of adding another props of the same purpose. So an alternative is deprecating the original `ariaLabel`.

emyarod:
for this repo I think overriding Downshift's `aria-label` is fine but it might be better to deprecate `ariaLabel` in v10, what do you think @asudoh ?

asudoh:
My apologies for being late on this @emyarod (should have checked this repo more frequently) - Sounds that `ariaLabel` is the way to go.

emyarod:
@asudoh in that case do we want to add `ariaLabel` prop to the remaining listbox components? for example filterable multiselect currently does not support that prop

asudoh:
Sorry @emyarod my mistake, I meant removing `ariaLabel`.

emyarod:
@asudoh will that be considered breaking change for the components that currently support it?

asudoh:
Sorry again @emyarod - "Deprecating" to be more accurate.

PR #2415, opened 2019-09-11
fix(TileGroup): Wrap RadioTiles inside TileGroup

jrpomeroy:
This adds a wrapper div element around the RadioTiles inside the TileGroup component so that a flex layout can be applied to the tiles to make them render horizontally.

Already merged into latest master. See https://github.com/carbon-design-system/carbon/pull/3947


jrpomeroy:
@asudoh I'm hoping we can get this TileGroup fix into v6, where we really need it right now.  Can you take a look?

PR #2421, opened 2019-10-08
fix(Loading): use description prop as SVG title

emyarod:
Closes https://github.com/carbon-design-system/carbon/issues/4194 (backport of https://github.com/carbon-design-system/carbon/pull/2955)

This PR passes the `description` prop to the `<Loading>` component SVG title

#### Testing / Reviewing

Ensure the SVG title reflects the `iconDescription` prop in `<InlineLoading>` and the `description` prop in `<Loading>`

### DIFF
--- a/src/components/InlineLoading/InlineLoading.js
+++ b/src/components/InlineLoading/InlineLoading.js
@@ -28,7 +28,7 @@ export default class InlineLoading extends React.Component {
     /**
      * Specify the description for the inline loading text
      */
-    description: PropTypes.string,
+    description: PropTypes.node,
 
     /**
      * Specify a description that would be used for the loading state SVG title
@@ -92,7 +92,7 @@ export default class InlineLoading extends React.Component {
     };
 
     const loadingText = (
-      <p className={`${prefix}--inline-loading__text`}>{description}</p>
+      <div className={`${prefix}--inline-loading__text`}>{description}</div>
     );
 
     return (
--- a/src/components/InlineLoading/InlineLoading.js
+++ b/src/components/InlineLoading/InlineLoading.js
@@ -31,6 +31,11 @@ export default class InlineLoading extends React.Component {
     description: PropTypes.string,
 
     /**
+     * Specify a description that would be used for the loading state SVG title
+     */
+    iconDescription: PropTypes.string,
+
+    /**
      * Provide an optional handler to be inovked when <InlineLoading> is
      * successful
      */
@@ -51,6 +56,7 @@ export default class InlineLoading extends React.Component {
     const {
       className,
       success,
+      iconDescription,
       description,
       onSuccess,
       successDelay,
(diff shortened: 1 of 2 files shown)


src/components/Loading/Loading.js
```diff
@@ -34,16 +34,29 @@ export default class Loading extends React.Component {
      * Specify whether you would like the small variant of <Loading>
      */
     small: PropTypes.bool,
+
+    /**
+     * Specify a description that would be used to describe the loading state
```
vpicone, line 39: Maybe some more context here around what the end result of this prop actually does (updating the title of the svg)

src/components/Loading/Loading-story.js
```diff
@@ -8,13 +8,14 @@
...
 
-import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { withKnobs, boolean, text } from '@storybook/addon-knobs';
 import Loading from '../Loading';
 
 const props = () => ({
   active: boolean('Active (active)', true),
   withOverlay: boolean('With overlay (withOverlay)', false),
   small: boolean('Small (small)', false),
+  description: text('Description (description)', 'Active loading indicator'),
```
vpicone: This doesn't need to be a knob in my opinion. With a descriptive prop-type comment, there's no reason someone would need to 'see' what their own svg title would look like.

src/components/InlineLoading/InlineLoading.js
```diff
@@ -30,6 +30,11 @@ export default class InlineLoading extends React.Component {
      */
     description: PropTypes.string,
 
+    /**
+     * Specify the description for the inline loading text
+     */
+    iconDescription: PropTypes.string,
```
vpicone, line 36: What's the difference between this and description? Can we just use description instead of adding a new prop? Looks like you might have accidentally not updated the comment block

src/components/InlineLoading/InlineLoading-story.js
```diff
@@ -16,6 +16,10 @@ import InlineLoading from '../InlineLoading';
 
 const props = () => ({
   success: boolean('Loading successful state (success)', false),
+  iconDescription: text(
```
vpicone: Same comment about unnecessary knob. Just because it's in the prop table, doesn't mean we need to make it a knob. I went on a rant about this [https://ibm-studios.slack.com/archives/GCQ5R263T/p1571254152018700?thread_ts=1571245479.016100&cid=GCQ5R263T](on slack), let me know if you need any more context.

vpicone (changes requested):

src/components/InlineLoading/InlineLoading.js
emyarod: `description` is meant to be a node instead of a string, and `iconDescription` is for the SVG title in `<Loading />`

asudoh (approved):
LGTM 👍 - Thanks @emyarod!

asudoh:
@vpicone Would you be able to re-review? Thanks!

vpicone (approved):
How long are we going to be backporting to v6?

emyarod:
until v8 is released I imagine

PR #2423, opened 2019-11-19
Added label props to ToggleSmall.js

ahosny752:


emyarod (commented):
I'm not sure if this is in line with the v9 spec but it looks like the check mark is removed

also not sure how high priority this is for us given the issue is resolved in v10

asudoh:
Agreed with @emyarod - Looks a breaking change.

PR #2424, opened 2019-12-06
fix(Tooltip): Prevent unnecessary setState calls

grommett:
This PR prevents the `handleClickOutside` from calling `setState` if the ToolTip is already closed.

The reason for this check is that, at scale, these calls create _**real**_ performance issues. 

Our app has over 1,000 `<Tooltip>`s. Each of these `Tooltip`s calls `setState({open: false})`. These unnecessary calls to `setState` makes toggling radio buttons and opening Modals very, very slow. Up to two seconds in some cases in our UI. After adding this check the UI behaves normally.

Performance screenshot below.

Screen Shot 2019-12-06 at 11 02 21 AM

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -366,7 +366,7 @@ class Tooltip extends Component {
       evt.target &&
       this._tooltipEl &&
       this._tooltipEl.contains(evt.target);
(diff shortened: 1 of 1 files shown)


asudoh (approved):
LGTM 👍 - Thanks @grommett!

emyarod (approved):
looks good to me

PR #2427, opened 2020-07-16
chore: bump focus-trap-react version

emyarod:
Ref https://github.com/carbon-design-system/carbon/issues/6467

This PR upgrades the `focus-trap-react` dependency to include upstream fixes


tw15egan (approved):

vpicone (approved):
@emyarod Doesn't look like `v7` had any breaking changes, but they added prop-types. We might want to check we're not breaking any of those prop-types. Other than that looks good!

Also, make sure you tag @carbon-design-system/system for reviews so me, Josefina and Scott aren't accidentally pinged for review

emyarod:
it looks like the system devs group is not enabled for this repo

